PR #12, opened 2018-01-12
Handle 404 response on Read

sl1pm4t:
If a datasource or alert notification have been deleted out of band (not using Terraform) the Grafana API will return a 404 error. The standard behaviour in the scenario is to remove the resource from the Terraform state, by setting a blank ID.
This was already happening in the Dashboard resource.


tombuildsstuff (approved):
LGTM - thanks for this 👍

PR #29, opened 2018-07-19
Resource Organization

mlclmj:
Adds support for creating and managing Grafana organizations and their membership with the addition of the `grafana_organization` resource.

This PR also adds acceptance tests and website documentation for the new resource, as well as updating the vendored version of the `go-grafana-api`.

### DIFF
--- a/grafana/resource_organization.go
+++ b/grafana/resource_organization.go
@@ -210,27 +210,27 @@ func collectUsers(d *schema.ResourceData) (map[string]OrgUser, map[string]OrgUse
 	return oldUsers, newUsers
 }
 
-func changes(oldUsers, newUsers map[string]OrgUser) map[string]UserChange {
-	changes := make(map[string]UserChange)
+func changes(oldUsers, newUsers map[string]OrgUser) []UserChange {
+	var changes []UserChange
 	for _, user := range newUsers {
 		oUser, ok := oldUsers[user.Email]
 		if !ok {
-			changes[user.Email] = UserChange{UserAdd, user}
+			changes = append(changes, UserChange{UserAdd, user})
 			continue
 		}
 		if oUser.Role != user.Role {
-			changes[user.Email] = UserChange{UserUpdate, user}
+			changes = append(changes, UserChange{UserUpdate, user})
 		}
 	}
 	for _, user := range oldUsers {
 		if _, ok := newUsers[user.Email]; !ok {
-			changes[user.Email] = UserChange{UserRemove, user}
+			changes = append(changes, UserChange{UserRemove, user})
 		}
 	}
 	return changes
 }
 
-func addIds(d *schema.ResourceData, meta interface{}, changes map[string]UserChange) (map[string]UserChange, error) {
+func addIds(d *schema.ResourceData, meta interface{}, changes []UserChange) ([]UserChange, error) {
 	client := meta.(*gapi.Client)
 	gUserMap := make(map[string]int64)
 	gUsers, err := client.Users()
@@ -240,7 +240,7 @@ func addIds(d *schema.ResourceData, meta interface{}, changes map[string]UserCha
 	for _, u := range gUsers {
 		gUserMap[u.Email] = u.Id
 	}
-	output := make(map[string]UserChange)
+	var output []UserChange
 	create := d.Get("create_users").(bool)
 	for _, change := range changes {
 		id, ok := gUserMap[change.User.Email]
@@ -256,7 +256,7 @@ func addIds(d *schema.ResourceData, meta interface{}, changes map[string]UserCha
 			id = user
 		}
 		change.User.Id = id
-		output[change.User.Email] = change
+		output = append(output, change)
 	}
 	return output, nil
 }
@@ -284,7 +284,7 @@ func createUser(meta interface{}, user string) (int64, error) {
 	return id, err
 }
 
-func applyChanges(meta interface{}, orgId int64, changes map[string]UserChange) error {
+func applyChanges(meta interface{}, orgId int64, changes []UserChange) error {
 	var err error
 	client := meta.(*gapi.Client)
 	for _, change := range changes {
--- a/grafana/resource_organization.go
+++ b/grafana/resource_organization.go
@@ -225,15 +225,20 @@ func changes(stateUsers, configUsers map[string]OrgUser) []UserChange {
 	for _, user := range configUsers {
 		sUser, ok := stateUsers[user.Email]
 		if !ok {
+			// User doesn't exist in Grafana's state for the organization, should be added.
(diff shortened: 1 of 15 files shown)


mlclmj:
cc @tonglil, @msuterski, @sbower

grafana/resource_organization.go
```diff
@@ -0,0 +1,305 @@
...
+
+func CreateOrganization(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gapi.Client)
+	name := d.Get("name").(string)
+	orgId, err := client.NewOrg(name)
+	if err != nil && err.Error() == "409 Conflict" {
+		return errors.New(fmt.Sprintf("Error: A Grafana Organization with the name '%s' already exists.", name))
+	}
+	if err != nil {
+		log.Printf("[DEBUG] creating Grafana organization %s", name)
```
tonglil: Are these `log.Printf` statements new? In the existing code, seems like it is only printed on errors?

mlclmj: You're right - It looks like I used a lot more logging than the rest of the provider does. I'll cut that down to just the warning for removing an organization from the state like the other resources do. Thanks!

grafana/resource_organization.go
```diff
@@ -0,0 +1,304 @@
...
+	roles := []string{"admins", "editors", "viewers"}
+	oldUsers, newUsers := make(map[string]OrgUser), make(map[string]OrgUser)
+	for _, role := range roles {
+		roleName := strings.Title(role[:len(role)-1])
+		old, new := d.GetChange(role)
+		for _, u := range old.([]interface{}) {
+			oldUsers[u.(string)] = OrgUser{0, u.(string), roleName}
+		}
+		for _, u := range new.([]interface{}) {
+			newUsers[u.(string)] = OrgUser{0, u.(string), roleName}
```
tonglil: What happens here if a user's email exists in multiple roles?
It seems like this map of users would overwrite with the last (in this case viewer `roleName`) for that user's string.

grafana/resource_organization.go
```diff
@@ -0,0 +1,304 @@
...
+		return err
+	}
+	return applyChanges(meta, orgId, changes)
+}
+
+func collectUsers(d *schema.ResourceData) (map[string]OrgUser, map[string]OrgUser) {
+	roles := []string{"admins", "editors", "viewers"}
+	oldUsers, newUsers := make(map[string]OrgUser), make(map[string]OrgUser)
+	for _, role := range roles {
+		roleName := strings.Title(role[:len(role)-1])
```
tonglil: This string manipulation between `admins` and `Admin` isn't my favorite, particularly using 2 different strings as the key/name to values, but I'm not seeing an alternative right now.

grafana/resource_organization.go
```diff
@@ -0,0 +1,304 @@
...
+		for _, u := range new.([]interface{}) {
+			newUsers[u.(string)] = OrgUser{0, u.(string), roleName}
+		}
+	}
+	return oldUsers, newUsers
+}
+
+func changes(oldUsers, newUsers map[string]OrgUser) []UserChange {
+	var changes []UserChange
+	for _, user := range newUsers {
```
tonglil: I would just add some comments here to explain that you are adding/updating new users, then removing non-existant old users.

grafana/resource_organization.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+	var output []UserChange
+	create := d.Get("create_users").(bool)
+	for _, change := range changes {
+		id, ok := gUserMap[change.User.Email]
+		if !ok && !create {
+			return nil, errors.New(fmt.Sprintf("Error adding user %s. User does not exist in Grafana.", change.User.Email))
+		}
+		if !ok && create {
+			user, err := createUser(meta, change.User.Email)
```
tonglil: `user` here can be `id` right?

grafana/resource_organization.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+	for _, user := range oldUsers {
+		if _, ok := newUsers[user.Email]; !ok {
+			changes = append(changes, UserChange{Remove, user})
+		}
+	}
+	return changes
+}
+
+func addIds(d *schema.ResourceData, meta interface{}, changes []UserChange) ([]UserChange, error) {
```
tonglil: I would name the function `addIdsToChanges`. For a while I couldn't tell what the difference between the input and output `[]UserChange`s were.

website/docs/r/organization.html.md
```diff
@@ -0,0 +1,95 @@
...
+
+* `editors` - (Optional) A list of email addresses corresponding to users who
+  should be given `editor` access to the organization. Note: users specified
+  here must already exist in Grafana unless 'create_users' is set to true.
+
+* `viewers` - (Optional) A list of email addresses corresponding to users who
+  should be given `viewer` access to the organization. Note: users specified
+  here must already exist in Grafana unless 'create_users' is set to true.
+
+Note - Users specified for each role-group (`admins`, `editors`, `viewers`)
```
tonglil: What happens if the same user is specified in two or more roles?

grafana/resource_organization.go
mlclmj: I hadn't considered that- yes it does, and then flip-flops on each apply... I'll add a sanity check to prevent an email being specified multiple times within an org

mlclmj: I also changed the `old`/`new` terminology to `state`/`config` to clarify that we're comparing between Grafana's state and the desired configuration

grafana/resource_organization.go
mlclmj: Yes definitely. I'm not sure what I was thinking there

grafana/resource_organization.go
```diff
@@ -185,53 +185,67 @@ func ReadUsers(d *schema.ResourceData, meta interface{}) error {
...
-			oldUsers[u.(string)] = OrgUser{0, u.(string), roleName}
+		// Get the lists of users read in from Grafana state (old) and configured (new)
+		state, config := d.GetChange(role)
+		for _, u := range state.([]interface{}) {
+			stateUsers[u.(string)] = OrgUser{0, u.(string), roleName}
 		}
-		for _, u := range new.([]interface{}) {
-			newUsers[u.(string)] = OrgUser{0, u.(string), roleName}
+		for _, u := range config.([]interface{}) {
+			email := u.(string)
```
tonglil: This is helpful, I think you should do it for above as well.

tonglil (commented):
Thanks, looks good.

grafana/resource_organization.go
mlclmj: Good point- best to play it safe

tonglil (approved):

PR #37, opened 2018-10-29
support boolean settings for alert notifications

### LINKED ISSUE #35: PagerDuty setting for grafana_alert_notification is coerced into an invalid value
While trying to set `grafana_alert_notification`s I ran into a problem with the setting for PagerDuty type setting of `autoResolve` being set to a value that didn't work for Grafana.

### Terraform Version
0.11.7

### Affected Resource(s)
- `grafana_alert_notification`

```
resource "grafana_alert_notification" "pagerduty" {
  name = "${var.service_name} - PagerDuty"
  type = "pagerduty"
  is_default = false

  settings {
    "integrationKey" = "${pagerduty_service_integration.grafana.integration_key}"
    "autoResolve" = true
  }
}
```

### Debug Output

### Expected Behavior
The alert notification should have an enabled `autoResolve`. 

### Actual Behavior
The value in Grafana appears to be a `"1"` even when the HCL has `true`. This results in a defaulted `false` for the behavior. The `"1"` is stored in the Grafana DB (as relayed to me by a Grafana.com staff member. 

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
Using hosted grafana.com

### References

I'll add more debug information as I can gain access to it.

DanCech:
Currently it isn't possible to provide settings to Grafana alert notifiers that expect boolean parameters, because they are always sent as JSON strings rather than booleans.

Since Terraform does not support maps with mixed value types, this PR transforms the string "true" into boolean `true` and the string "false" into boolean `false` before assembling the JSON to send to Grafana.

It also transforms the boolean values returned by the Grafana API back into string values so that everything works correctly.

It's not the most graceful solution, but it does seem to work well.  Any suggestions for alternative approaches welcomed!

Fixes #35

### DIFF
--- a/grafana/resource_alert_notification.go
+++ b/grafana/resource_alert_notification.go
@@ -95,11 +95,23 @@ func ReadAlertNotification(d *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
+	settings := map[string]interface{}{}
+	for k, v := range alertNotification.Settings.(map[string]interface{}) {
+		boolVal, ok := v.(bool)
+		if ok && boolVal {
+			settings[k] = "true"
+		} else if ok && !boolVal {
+			settings[k] = "false"
+		} else {
+			settings[k] = v
+		}
+	}
+
 	d.Set("id", alertNotification.Id)
 	d.Set("is_default", alertNotification.IsDefault)
 	d.Set("name", alertNotification.Name)
 	d.Set("type", alertNotification.Type)
-	d.Set("settings", alertNotification.Settings)
+	d.Set("settings", settings)
 
 	return nil
 }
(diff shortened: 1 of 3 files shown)


mlclmj (commented):
This looks good to me! This might be something that will be fixed with rich value types the Terraform [0.12 release](https://www.hashicorp.com/blog/terraform-0-12-rich-value-types), but for now it seems like a good workaround. Could we throw something in the provider documentation for alert notifications to document this behavior?

DanCech:
I'm unclear on whether 0.12 supports a type like `map(interface{})`, which is what we'd need to be able to support this use-case natively.  I'll update the docs so we can get this merged 😃

mlclmj (approved):
This looks good to me!

PR #65, opened 2019-07-12
Add support for more data_source json_data values

kalinon:
Prometheus and other data_source providers need more values than just the base ones included. I have added support for `http_method`, `query_timeout`, and `time_interval`. These also needed changes in the the grafana API lib, which i have an open PR for [here](https://github.com/nytm/go-grafana-api/pull/23). However until it is merged i have pointed it to my fork [here](https://github.com/kalinon/go-grafana-api)


mlclmj:
Closing this as these were added via a big batch in #90 but thank you for the upstream PR @kalinon!

Issue #42, opened 2019-01-11
Allow skipping TLS verify in datasources


subesokun:
Please add support for setting the `skip TLS verify` flag via the `grafana_data_source` resource for Prometheus datasources. Our test environment is using LE staging certificates which will cause Grafana to fail fetching data due to an untrusted certificate.

### Affected Resource(s)
Please list the resources as a list, for example:
- grafana_data_source

mlclmj:
Hey @subesokun, it looks like support for that was added to the client library we use so I don't see any reason we couldn't support this in the provider. If you have the opportunity to, feel free to open a PR for it, if not I can take a look at it when I get a chance.

ericfourrier:
+1 for postgres also

cwcordell:
I'd like to contribute to this issue.
In looking at the code from the go-grafana-api repo, there is a TlsSkipVerify property but it is within the JSONData struct. Terraform requires that AuthType and DefaultRegion be populated but for a use case that needs to use the skip_tls_verify property and is not using a cloud provider it is not feasible. 

I'd like to leave the TlsSkipVerify property for compatibility with existing implementations but add a SkipTlsVerify property under DataSource. This would involve making changes to the Grafana datasource, which I'll do, as well. 

Please provide feedback concerning this implementation.

cwcordell:
I've opened a PR for the solution to this issue. https://github.com/terraform-providers/terraform-provider-grafana/pull/68

PR #68, opened 2019-08-22
add-tls-skip-verify

cwcordell:
Currently, tls-skip-verify is not available as a main level option. This modification adds the ability to set the tls-skip-verify option.

### DIFF
--- a/grafana/resource_data_source.go
+++ b/grafana/resource_data_source.go
@@ -198,7 +198,6 @@ func ReadDataSource(d *schema.ResourceData, meta interface{}) error {
 	d.Set("type", dataSource.Type)
 	d.Set("url", dataSource.URL)
 	d.Set("username", dataSource.User)
-	d.Set("skip_tls_verify", dataSource.User)
 
 	return nil
 }
--- a/grafana/resource_data_source.go
+++ b/grafana/resource_data_source.go
@@ -71,6 +71,12 @@ func ResourceDataSource() *schema.Resource {
 				Sensitive: true,
 			},
 
+			"tls_skip_verify": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  "",
+			},
+
 			"json_data": {
 				Type:     schema.TypeList,
 				Optional: true,
@@ -192,6 +198,7 @@ func ReadDataSource(d *schema.ResourceData, meta interface{}) error {
 	d.Set("type", dataSource.Type)
 	d.Set("url", dataSource.URL)
 	d.Set("username", dataSource.User)
+	d.Set("skip_tls_verify", dataSource.User)
 
 	return nil
 }
@@ -241,6 +248,7 @@ func makeJSONData(d *schema.ResourceData) gapi.JSONData {
 		DefaultRegion:           d.Get("json_data.0.default_region").(string),
 		CustomMetricsNamespaces: d.Get("json_data.0.custom_metrics_namespaces").(string),
 		AssumeRoleArn:           d.Get("json_data.0.assume_role_arn").(string),
+		TlsSkipVerify:           d.Get("tls_skip_verify").(bool),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


cwcordell:
This PR resolved issue #42

grafana/resource_data_source.go
```diff
@@ -241,6 +247,7 @@ func makeJSONData(d *schema.ResourceData) gapi.JSONData {
 		DefaultRegion:           d.Get("json_data.0.default_region").(string),
 		CustomMetricsNamespaces: d.Get("json_data.0.custom_metrics_namespaces").(string),
 		AssumeRoleArn:           d.Get("json_data.0.assume_role_arn").(string),
+		TlsSkipVerify:           d.Get("tls_skip_verify").(bool),
```
mlclmj, line 250: Could we source this from the `json_data` attribute instead since it's part of that struct?

grafana/resource_data_source.go
```diff
@@ -71,6 +71,12 @@ func ResourceDataSource() *schema.Resource {
 				Sensitive: true,
 			},
 
+			"tls_skip_verify": {
```
mlclmj, line 74: (see other comment below too) I think it might be better to have this be an attribute of `json_data` since that's how it was implemented in the upstream library → https://github.com/nytm/go-grafana-api/pull/13/files#diff-4fadc66382f2a85543cefe051bedef99R39

mlclmj (commented):
Hey @cwcordell, thanks for the PR! I left a couple comments about the configured location (and if we could add a teeny note about this into the [docs](https://github.com/terraform-providers/terraform-provider-grafana/blob/master/website/docs/r/data_source.html.md) that'd be great) but other than that this looks great! 😄

grafana/resource_data_source.go
colinrymer: Hey, @mlclmj, I paired a bit with @cwcordell on this and can weigh in on why the attribute was placed where it was:

We started out having the same thought as you and initially looked at putting it in the `json_data` attribute, but while that attribute is optional, if you provide it, you *must* provide 2 of the keys defined within it: `auth_type`, `default_region`.  We didn't see an obvious way to make those those keys optional while ensuring that if one of them is set, the other is.

Because of that, we decided that putting this as a separate attribute made more sense and it more closely aligns with the Grafana UI, where "Skip TLS Verify" is an option at the same level as "Basic Auth", etc.

If you're good with the above rationale and are otherwise okay with this PR, I can make sure we get the docs updated as part of this as well.

Thanks!

grafana/resource_data_source.go
cwcordell: There are some issues around the json_data block where certain attributes are required and not needed. In the GUI for adding the resource, skip TLS verify is not bound to any other requirements. I believe TlsSkipVerify should be independant of the json_data block due to this.

grafana/resource_data_source.go
cwcordell: Colin explained it well.

grafana/resource_data_source.go
cwcordell: I will add some doc notes for the implementation. Thanks!

colinrymer:
@mlclmj, care to re-review this in light of the responses from @cwcordell and me?

colinrymer:
Bumping this again, @mlclmj. We're keen to leverage this functionality, but don't want to maintain our own fork of the provider.

Thanks!

mlclmj:
Hey! Sorry for the delay here folks. 

Looking at this a little more and from your comments, I think the root problem here is really that the `json_data` and `secure_json_data` attributes only support a CloudWatch datasource at the moment, and I think we can move beyond that and enable others by adding conditional validation of the structs based on the datasource type. I'll open an issue detailing this.

I think it's great to add the `tls_skip_verify` option, but I'm still not keen on adding it outside of the `json_data` struct. While I get that it would be convenient in this case, it breaks consistency with both the upstream provider and the Grafana API and leaving a CloudWatch-specific struct doesn't do us any good the next time another field needs to be added.

mlclmj:
I'm going to close this since I believe that support for this use case was merged earlier though #90, but please feel free to reopen this PR or a new issue if there's a use case I've missed here.

PR #73, opened 2019-10-03
Argument names must not be quoted

tomweston:
Argument names must not be quoted in grafana_alert_notification settings


mlclmj (approved):
Thanks!

PR #74, opened 2019-10-18
exporting uid from response

rbkasat:


mlclmj (changes requested):
Hey @rbkasat -- I think we need to add `uid` to the schema as well for this

tonglil:
I believe all of the required changes are in https://github.com/terraform-providers/terraform-provider-grafana/pull/87, and thus this can be closed in favor of that one.

tonglil:
Closed by #87.

PR #75, opened 2019-10-20
deps: Bump nytm/go-grafana-api to 0.2.0

radeksimko:
This is to unblock https://github.com/terraform-providers/terraform-provider-grafana/pull/46


mlclmj (approved):
👍

PR #78, opened 2019-12-01
Mark secret_key in secure_json_data as sensitive

Infra-Red:
As it contain sensitive information.

Currently there is a bug in Terraform, which ignores `Sensitive: true` on fields with type List.

See https://github.com/hashicorp/terraform-plugin-sdk/issues/76 for details.


mlclmj (approved):
Hey @Infra-Red--thanks for flagging this 👍

PR #82, opened 2019-12-29
terraform-provider-grafana - dashboard.uid

### LINKED ISSUE #80: Editing Dashboard 'title' creates new dashboard - leaving old dashboard behind
### Terraform Version
Terraform v0.12.9
- provider.grafana v1.5.0

### Affected Resource(s)
- grafana_dashboard

```hcl
provider "grafana" {
  url = "https://grafana.${var.platform_domain_name}"
  auth = "${var.grafana_admin_username}:${var.grafana_admin_password}"
}

  config_json = "${file("basic_graph.json")}"
}
```

### Expected Behavior
Based on understanding of how the Grafana provider works, running `terraform plan` would have shown that my dashboard resource was going to be replaced. After running `terraform apply` I would see only the renamed dashboard in Grafana UI rather than both new and old named dashboards.

### Actual Behavior
Running `terraform plan` shows that `grafana_dashboard.test will be updated in-place`. Running `terraform apply` performs the indicated name change for dashboard - in place.
When viewing results within Grafana UI, the new dashboard exists with its rename (and a different uid. The old dashboard is still there as well.

### Steps to Reproduce
1. Create visualization through Grafana
2. Export .json file
3. Remove dashboard from Grafana UI
4. Add resource to terraform config file
5. Run `terraform plan` and confirm a new dashboard resource will be added
6. Run `terraform apply` and see new dashboard in Grafana UI
7. Update .json config file to change "title" of dashboard
8. Run `terraform plan`
9. Run `terraform apply`
10. View both new and old dashboards on Grafana UI

matthiasng:
### Affected Resource(s)
- [dashboard](https://www.terraform.io/docs/providers/grafana/r/dashboard.html)

### Description
Provider removes uid field from the dashboard configuration, no matter if we want to create or update the dashboard. In case we want to update the title of an existing dashboard the provider instead creates a new one and leave the old named dashboard behind. This happens because the grafana API creates a new dashboard if `uid` and `id` not set.

To revole this problem, the provider now take care of the `uid` field.

Note: Only works with v5.0+ (unchanged pre v5.0 behaviour)

### References
resolves #80

### Tests
Grafana v4.5 and latest, make testacc + manual test


matthiasng:
My fault. Wrong branch... See PR #84

matthiasng:
My fault. Wrong branch... See PR #84

PR #84, opened 2020-01-05
terraform-provider-grafana - dashboard.uid

### LINKED ISSUE #80: Editing Dashboard 'title' creates new dashboard - leaving old dashboard behind

matthiasng:
### Affected Resource(s)
- [dashboard](https://www.terraform.io/docs/providers/grafana/r/dashboard.html)

### Description
Provider removes uid field from the dashboard configuration, no matter if we want to create or update the dashboard. In case we want to update the title of an existing dashboard the provider instead creates a new one and leave the old named dashboard behind. This happens because the grafana API creates a new dashboard if `uid` and `id` not set.

To revole this problem, the provider now take care of the `uid` field.

Note: Only works with v5.0+ (unchanged pre v5.0 behaviour)

### References
resolves #80

### Tests
Grafana v4.5 and latest, make testacc + manual test


shalyng:
@mlclmj - Any updates on when we might see this get merged?

mlclmj (commented):
Since the upstream library now supports managing dashboards by uid, how would you feel about migrating to using that and deprecating the use of slugs entirely?

matthiasng:
> Since the upstream library now supports managing dashboards by uid, how would you feel about migrating to using that and deprecating the use of slugs entirely?

@mlclmj I think thats a good idea. My current PR is more a workaround to support renaming.  I will start working next week on this.

Just to rember, this means Grafana < 5.0 wont work longer.

mlclmj:
Awesome! 

I think it would be fair to publish a breaking change if we need to, given that Grafana v5 has been available for over two years now (and Grafana has progressed 3 major versions).

But I'll also put on my thinking cap in the meantime and see if I can come up with a sensible way we can do this that keeps backwards compatibility.

matthiasng:
After digging a bit through grafana's documentation, i think it makes sense to drop support for versions prior v5.

State migration sould be possible via https://grafana.com/docs/grafana/latest/http_api/dashboard/#get-dashboard-by-slug
If someone uses the new provider with an old grafana instance, we can fail with something like "please upgrade grafara or use provider version <= 1.6.0"

I would be fine with this. What dou you think ?

matthiasng:
@mlclmj code is ready to review

This PR now includes  a complete state migration from slug to uid. The only downside that i can see is, if someone relies on the ID attribute (which is not documented), you get the uid instead now.

mlclmj:
Awesome! I think dropping support for Grafana v4 is a breaking change, (which is completely ok), but thanks so much for adding the state migration function. I think we'll just have to mention the changed use of the `id` field in the release notes as well.

I'll try to test this out later this week so we can get this rolled into a release (I might dig around and see if there's any other major changes we could add as well).

website/docs/r/dashboard.html.md
```diff
@@ -8,16 +8,33 @@ description: |-
 
 # grafana\_dashboard
 
+~> **NOTE:** This resource requires Grafana v5.0+. You can use the provider prior or equal to version 1.5.0 if you are using an older Grafana version.
+
 The dashboard resource allows a dashboard to be created on a Grafana server.
 
 ## Example Usage
 
+### Basig usage
```
tonglil:
```suggestion
### Basic usage
```

tonglil (approved):
LGTM, one typo

tonglil:
We should do a release for v1.6.0 before merging this one.

tonglil:
@mlclmj we should do a release for v1.6.0 before merging this one for v2.0.0.

neelneelpurk:
@mlclmj @tonglil - Any updates on when this will get merged?

trotttrotttrott:
Closing in favor of https://github.com/grafana/terraform-provider-grafana/pull/137.

Issue #54, opened 2019-03-25
[Feature Request] Support setting version for Elasticsearch datasource


mattrussell-sonocent:
It'd be great to be able to configure the version of Elasticsearch (and potentially other configurable properties of the data source):

![image](https://user-images.githubusercontent.com/45822724/54931268-933f9700-4f10-11e9-89f8-4b0567ee9edd.png)

### Affected Resource(s)
- grafana_data_source

jmanuelortizn:
same here

tonglil:
Done in https://github.com/terraform-providers/terraform-provider-grafana/pull/90

PR #90, opened 2020-04-13
Update resource grafana_data_source

### LINKED ISSUE #22: expand json_data field usage.
Data sources have more that already defined json_data. Elasticsearch have own set of variables, tlsSkipVerify  also is not supported,

### LINKED ISSUE #42: Allow skipping TLS verify in datasources

mlclmj:
## Summary
Makes long-overdue updates to the data source resource to introduce additional datasource types that can be fully configured with this provider including:
- Elasticsearch
- Graphite
- OpenTSDB
- Prometheus
- MySQL
- MSSQL
- Postgres

This change also adds better testing for the types of datasources that can be added.

## Details
- Updated the upstream library to [v0.3.1](https://github.com/nytm/go-grafana-api/releases/tag/v0.3.1)
- Reordered schema fields into alphabetical order for readability/maintainability
- Added table tests for the various data source types supported. Note that the Grafana API doesn't actually validate the presence of any of these fields or their values, however this is valuable in making sure that the upstream library is parsing and submitting values to Grafana as expected.
- As part of this work, the `access_key` and `secret_key` fields of the `secure_json_data` block are now marked as optional rather than required. Previously this was in place as the only datasource type using the `secret_json_data` block was Cloudwatch, though this is no longer the case. This shouldn't result in a breaking change for anyone configuring Cloudwatch datasources through this provider.
- Vendor folder was updated using `go mod vendor`

Resolves #22 
Resolves #42


mlclmj:
Resolves #54

mlclmj:
Resolves #68

tonglil (approved):
Nice set of changes, LGTM

moseskt:
@mlclmj any eta for full support for configuring Azure Monitor via terraform?

PR #98, opened 2020-06-03
Fix build, use -mod=readonly

tonglil:
Bump the go version.


mlclmj:
awesome!

I think we need a `.go-version` file as well, like how the AWS provider does → https://github.com/terraform-providers/terraform-provider-aws/blob/master/.go-version

tonglil:
Do you know what that file does? I actually can't tell and have never encountered that before.

mlclmj (approved):
thanks!

PR #100, opened 2020-06-07
resource/data_source: add support for stackdriver privatekey

### LINKED ISSUE #91: grafana_data_source does not set the service account key for the stackdriver datasource
### Terraform Version
0.12.24

### Affected Resource(s)
- grafana_data_source

```hcl
resource "grafana_data_source" "stackdriver-datasource" {
name = local.datasource_name
type = "stackdriver"
secure_json_data {
access_key = local.service_account_key
secret_key = ""
}
}
```
### Expected Behavior
What should have happened?
The stackdriver datasource is created in Grafana with the service account key defined.

### Actual Behavior
What actually happened?
The stackdriver datasource is created, however, no key is defined. Manually setting the key in the datasource works, and allows the datasource to work. Terraform does not give any indication of failure before or after manually setting the key.

### Steps to Reproduce
1. `terraform apply`

anGie44:
Closes #91 

Release note for [CHANGELOG](https://github.com/terraform-providers/terraform-provider-grafana/blob/master/CHANGELOG.md):
```release-note
resource/data_source: add private_key attribute for stackdriver data_sources
```

Output of acceptance tests:
```
=== RUN   TestProvider
--- PASS: TestProvider (0.00s)
=== RUN   TestProvider_impl
--- PASS: TestProvider_impl (0.00s)
=== RUN   TestAccAlertNotification_basic
--- PASS: TestAccAlertNotification_basic (0.18s)
=== RUN   TestAccAlertNotification_invalid_frequence
--- PASS: TestAccAlertNotification_invalid_frequence (0.04s)
=== RUN   TestAccAlertNotification_reminder_no_frequence
--- PASS: TestAccAlertNotification_reminder_no_frequence (0.03s)
=== RUN   TestAccDashboard_basic
--- PASS: TestAccDashboard_basic (0.33s)
=== RUN   TestAccDashboard_folder
--- PASS: TestAccDashboard_folder (0.34s)
=== RUN   TestAccDashboard_disappear
--- PASS: TestAccDashboard_disappear (0.16s)
=== RUN   TestAccDataSource_basic
--- PASS: TestAccDataSource_basic (1.97s)
=== RUN   TestAccFolder_basic
--- PASS: TestAccFolder_basic (0.18s)
=== RUN   TestAccOrganization_basic
--- PASS: TestAccOrganization_basic (0.48s)
=== RUN   TestAccOrganization_users
--- PASS: TestAccOrganization_users (0.74s)
=== RUN   TestAccOrganization_defaultAdmin
--- PASS: TestAccOrganization_defaultAdmin (0.49s)
```

### DIFF
--- a/grafana/resource_data_source.go
+++ b/grafana/resource_data_source.go
@@ -188,6 +188,11 @@ func ResourceDataSource() *schema.Resource {
 							Optional:  true,
 							Sensitive: true,
 						},
+						"private_key": {
+							Type:      schema.TypeString,
+							Optional:  true,
+							Sensitive: true,
+						},
 						"secret_key": {
 							Type:      schema.TypeString,
 							Optional:  true,
@@ -369,6 +374,7 @@ func makeSecureJSONData(d *schema.ResourceData) gapi.SecureJSONData {
 		AccessKey:         d.Get("secure_json_data.0.access_key").(string),
 		BasicAuthPassword: d.Get("secure_json_data.0.basic_auth_password").(string),
 		Password:          d.Get("secure_json_data.0.password").(string),
+		PrivateKey:        d.Get("secure_json_data.0.private_key").(string),
 		SecretKey:         d.Get("secure_json_data.0.secret_key").(string),
 		TlsCACert:         d.Get("secure_json_data.0.tls_ca_cert").(string),
 		TlsClientCert:     d.Get("secure_json_data.0.tls_client_cert").(string),
--- a/website/docs/r/data_source.html.md
+++ b/website/docs/r/data_source.html.md
@@ -13,8 +13,8 @@ The data source resource allows a data source to be created on a Grafana server.
 ## Example Usage
 
 The required arguments for this resource vary depending on the type of data
-source selected (via the `type` argument). The following example is for
-InfluxDB. See [Grafana's Data Sources Guides][datasources] for more details on
+source selected (via the `type` argument). The following examples are for
+InfluxDB, CloudWatch, and Google Stackdriver. See [Grafana's Data Sources Guides][datasources] for more details on
 the supported data source types and the arguments they use.
 
 [datasources]: https://grafana.com/docs/grafana/latest/features/datasources/
@@ -47,6 +47,19 @@ resource "grafana_data_source" "test_cloudwatch" {
   secure_json_data {
     access_key = "123"
     secret_key = "456"
+  }
+}
+```
+
+For a Stackdriver datasource:
+
+```hcl
+resource "grafana_data_source" "test_stackdriver" {
+  type = "stackdriver"
+  name = "sd-example"
+
+  secure_json_data {
+    private_key = "-----BEGIN PRIVATE KEY-----\nprivate-key\n-----END PRIVATE KEY-----\n"
   }
 }
 ```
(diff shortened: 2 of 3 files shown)


grafana/resource_data_source.go
```diff
@@ -169,6 +169,7 @@ func ResourceDataSource() *schema.Resource {
 			},
 			"secure_json_data": {
 				Type:      schema.TypeList,
+				MaxItems:  1,
```
anGie44: from the `makeSecureJSONData` method, it only looks at the first element so we could reinforce the number of elements here

grafana/resource_data_source.go
```diff
@@ -365,13 +371,35 @@ func makeJSONData(d *schema.ResourceData) gapi.JSONData {
...
-	return gapi.SecureJSONData{
-		AccessKey:         d.Get("secure_json_data.0.access_key").(string),
-		BasicAuthPassword: d.Get("secure_json_data.0.basic_auth_password").(string),
-		Password:          d.Get("secure_json_data.0.password").(string),
-		SecretKey:         d.Get("secure_json_data.0.secret_key").(string),
-		TlsCACert:         d.Get("secure_json_data.0.tls_ca_cert").(string),
-		TlsClientCert:     d.Get("secure_json_data.0.tls_client_cert").(string),
-		TlsClientKey:      d.Get("secure_json_data.0.tls_client_key").(string),
+	secureJSONData := gapi.SecureJSONData{}
+	if v, ok := d.GetOk("secure_json_data"); ok {
```
mlclmj: I'm not sure this is necessary, why these changes?

The go-grafana-api won't include any of these fields if they're set to default values.

grafana/resource_data_source.go
mlclmj: I worry that this might cause confusion for others contributing to this resource.

If you feel strongly about this change, could we at least add this to `json_data` as well and add comments clarifying what the effect of this actually is?

website/docs/r/data_source.html.md
```diff
@@ -191,6 +204,8 @@ fields to operate properly.
 
 * `secret_key` - (CloudWatch) The secret key to use to access the data source.
 
+* `private_key` - (Stackdriver) The [service account key](https://cloud.google.com/iam/docs/creating-managing-service-account-keys) `private_key` to use to access the data source.
```
mlclmj: could we keep these alphabetized?

grafana/resource_data_source.go
```diff
@@ -193,6 +194,11 @@ func ResourceDataSource() *schema.Resource {
 							Optional:  true,
 							Sensitive: true,
 						},
+						"private_key": {
```
mlclmj: sorry to sound like a broken record, could we also keep these in alpha order?

mlclmj (commented):
I left a couple comments but for the most part this looks great! thanks for jumping in to work on this!

grafana/resource_data_source.go
anGie44: sure thing!

grafana/resource_data_source.go
anGie44: oh no worries, I can revert this back. when adding the new field `private_key`, I noticed we're only getting the first element of the `secure_json_data` set so if a user were to put more then one set in their config,  then only the first will get interpreted. from a user perspective, I thought it may cause confusion to be able to provide more than 1 set but not sure if that's something to really guard against.

grafana/resource_data_source.go
anGie44: wasn't sure if the struct's fields should be set to empty strings if the API will omit them in the long run, but i'll revert to keep the same pattern as before 👍

anGie44:
thanks @mlclmj for the review! please find my latest commits with the requested changes :)

mlclmj (approved):
Looks great! Mind rebasing to fix the build?

PR #102, opened 2020-06-08
resource/playlist: add new resource

### LINKED ISSUE #25: [Feature Request] Manage Grafana playlists through Terraform
Hi there,

This is a request to add Grafana playlist resource to the provider
http://docs.grafana.org/reference/playlist/

### Terraform Version
Terraform v0.11.7
+ provider.grafana v1.0.2

### Affected Resource(s)
- grafana_dashboard

### Expected Behavior
Terraform provider is playlist aware

### Actual Behavior
Terraform managed dashboards are often recreated with a new ID.  Playlists reference dashboards by ID and will lose previously managed dashboards.  If terraform was playlist-aware, the playlist resource could reference a list of dashboards.

A grafana "bug" that has been exposed with this - if a playlist ends up empty (no dashboards attached), it becomes broken/undeletable/uneditable.  To fix this, we've manually editable the database.

anGie44:
Closes #25 

Notes:
* the `go-grafana-api` `PlaylistItem` struct doesn't store a dashboard ID so dashboards are referenced by name instead
* schema naming conventions for dashboards in a playlist (i.e. `item`) match those in `go-grafana-api`
* made my best guess on the description of `item` attributes `type` and `value`

Output of acceptance tests:
```
--- PASS: TestAccPlaylist_disappears (0.69s)
--- PASS: TestAccPlaylist_basic (1.01s)
--- PASS: TestAccPlaylist_update (1.84s)
```


tonglil (approved):
LGTM, thanks!

grafana/resource_playlist_test.go
```diff
@@ -0,0 +1,220 @@
...
+				),
+			},
+			{
+				Config: testAccPlaylistConfigUpdate(updatedName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccPlaylistCheckExists(rn, &playlist),
+					resource.TestMatchResourceAttr(rn, "id", regexp.MustCompile(`\d+`)),
+					resource.TestCheckResourceAttr(rn, "name", updatedName),
+					resource.TestCheckResourceAttr(rn, "item.#", "1"),
+					resource.TestMatchResourceAttr(rn, "id", regexp.MustCompile(`\d+`)),
```
matthiasng: duplicate
```suggestion
```

anGie44: ahh thanks for catching @matthiasng!

grafana/resource_playlist.go
```diff
@@ -0,0 +1,171 @@
...
+func resourcePlaylistCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gapi.Client)
+
+	playlist := gapi.Playlist{
+		Name:     d.Get("name").(string),
+		Interval: d.Get("interval").(string),
+		Items:    expandPlaylistItems(d.Get("item").(*schema.Set).List()),
+	}
+
+	log.Printf("[DEBUG] Creating Playlist %s", playlist.Name)
```
tonglil, line 69: Could you remove these debug messages?

grafana/resource_playlist.go
```diff
@@ -0,0 +1,171 @@
...
+		Delete: resourcePlaylistDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
```
matthiasng, line 26: i think using forcenew here is not a good idea. you sould change you update method, to use the playlist id, instead of recreating the playlist.
See https://github.com/nytm/go-grafana-api/blob/master/playlist.go#L58

grafana/resource_playlist.go
```diff
@@ -0,0 +1,171 @@
...
+							Type:     schema.TypeInt,
+							Required: true,
+						},
+						"title": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"type": {
+							Type:     schema.TypeString,
+							Optional: true,
```
matthiasng, line 47: This souldn't be "required" ? Creating a playlist without a type and value make no sense and also grafana fails.

grafana/resource_playlist.go
```diff
@@ -0,0 +1,171 @@
...
+						},
+						"value": {
+							Type:     schema.TypeString,
+							Optional: true,
```
matthiasng, line 51: Same as "type" field

grafana/resource_playlist.go
```diff
@@ -0,0 +1,171 @@
...
+				Required: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"order": {
+							Type:     schema.TypeInt,
+							Required: true,
+						},
+						"title": {
+							Type:     schema.TypeString,
+							Required: true,
```
matthiasng, line 43: I looked at the grafana documentation, but I'm not sure what the title is for. After some manual tests, I think this field is useless. It is not possible to set or change the title in the plalist editor ui. If it is set through the API, only the line in the playlist editor will change, nothing more. Therefore I think that this field should be optionally or removed completely.

matthiasng (commented):
The usage example in the documentation is okay, but you missed something importent. There is not way to connect a Terraform managed dashboard with a Terraform managed playlist.

This is not a problem of your playlist implementation, but of the dashboard resource:
- `grafana_dashboard` doesn't support dashboard tags
- `grafana_dashboard` has not [Grafana Dashboard ID](https://grafana.com/docs/grafana/latest/http_api/dashboard/#create-update-dashboard) (see response example json `dashboard.id: 104` ) attribute.

So something like this wont work:
```
resource "grafana_dashboard" "example_1" {
  config_json = file("example-1.json")
}

resource "grafana_playlist" "playlist" {
  name     = "test"
  interval = "5m"

  item {
    order = 1
    type  = "dashboard_by_id"
    value = grafana_dashboard.example_1.dashboard_id
  }
}

```
In my opinion, adding a `dashboard_id` to `grafana_dashboard` read method should be sufficient to resolve the problem.

All comment are results from a test with Grafana 7.0.3
Let me know if i can help somewhere.

grafana/resource_playlist.go
anGie44: will do!

grafana/resource_playlist.go
anGie44: ah true the `Id` is accessible, will change this 👍

grafana/resource_playlist.go
anGie44: tbh i looking at the API docs, it was unclear what was required vs. optional but interesting enough you can make a playlist with just the `title` and `interval` like seen in the Grafana UI and the `basic` acceptance test..I don't have strong feelings one way or the other so happy to change this to `Required` if that makes more sense

grafana/resource_playlist.go
anGie44: ^^similar response added to the `type` comment

grafana/resource_playlist.go
anGie44: ah ok, I thought the `title` was equivalent to the dashboard `name` you see in the UI -- happy to remove as well

anGie44: this GET and subsequent POST method gave me the impression the title was the dashboard name: https://grafana.com/docs/grafana/latest/http_api/playlist/#get-playlist-dashboards

anGie44:
hi @matthiasng, thanks for the review and your testing. as noted in the PR description, not having the dashboard `ID` accessible is definitely a limitation when using this resource so I totally agree. if you could clarify the definitions of the playlistitem fields that would be great! i've commented above where I have room for doubt

matthiasng:
@anGie44 the playlist docs are not the best. The whole playlist API makes zero sense and i think this part of grafana is not well maintained.

All fields are optional. So something like this works:
```
//create request
{
    "name": "test",
    "interval": "5m",
    "items": [
        {},
        {}
    ]
}

//create response
{
  "id": 14,
  "name": "test",
  "interval": "5m"
}

//update request
{
    "playlistId": 14,
    "name": "???",
    "interval": "5m",
    "items": [
        {},
        {},
        {}
    ]
}

//update response
{
  "id": 13,
  "name": "???",
  "interval": "5m",
  "items": [
    {
      "id": 114,
      "playlistid": 13,
      "type": "",
      "title": "",
      "value": "",
      "order": 1
    },
    {
      "id": 115,
      "playlistid": 13,
      "type": "",
      "title": "",
      "value": "",
      "order": 2
    },
    {
      "id": 116,
      "playlistid": 13,
      "type": "",
      "title": "",
      "value": "",
      "order": 3
    }
  ]
}

```

** PlaylistItem fields: **

`order` Doesn't matters. Grafana always uses the JSON order.

`type` Can be `dashboard_by_id` or `dashboard_by_tag`.

`value` this can be a dashboard id or a tag. depends on the type

`title` You are right, the title sould be the dashbaord name or tag. The following can be found in [Grafanas UI code](https://github.com/grafana/grafana/blob/master/public/app/features/playlist/playlist_edit_ctrl.ts):
```js

  addPlaylistItem(playlistItem: PlaylistItem) {
    // playlistItem already contains the title which is the the name of the selected dashboard
    playlistItem.value = playlistItem.id.toString();
    playlistItem.type = 'dashboard_by_id';
    playlistItem.order = this.playlistItems.length + 1;

    this.playlistItems.push(playlistItem);
    this.filterFoundPlaylistItems();
  }

  addTagPlaylistItem(tag: { term: any }) {
    const playlistItem: any = {
      value: tag.term,
      type: 'dashboard_by_tag',
      order: this.playlistItems.length + 1,
      title: tag.term,
    };

    ...
  }
```

I think it would be best to handle the `title` like Grafanas UI and not allow the user to change it.
- `dashboard_by_tag` -> same value as `value`
- `dashboard_by_id` -> dashboard name

trotttrotttrott (commented):
@anGie44 this is really great! A lot has changed since you first opened this though. Would you mind rebasing to get it caught up?

anGie44:
Hi @trotttrotttrott 👋 Apologies for the delay here! I'm going to revisit this work today -- lmk if there's anything else I should take into account, aside from upstream API changes and/or changes in the repo's master branch.

anGie44:
Closing in favor of #264

PR #114, opened 2020-09-02
Settings is an argument not a block

Arola1982:
Copying the example in the readme, fails with error

```
Error: Unsupported block type
```


dotdc:
+1 This worked for me.

trotttrotttrott (approved):
Thanks @Arola1982!

PR #116, opened 2020-09-17
Change user lists to sets

medains:
Users as lists may not be reported back from grafana in the same order - resulting in continually reported resource changes.

Treating the users as sets instead deals with this nicely.

### DIFF
--- a/grafana/resource_organization.go
+++ b/grafana/resource_organization.go
@@ -206,7 +206,7 @@ func collectUsers(d *schema.ResourceData) (map[string]OrgUser, map[string]OrgUse
 		roleName := strings.Title(role[:len(role)-1])
 		// Get the lists of users read in from Grafana state (old) and configured (new)
 		state, config := d.GetChange(role)
-		for _, u := range state.([]interface{}) {
+		for _, u := range state.(*schema.Set).List() {
 			email := u.(string)
 			// Sanity check that a user isn't specified twice within an organization
 			if _, ok := stateUsers[email]; ok {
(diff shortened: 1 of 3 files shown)


trotttrotttrott (commented):
Thanks, @medains. I agree with this change, but it breaks some acceptances tests because we can't read elements from sets as the tests are trying to do. Could you please update them? Removing the following should do it:

* https://github.com/grafana/terraform-provider-grafana/blob/71990e29fb437b00cb578a8d4ee502302da620f6/grafana/resource_organization_test.go#L65-L67
* https://github.com/grafana/terraform-provider-grafana/blob/71990e29fb437b00cb578a8d4ee502302da620f6/grafana/resource_organization_test.go#L89-L91
* https://github.com/grafana/terraform-provider-grafana/blob/71990e29fb437b00cb578a8d4ee502302da620f6/grafana/resource_organization_test.go#L140-L142
* https://github.com/grafana/terraform-provider-grafana/blob/71990e29fb437b00cb578a8d4ee502302da620f6/grafana/resource_organization_test.go#L164-L169

Also please remove the following from the [organization resource docs](https://github.com/grafana/terraform-provider-grafana/blob/71990e29fb437b00cb578a8d4ee502302da620f6/website/docs/r/organization.html.md):

> Note - Users specified for each role-group (admins, editors, viewers) should be listed in ascending alphabetical order (A-Z). By defining users in alphabetical order, Terraform is prevented from detecting unnecessary changes when comparing the list of defined users in the resource to the (ordered) list returned by the Grafana API.

medains:
Updated as requested

trotttrotttrott (approved):
Awesome. Thanks, @medains 👍

PR #118, opened 2020-09-24
Improve README.md and GNUmakefile

gw0:
Some improvements to `README.md` and `GNUmakefile` to make it clearer how to only build it in current directory or install as a Go binary with a symlink from Terraform user plugins directory (to make it work with Terraform).

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -7,7 +7,12 @@ GRAFANA_VERSION ?= "latest"
 default: build
 
 build: fmtcheck
+	go build
+
+install: build
 	go install
+	mkdir -p ~/.terraform.d/plugins
+	ln -fs $(GOPATH)/bin/terraform-provider-grafana $(HOME)/.terraform.d/plugins/
 
 test: fmtcheck
 	go test -i $(TEST) || exit 1
--- a/README.md
+++ b/README.md
@@ -23,25 +23,38 @@ $ mkdir -p $GOPATH/src/github.com/terraform-providers; cd $GOPATH/src/github.com
 $ git clone git@github.com:terraform-providers/terraform-provider-grafana
 ```
 
-Enter the provider directory and build the provider
+Build the provider in place:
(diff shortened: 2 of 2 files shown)


GNUmakefile
```diff
@@ -7,7 +7,12 @@ GRAFANA_VERSION ?= "latest"
 default: build
 
 build: fmtcheck
+	go build
+
+install: build
 	go install
+	mkdir -p ~/.terraform.d/plugins
+	ln -fs $(GOPATH)/bin/terraform-provider-grafana $(HOME)/.terraform.d/plugins/
```
trotttrotttrott, line 15: I don't think this will work with Terraform 0.13.x. I do like the idea of making it easy for folks to install a test binary locally though. I did some experimentation recently and ended up with this to make it work:

```
▸ tree .terraform/
.terraform/
└── plugins
    ├── registry.terraform.io
    │   └── grafana
    │       └── grafana
    │           └── 1.5.0
    │               └── darwin_amd64
    │                   └── terraform-provider-grafana
    └── selections.json

6 directories, 2 files
```

This was not in my home directory though. It was in a root module. My provider config was:

```
terraform {
  required_providers {
    grafana = {
      source = "grafana/grafana"
    }
  }
}

provider "grafana" {
  url  = "http://localhost:3001"
  auth = "admin:admin"
}
```

And initialized with:

```
terraform init -get-plugins=false
```
...to make sure Terraform didn't attempt to download the latest, published binary.

I would imagine we could at least get the installation part would work similarly putting the binary in $HOME.

trotttrotttrott (commented):
@gw0 thanks for this, but a lot has changed since it was opened and I'm afraid much of it is outdated now. Feel free to update if you'd like, otherwise, let's close this for now.

trotttrotttrott:
Closing this since it's a bit outdated. Thanks, @gw0.

PR #119, opened 2020-09-24
Add support for providing additional HTTP headers

gw0:
Sometimes the Grafana server is behind a proxy or application firewall and it is necessary to provide additional HTTP headers to establish a connection, such as authentication tokens or cookies. In general sense these are represented as key-value pairs of strings. To also add support to provide these using an environment variable `GRAFANA_HTTP_HEADERS` it should contain a JSON-encoded values.

Usage:

```
provider "grafana" {
  url  = "http://172.17.0.1:3000"
  auth = "admin:admin"
  headers = {
    "Cookies" = "foo=bar"
  }
}
```

```bash
$ GRAFANA_HTTP_HEADERS='{"Cookies":"foo=bar"}' terraform plan
```

Depends on PR https://github.com/nytm/go-grafana-api/pull/63.

### DIFF
--- a/grafana/provider.go
+++ b/grafana/provider.go
@@ -1,6 +1,9 @@
 package grafana
 
 import (
+	"encoding/json"
+	"os"
+
 	"github.com/hashicorp/terraform/helper/logging"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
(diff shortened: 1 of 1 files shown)


gw0:
Due to changed dependency this PR was replaced with PR https://github.com/grafana/terraform-provider-grafana/pull/134.

PR #120, opened 2020-10-04
Adding team resource functionality

jonathan-dorsey:
This pull request adds team and team member functionality to the provider. Note: This code won't work until the PR https://github.com/nytm/go-grafana-api/pull/64 is incorporated into the Grafana go api. We need the team id to be returned to the provider when a team is added.


trotttrotttrott:
Hi @jonathan-dorsey, this looks great. Since you opened this, the api client got switched to https://github.com/grafana/grafana-api-golang-client so we can have folks in the `grafana` org maintain it. Would you mind updating and rebasing?

jonathan-dorsey:
> Hi @jonathan-dorsey, this looks great. Since you opened this, the api client got switched to https://github.com/grafana/grafana-api-golang-client so we can have folks in the `grafana` org maintain it. Would you mind updating and rebasing?

@trotttrotttrott  - thanks for the update! I opened a new [PR](https://github.com/grafana/grafana-api-golang-client/pull/6) for the grafana project. Let me know if there is anything else you would like me to change/add.

jonathan-dorsey:
@trotttrotttrott - PR is ready for review. I updated the project to use the new version of the go api that includes my changes for the team ID.

trotttrotttrott (commented):
Could you run `go mod vendor` so `vendor/` is updated? Not sure why CI uses `-mod=readonly`. Will sort that out later. Also if you don't mind, `go mod tidy`. There's some extra API client references left over.

jonathan-dorsey:
@trotttrotttrott  - commands run. Let me know if you need anything else. Thanks!

grafana/resource_team.go
```diff
@@ -0,0 +1,264 @@
...
+func ReadMembers(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gapi.Client)
+	teamID, _ := strconv.ParseInt(d.Id(), 10, 64)
+	teamMembers, err := client.TeamMembers(teamID)
+	if err != nil {
+		return err
+	}
+	memberSlice := []string{}
+	for _, teamMember := range teamMembers {
+		memberSlice = append(memberSlice, teamMember.Login)
```
trotttrotttrott:
```suggestion
		memberSlice = append(memberSlice, teamMember.Email)
```

grafana/resource_team.go
```diff
@@ -0,0 +1,264 @@
...
+}
+
+func CreateTeam(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gapi.Client)
+	name := d.Get("name").(string)
+	email := d.Get("email").(string)
+	teamID, err := client.AddTeam(name, email)
+	if err != nil && err.Error() == "409 Conflict" {
+		return errors.New(fmt.Sprintf("Error: A Grafana Team with the name '%s' already exists.", name))
+	}
```
trotttrotttrott: I think we should remove this. Testing against Grafana 7.2.1, we already get a decent error message from the API:

```
Error: status: 409, body: {"message":"Team name taken"}
```

grafana/resource_team.go
```diff
@@ -0,0 +1,264 @@
...
+		u := change.Member
+		switch change.Type {
+		case AddMember:
+			err = client.AddTeamMember(teamId, u.ID)
+		case RemoveMember:
+			err = client.RemoveMemberFromTeam(teamId, u.ID)
+		}
+		if err != nil && err.Error() != "409 Conflict" {
+			return err
+		}
```
trotttrotttrott: I think we can remove the 409 check. If adding a user that's already a member, we'll get a 400. If removing a member that's not on the team, 404. In any case, if there's error, it seems to me we can return it.

grafana/resource_team.go
```diff
@@ -0,0 +1,264 @@
...
+
+	d.SetId(strconv.FormatInt(teamID, 10))
+	return UpdateMembers(d, meta)
+}
+
+func ReadTeam(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gapi.Client)
+	teamID, _ := strconv.ParseInt(d.Id(), 10, 64)
+	resp, err := client.Team(teamID)
+	if err != nil && err.Error() == "404 Not Found" {
```
trotttrotttrott:
```suggestion
	if err != nil && strings.HasPrefix(err.Error(), "status: 404") {
```

Full error string would be `status: 404, body: {"message":"Team not found"}` because of https://github.com/grafana/grafana-api-golang-client/blob/c87fc20af1eab8c2dce94aa3e07a631508150ce5/client.go#L69. It's unfortunate that we're matching a string pattern here, but it'll have to do for now.

grafana/resource_team.go
```diff
@@ -0,0 +1,264 @@
...
+	client := meta.(*gapi.Client)
+	teamID, _ := strconv.ParseInt(d.Id(), 10, 64)
+	return client.DeleteTeam(teamID)
+}
+
+func ExistsTeam(d *schema.ResourceData, meta interface{}) (bool, error) {
+	client := meta.(*gapi.Client)
+	teamID, _ := strconv.ParseInt(d.Id(), 10, 64)
+	_, err := client.Team(teamID)
+	if err != nil && err.Error() == "404 Not Found" {
```
trotttrotttrott:
```suggestion
	if err != nil && strings.HasPrefix(err.Error(), "status: 404") {
```

website/docs/r/team.html.md
```diff
@@ -0,0 +1,59 @@
...
+```hcl
+# Create a Grafana team with defined membership. The resource
+# requires users to already exist in the system
+resource "grafana_team" "test-team" {
+    name         = "Test Team"
+    email        = "teamemail@example.com"
+    members       = [
+        "viewer-01@example.com"
+    ]
+}
```
trotttrotttrott: Let's format this...

```suggestion
resource "grafana_team" "test-team" {
  name  = "Test Team"
  email = "teamemail@example.com"
  members = [
    "viewer-01@example.com"
  ]
}
```

trotttrotttrott (commented):
@jonathan-dorsey I found a few more things that need attention. Apologies for the back and forth.

jonathan-dorsey:
@trotttrotttrott  - no worries! I appreciate the feedback and these are all good changes. A lot of these were just bad changes I furthered by copying from the org code. I am almost ready to start using the provider to manage a new Grafana instance. During that process I will do a lot more testing of edge scenarios and enter new defects for the org code as well. I would like to add some tests for some of those 404/409 errors, but I think it would require either applying changes to Grafana outside of terraform or manipulating the terraform state within the test and I'm not sure how to do either without some more research. 

At any rate - changes have been incorporated and should be ready for another review. Thanks again!

trotttrotttrott (approved):
LGTM! Thanks @jonathan-dorsey!

PR #125, opened 2020-10-15
grafana_user is importable

trotttrotttrott:
`grafana_user` resource supports [importability](https://www.terraform.io/docs/import/importability.html).

### DIFF
--- a/grafana/resource_user.go
+++ b/grafana/resource_user.go
@@ -1,11 +1,12 @@
 package grafana
 
 import (
+	"errors"
+	"fmt"
 	"strconv"
 
-	"github.com/hashicorp/terraform/helper/schema"
-
 	gapi "github.com/grafana/grafana-api-golang-client"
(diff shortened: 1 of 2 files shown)


macgreagoir (approved):

grafana/resource_user.go
```diff
@@ -101,3 +106,25 @@ func DeleteUser(d *schema.ResourceData, meta interface{}) error {
 	}
 	return client.DeleteUser(id)
 }
+
+func ExistsUser(d *schema.ResourceData, meta interface{}) (bool, error) {
+	client := meta.(*gapi.Client)
+	userId, _ := strconv.ParseInt(d.Id(), 10, 64)
+	_, err := client.User(userId)
```
jvrplmlmn, line 113: Took a look at the Grafana client to see if we could compare against specific UserNotFound errors, but it it seems to be a simple HTTP wrapper, so this is fine 👍

jvrplmlmn (approved):

Issue #51, opened 2019-03-12
[Feature Request] Support teams, admin users, users 


alexgdmx:
In current implementation, it appears terraform-provider-grafana does not support Grafana  teams, admin users, users . It would be great to be able to create them

/api/org/users PATCH to add native admin not role
/api/admin/users
/api/users
/api/teams

peimanja:
Any thoughts on this one? can be really useful for us maintaining teams/users and permissions there

brycefisher:
I'm also pretty interested in this feature. Ultimately, I'd want to be able to set permissions on dashboards by team. That's when this would be super valuable for me.

To me, this is the functionality I'd like to see:

- Create/delete/update teams
- Assign dashboard permissions per team
- Add/remove users from a team

I'd be willing to contribute a PR to that end. @mlclmj would you be interested in receiving a PR (or series of PRs) like this?

mlclmj:
Hey @brycefisher for sure! This sounds like a great addition if you're willing to build it out. A series of smaller PRs is always encouraged since they're easier to review 🙂

I'm guessing this probably requires changes to the `go-grafana-api` library as well--feel free to request a review from me in either and I'll do my best to get to them.

brycefisher:
Yay!
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

tonglil:
Some of the required APIs asked for have been merged in the library if anyone wants to give this a shot.

Folder permissions: https://github.com/nytm/go-grafana-api/pull/36
Team: https://github.com/nytm/go-grafana-api/pull/37

There is also an open but stale PR with the implementation of some other APIs.
As @mlclmj said, smaller PRs are much easier to review if anyone wants to break out those changes and try implementing their corresponding Terraform resources.

trotttrotttrott:
Hi, I'd like to work on this if no one has started yet. @tonglil feel free to assign to me if you'd like.

jonathan-dorsey:
PR https://github.com/grafana/terraform-provider-grafana/pull/120 submitted to add Team/Team Member functionality to the provider.

pascal-hofmann:
PR #150 submitted to add dashboard permission functionality to the provider.

julienduchesne:
Teams and users (admin or not) are fully supported

PR #128, opened 2020-10-18
Adding folder permission functionality

jonathan-dorsey:
This pull request adds the ability to specify folder permissions in the provider. A few things to note here: 
1) Since the folder permissions are tied to folder, you don't really create or delete them. It's always an update. We assume a folder has to exist (and therefore default permissions created) and that deleting the folder will also delete the corresponding folder permissions. The permissions were not added to the existing folder resource since it requires a separate API call and can then lead to partial state problems. 
2) The API call requires that all permissions be sent for every API request. If a permissions is not sent, it assumes it needs to be deleted. It's for this reason that I grouped all permissions into a single resource per folder as opposed to a single resource per permission. 
3) There is minimal testing here. I couldn't figure out how to validate the actual permissions created in the unordered set. Using the standard permissions.#.field notation doesn't work with unordered sets and I could not find another way around it. If someone knows a better option for validating those fields I'm more than happy to update the PR. 

This is in support of feature request https://github.com/grafana/terraform-provider-grafana/issues/51


jonathan-dorsey:
@trotttrotttrott - another PR to add additional resources to the provider. Would appreciate some feedback on the approach and testing comments.

website/docs/r/folder_permission.html.md
```diff
@@ -0,0 +1,66 @@
...
+resource "grafana_user" "user" {
+  email = "user.name@example.com"
+}
+
+resource "grafana_folder" "collection" {
+  title = "Folder Title"
+}
+
+resource "grafana_dashboard_permission" "collectionPermission" {
+  folder_uid = "${grafana_folder.colection.uid}"
```
trotttrotttrott:
```suggestion
  folder_uid = grafana_folder.colection.uid
```

Interpolation-only expressions are deprecated.

website/docs/r/folder_permission.html.md
```diff
@@ -0,0 +1,66 @@
...
+
+resource "grafana_user" "user" {
+  email = "user.name@example.com"
+}
+
+resource "grafana_folder" "collection" {
+  title = "Folder Title"
+}
+
+resource "grafana_dashboard_permission" "collectionPermission" {
```
trotttrotttrott:
```suggestion
resource "grafana_folder_permission" "collectionPermission" {
```

grafana/resource_folder_permission.go
```diff
@@ -0,0 +1,164 @@
...
+
+	return nil
+}
+
+func DeleteFolderPermissions(d *schema.ResourceData, meta interface{}) error {
+	//there is no delete call for folder permissions. we *could* try to delete
+	//all of the individual permissions for the folder, but that would just leave
+	//us with an inoperable folder. however, in cases where a folder is deleted,
+	//we want the corresponding folder permissions to be removed from state without
+	//making a call to the API which would just fail if the folder is gone.
```
trotttrotttrott: I think we do want to remove all permissions. Admins will still be able to access the folder.

image

We'd want to check if the folder exists first. If it's gone, no need to make the API call.

jonathan-dorsey: Yeah this is a good point, but checking if the folder exists is a little more difficult than it should be. The Get/Put for folder permissions are done using the folder uid so that's what I used to tie the permissions to the folder. While the Grafana folder API provides the option to select a folder by UID, the go api does not. So I can switch the permissions to use the folder id and make the uid a computed value, I can make the delete a little more hacky trying to find the folder by uid by selecting all folders and finding the right one, or I can go update the go api to support that extra select by uid call. Any preferences?

jonathan-dorsey: Since the permissions *should* be referencing the folder.folder_uid wouldn't all of the destroys happen in reverse order? So maybe it's ok to remove permissions without checking for the existence? If someone does something really hacky and hard-codes a uid value and the proper destroy order doesn't happen maybe we can ignore a failure because the parent folder doesn't exist?

jonathan-dorsey:
@trotttrotttrott - I went ahead and pushed some changes that just ignore the error if the folder doesn't exist on the delete. It's a wasted API call which is unfortunate, but still more efficient than trying to query for the folder. If you'd prefer I refactor the permissions to make the uid a computed value or make changes to the go api I can still do that. Just let me know what you would like there.

grafana/resource_folder_permission.go
```diff
@@ -0,0 +1,175 @@
...
+	//we will simply remove all permissions, leaving a folder that only an admin can access.
+	//if for some reason the parent folder doesn't exist, we'll just ignore the error
+	client := meta.(*gapi.Client)
+
+	folderUID := d.Get("folder_uid").(string)
+	emptyPermissions := gapi.PermissionItems{}
+
+	err := client.UpdateFolderPermissions(folderUID, &emptyPermissions)
+	if err != nil {
+		if err.Error() == "404 Not Found" {
```
trotttrotttrott:
```suggestion
		if strings.HasPrefix(err.Error(), "status: 404") {
```

trotttrotttrott (commented):
Your delete solution looks good to me. The 404 check just needs updating.

grafana/resource_folder_permission.go
jonathan-dorsey: Doh! Good catch... I keep copying that from other places. Went ahead and updated the resource_folder.go since that's where I copied it from.

trotttrotttrott (approved):
Thanks @jonathan-dorsey!

PR #130, opened 2020-10-20
Apply changes to match the change to a Config struct

medains:
Changes occurred in grafana-api-golang-client - changes here to avoid breaking.

### DIFF
--- a/grafana/provider.go
+++ b/grafana/provider.go
@@ -1,9 +1,13 @@
 package grafana
 
 import (
+	"strings"
+	"net/url"
+
 	"github.com/hashicorp/terraform/helper/logging"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
+	"github.com/hashicorp/go-cleanhttp"
 
 	gapi "github.com/grafana/grafana-api-golang-client"
 )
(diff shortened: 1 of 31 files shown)


medains:
This is an illustrative PR.

I think that the changes to grafana/grafana-api-golang-client merit changes to the provider (split auth in the schema to user/pass at least)

Discuss.... ;)

grafana/provider.go
```diff
@@ -41,15 +45,19 @@ func Provider() terraform.ResourceProvider {
...
+	cli := cleanhttp.DefaultClient()
+	cli.Transport = logging.NewTransport("Grafana", cli.Transport)
 	client, err := gapi.New(
-		d.Get("auth").(string),
 		d.Get("url").(string),
+		gapi.Config{
+			BasicAuth: url.UserPassword(splitAuth[0], splitAuth[1]),
+			Client: cli,
+		},
 	)
```
trotttrotttrott: This doesn't account for the possibility that `auth` is an API key. What do you think about something like this instead...

```golang
	auth := strings.SplitN(d.Get("auth").(string), ":", 2)
	cli := cleanhttp.DefaultClient()
	cli.Transport = logging.NewTransport("Grafana", cli.Transport)
	cfg := gapi.Config{
		Client: cli,
	}
	if len(auth) == 2 {
		cfg.BasicAuth = url.UserPassword(auth[0], auth[1])
	} else {
		cfg.APIKey = auth[0]
	}
	client, err := gapi.New(d.Get("url").(string), cfg)
```

trotttrotttrott (commented):
This is great, thanks for working on it. Just one issue I noticed about using API keys instead of basic auth.

trotttrotttrott (commented):
Also, please run `make fmt && go mod vendor` and commit the changes.

medains:
Suggested changes made

trotttrotttrott (approved):
Excellent. Thanks @medains!

PR #134, opened 2020-11-02
Add support for providing additional HTTP headers

gw0:
Sometimes the Grafana server is behind a proxy or application firewall and it is necessary to provide additional HTTP headers to establish a connection, such as authentication tokens or cookies. In general sense these are represented as key-value pairs of strings. To also add support to provide these using an environment variable `GRAFANA_HTTP_HEADERS`, this PR assumes it contains a JSON-encoded map.

Usage:

```
provider "grafana" {
  url  = "http://172.17.0.1:3000"
  auth = "admin:admin"
  headers = {
    "Cookies" = "foo=bar"
  }
}
```

```bash
$ GRAFANA_HTTP_HEADERS='{"Cookies":"foo=bar"}' terraform plan
```

This PR is based on https://github.com/grafana/terraform-provider-grafana/pull/119 and depends on PR https://github.com/grafana/grafana-api-golang-client/pull/16.

### DIFF
--- a/grafana/provider.go
+++ b/grafana/provider.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net/url"
+	"os"
 	"regexp"
 	"strings"
 
@@ -59,6 +61,13 @@ func Provider(version string) func() *schema.Provider {
 					DefaultFunc:  schema.EnvDefaultFunc("GRAFANA_AUTH", nil),
 					Description:  "API token or basic auth username:password. May alternatively be set via the `GRAFANA_AUTH` environment variable.",
 					AtLeastOneOf: []string{"auth", "cloud_api_key", "sm_access_token"},
+				},
+				"http_headers": {
+					Type:        schema.TypeMap,
+					Optional:    true,
+					Sensitive:   true,
+					DefaultFunc: JSONEnvDefaultFunc("GRAFANA_HTTP_HEADERS", nil),
+					Description: "Optional. HTTP headers mapping keys to values used for accessing the Grafana API. May alternatively be set via the `GRAFANA_HTTP_HEADERS` environment variable in JSON format.",
 				},
 				"retries": {
 					Type:        schema.TypeInt,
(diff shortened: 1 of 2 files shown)


grafana/provider.go
```diff
@@ -28,6 +30,14 @@ func Provider() terraform.ResourceProvider {
 				DefaultFunc: schema.EnvDefaultFunc("GRAFANA_AUTH", nil),
 				Description: "Credentials for accessing the Grafana API.",
 			},
+			"headers": {
```
inkel: `headers` is too generic, can you rename it to `http_headers`? It also maps better to the environment variable name.

grafana/provider.go
```diff
@@ -58,10 +68,46 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	} else {
 		cfg.APIKey = auth[0]
 	}
+
+	headersObj := d.Get("headers").(map[string]interface{})
+	if headersObj != nil && len(headersObj) == 0 {
+		// Workaround for a bug when DefaultFunc returns a TypeMap
+		headersObjAbs, _ := JsonEnvDefaultFunc("GRAFANA_HTTP_HEADERS", nil)()
```
inkel: Why are you deciding to ignore the error here? If `GRAFANA_HTTP_HEADERS` contains invalid JSON no error will be reported to the user but the provider will continue to work, so the end user won't know that their configuration was ignored.

grafana/provider.go
```diff
@@ -58,10 +68,46 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	} else {
 		cfg.APIKey = auth[0]
 	}
+
+	headersObj := d.Get("headers").(map[string]interface{})
+	if headersObj != nil && len(headersObj) == 0 {
```
inkel: This can be simplified, as a `nil` map will always have a `len` of 0:
```suggestion
	if len(headersObj) == 0 {
```

grafana/provider.go
```diff
@@ -58,10 +68,46 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
...
+		headersObj = headersObjAbs.(map[string]interface{})
+	}
+	if headersObj != nil {
+		// Convert headers from map[string]interface{} to map[string]string
+		headers := make(map[string]string)
+		for k, v := range headersObj {
+			switch v := v.(type) {
+			case string:
+				headers[k] = v
+			}
```
inkel: For a single case type conversion, it's better to use a [type assertion](https://go.dev/ref/spec#Type_assertions) instead:
```suggestion
			if v, ok := v.(string); ok {
				headers[k] = v
			}
```

inkel (changes requested):
Thanks @gw0 for your contribution! Overall I like the idea, although I've left some comments.

grafana/provider.go
gw0: Changed.

grafana/provider.go
gw0: You are right, returning the error in this case.

grafana/provider.go
gw0: Simplified.

grafana/provider.go
gw0: Simplified.

gw0:
@inkel Thank you for your feedback. I rebased on top of master and included your suggested changes.

grafana/provider.go
gw0: Reverted, because it results in:

```
panic: interface conversion: interface {} is nil, not map[string]interface {} [recovered]
	panic: interface conversion: interface {} is nil, not map[string]interface {}
```

gw0:
Fixed issues discovered by CI.

julienduchesne (approved):
LGTM!

inkel:
Thank you @gw0 for the changes!

PR #137, opened 2020-11-06
Use UID to maintain dashboard state instead of slug

### LINKED ISSUE #95: [Feature request] Have Grafana dashboards be stored by uid instead of slug
Hi there,

Currently Grafana dashboards are stored by their slug instead of uid using one of the newer versions of Grafana. This causes failures with having multiple dashboards being named the same under different folders in the same Grafana instance. With changing to uid for dashboards, this should resolve the issue. 

### Affected Resource(s)
- resource_dashboard

volgorean:
## Changes
Swaps out UID instead of slug as the primary ID for managing the `dashboards` resource state fixing slug collision issues when having multiple dashboards with the same name (ie 'overview' in multiple folders.) Closes #95.

## Tests
* AC tests pass
* Manually tested full life-cycle in Grafana 7.2.1 to verify this fixes slug collision issues.

### DIFF
--- a/grafana/resource_dashboard.go
+++ b/grafana/resource_dashboard.go
@@ -22,11 +22,6 @@ func ResourceDashboard() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
-			"slug": {
-				Type:     schema.TypeString,
-				Computed: true,
-			},
-
 			"dashboard_id": {
 				Type:     schema.TypeInt,
 				Computed: true,
@@ -62,7 +57,7 @@ func CreateDashboard(d *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
-	d.SetId(resp.Slug)
+	d.SetId(resp.UID)
 
 	return ReadDashboard(d, meta)
 }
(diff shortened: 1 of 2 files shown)


emedvedev:
Any chance this can get merged? We're having problems because we have dashboards with the same name in different folders and the provider can't remove them properly.

volgorean:
If there is anything else I can do on this front I'd be happy to do it!

trotttrotttrott:
@volgorean I agree with the change, but it will make the provider want to recreate dashboards already in state. Attempting to recreate them will fail which will mean users will have to import all dashboards. This could be a big job for some folks. Can we instead deprecate slug and fallback if necessary?

volgorean:
@trotttrotttrott I can give that a try, _might_ not get to it until the weekend though.

matthiasng:
@volgorean have a look at: https://github.com/grafana/terraform-provider-grafana/pull/84/files#diff-d3b5fa10be0054bad74d5b2028ea65350bbdb16aad37b35f78d87625b8d23056R210

The err message for sure is a bit outdated, but the rest sould be fine.

trotttrotttrott:
Closing in favor of https://github.com/grafana/terraform-provider-grafana/pull/217

PR #155, opened 2021-01-08
Add full Stackdriver data source support

### LINKED ISSUE #105: [enhancement] resource/datasource: Incomplete Stackdriver Support
@ryanaross raised in #91:

> I attempted to use a build of master but it still seems to be missing some things that would make it required to work.
https://grafana.com/docs/grafana/latest/features/datasources/stackdriver/#configure-the-data-source-with-provisioning
It seems like the provider would also need to pass defaultProject, clientEmail, etc fields as well. I have the private_key being passed now successfully but in the Grafana interface it looks unconfigured which is not how it looks when the key is added properly.

ryanaross:
## What
Adds 4 new fields to JsonData in the data source management.

## Why
Closes issue #105 . As the issue states, the Cloud Monitoring/Stackdriver data source needs more fields in the jsonData portion in order to be configured and work.

Grafana Documentation:
https://grafana.com/docs/grafana/latest/datasources/cloudmonitoring/#configure-the-data-source-with-provisioning

## Verification
`make test` succeeds. I also compiled using `make build`, and used the new binary to create a new Stackdriver data source successfully in a local Grafana instance.

### DIFF
--- a/grafana/resource_data_source.go
+++ b/grafana/resource_data_source.go
@@ -62,11 +62,23 @@ func ResourceDataSource() *schema.Resource {
 							Type:     schema.TypeString,
 							Optional: true,
 						},
+						"authentication_type": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"client_email": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
 						"conn_max_lifetime": {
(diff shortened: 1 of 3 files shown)


trotttrotttrott (approved):
Looks good @ryanaross. Thanks for tracking this for so long 👍

PR #157, opened 2021-01-22
is_admin added to grafana_user.

AlirezaKm:
### What
Added 1 new field (is_admin) to grafana_user to create grafana admin user.

### Why ?
Create admin user in the default organization.

### Verification
* `make testacc` passed.
* Tests in Grafana 7.3.6.

### DIFF
--- a/grafana/resource_user.go
+++ b/grafana/resource_user.go
@@ -37,6 +37,11 @@ func ResourceUser() *schema.Resource {
 				Required:  true,
 				Sensitive: true,
 			},
+			"is_admin": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


trotttrotttrott (approved):
Looks good, thanks!

caoimheharvey:
Hi, I'm having issues when using this, I'm able to create the user, but when `is_admin` is set to `true` it seems to still create it as a `Viewer`. I'm not able to spot anything useful in the debug logs during the terraform apply to help explain this. 
I tried also manually updating the test user to `Admin` then setting `is_admin` to false for the user that was created and it also didn't revert the permissions on the Grafana dashboard to `Viewer`. 
I'm looking at the User struct here and it looks like by default the `isAdmin` is omitted if not provided explicitly. 
https://github.com/grafana/grafana-api-golang-client/blob/master/user.go#L20
My grasp on GoLang is a bit iffy so perhaps I'm looking at the wrong thing, but any help would be appreciated. 
https://github.com/grafana/terraform-provider-grafana/pull/157/commits/93dbbf471bf0f4c5b5039437aa2ae060047b0772#diff-bb873edc41663d158ebb4afe9130d1840ff2575b76f90d21eee7498c1c2a2ab4R55
```
user := gapi.User{
		Email:    d.Get("email").(string),
		Name:     d.Get("name").(string),
		Login:    d.Get("login").(string),
		Password: d.Get("password").(string),
	}
```
As far as I can tell, the code in the PR seems to update the resource in the Terraform state but I'm not seeing where its updating on the Grafana instance itself, maybe I'm not looking in the right place. 
I can raise a PR to resolve this if this is in fact the issue :)

PR #165, opened 2021-02-13
Allow dependabot to check go modules

jlosito:
This allows dependabot to check any go dependencies which this project uses on a weekly basis and submit pull requests with version bumps in order to keep packages up-to-date.

https://docs.github.com/en/github/administering-a-repository/keeping-your-actions-up-to-date-with-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+---
+version: 2
+updates:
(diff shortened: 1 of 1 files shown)


trotttrotttrott (approved):
Sounds good to me. Thanks, @jlosito.

PR #171, opened 2021-02-20
Allow dependabot to check github actions

jlosito:
This change allows dependabot to check any GitHub actions which this project uses and submit pull requests with version bumps in order to keep packages up-to-date.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,11 @@
 ---
 version: 2
 updates:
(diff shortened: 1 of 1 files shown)


trotttrotttrott (approved):
Thanks! Anticipating most automation to happen with Drone, but we'll at least have the release workflow around since that's the [recommended approach](https://www.terraform.io/docs/registry/providers/publishing.html#github-actions-preferred-).

PR #176, opened 2021-03-09
Upgrade to Terraform Plugin SDK v2

trotttrotttrott:
Minimal changes to migrate to [SDK v2](https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html). Closes https://github.com/grafana/terraform-provider-grafana/issues/159. Additional SDK v2 feature integrations will be done in a separate changeset (https://github.com/grafana/terraform-provider-grafana/issues/175).

[tf-sdk-migrator](https://github.com/hashicorp/tf-sdk-migrator) was primary used for code changes. Minor changes had to be made manually to fix acceptance tests.

**[Support for Terraform 0.11 and Below is being dropped](https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html#dropped-support-for-terraform-0-11-and-below)**. The readme now includes a note about a Terraform 0.12+ requirement.

This will not incur breaking changes for existing codebases (assuming use of Terraform 0.12+).

PR #177, opened 2021-03-10
Remove string interpolation in docs

julienduchesne:
Following https://github.com/grafana/terraform-provider-grafana/pull/176

Small thing I noticed while checking out the docs
Since the plugin now requires 0.12, the string interpolation can be updated
Other plugins such as the AWS one have updated their docs that way too


trotttrotttrott (approved):
Cool, thanks @julienduchesne 👍

PR #187, opened 2021-04-26
Migrate Legacy Documentation to tfplugindocs

trotttrotttrott:
Managing documentation in `website/` is now considered legacy. This PR migrates to using [tfplugindocs](https://github.com/hashicorp/terraform-plugin-docs) for [generating documentation](https://www.terraform.io/docs/registry/providers/docs.html#generating-documentation).

There is minimal actual content change. This approach just migrates us off of the legacy directory structure and will make documentation maintenance easier:

* Schema descriptions are managed in the actual schema definitions. Before the descriptions were static markdown which makes it difficult to ensure parity with the actual schema.
* Examples are in standalone `.tf` files. This allows us to use `terraform fmt` to ensure formatting is correct and reuse them in tests to ensure they actually work (this will hopefully come later).
* Consistency from templating.


julienduchesne (approved):
That is very cool! TIL about that tool

PR #207, opened 2021-05-31
Add resources for managing fine-grained access control in Grafana Enterprise

vtorosyan:
## Context

Grafana Enterprise 8.0 introduces [fine-grained access control](https://grafana.com/docs/grafana/next/enterprise/access-control/) which allows manage access of users with fine-grained permissions. 

For managing fine-grained roles, an [HTTP API](https://grafana.com/docs/grafana/next/http_api/access_control/) is available and we would like to leverage that fact and allow users to provision custom roles, grant or revoke access using terraform.

## Scope of the PR

PR introduces two new resources: 
1. `grafana_role` - which allows to create, update or delete [fine-grained custom role](https://grafana.com/docs/grafana/next/enterprise/access-control/roles/#custom-roles)
2. `grafana_builtin_role_assignment` which allows to create or remove [built-in role assignments](https://grafana.com/docs/grafana/next/enterprise/access-control/roles/#built-in-role-assignments)

Fixes https://github.com/grafana/grafana-enterprise/issues/1055

## Other notes

- The fine-grained access control is only available in the Grafana Enterprise and is in Beta at this moment. This is reflected in the resource documentation and link is provided to the official documentation.
- Since fine-grained access control is only in the Grafana Enterprise, we can't use the same image to run acceptance tests. Moreover, Grafana Enterprise requires a license to run and the new endpoints are only available with valid license. Thus, running acceptance tests in the container is challenging. I've added a new build tag for running enterprise relevant tests on local dev environment. Note that this comes with a trade-off that we won't run these tests in the CI, so I'd appreciate suggestions for better setup.


vtorosyan:
We would need to merge the PR on the client side https://github.com/grafana/grafana-api-golang-client/pull/26 and update the dependency, before running the CI.

grafana/provider.go
```diff
@@ -77,16 +77,18 @@ func Provider() *schema.Provider {
...
+			"grafana_dashboard_permission":    ResourceDashboardPermission(),
+			"grafana_data_source":             ResourceDataSource(),
+			"grafana_folder":                  ResourceFolder(),
+			"grafana_folder_permission":       ResourceFolderPermission(),
+			"grafana_organization":            ResourceOrganization(),
+			"grafana_team":                    ResourceTeam(),
+			"grafana_team_preferences":        ResourceTeamPreferences(),
+			"grafana_user":                    ResourceUser(),
+			"grafana_role":                    ResourceRole(),
+			"grafana_builtin_role_assignment": ResourceBuiltInRoleAssignment(),
```
trotttrotttrott: Can we list these alphabetically?

GNUmakefile
```diff
@@ -3,6 +3,9 @@ GRAFANA_VERSION ?= latest
 testacc:
 	TF_ACC=1 go test ./... -v $(TESTARGS) -timeout 120m
 
+testacc-enterprise: testacc
+	TF_ACC=1 go test ./... -tags="enterprise" -v $(TESTARGS) -timeout 120m
```
trotttrotttrott, line 7: Great idea using a build tag. It'd be nice to add this to CI eventually though too. Right now we run the acceptance tests without enterprise enabled. Not sure how hard it would be to get an enterprise enabled container in our Drone build.

trotttrotttrott (commented):
Looks great. Just a small nit. I'd say it's ready to ship once the vendoring is updated 👍

trotttrotttrott (approved):

PR #214, opened 2021-06-11
Allow to overwrite while creating dashboard from terraform

szemek:
When dashboard already exists, creating it from terraform returns following error:

```
Error: status: 412, body: {"message":"The dashboard has been changed by someone else","status":"version-mismatch"}
```

I'd like to have a possibility to overwrite dashboards without importing them one by one.

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -16,6 +16,7 @@ local pipeline(name, trigger) = {
       name: 'tests',
       image: golang,
       commands: [
+        'sleep 5', // https://docs.drone.io/pipeline/docker/syntax/services/#initialization
         'make testacc',
       ],
       environment: {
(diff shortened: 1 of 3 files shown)


trotttrotttrott (approved):
Makes sense. Thank you 👍

And thanks for adding the sleep command in Drone. We've had a few failures recently because of tests starting before the Grafana service was up.

wout-konings:
I am still running into some issues with this. 

I have the following:

`resource "grafana_dashboard" "dashboards" {
  for_each    = fileset(path.module, "../dashboards/*.json")
  config_json = file("${path.module}/${each.key}")
  folder = try(grafana_folder.folder[regex("${var.folder_regex}", "${each.key}")].id, 0)
  overwrite = true
}`

Which worked without the overwrite. However, I am getting the following error:

╷
│ Error: Unsupported argument
│ 
│   on main.tf line 39, in resource "grafana_dashboard" "dashboards":
│   39:   overwrite = true
│ 
│ An argument named "overwrite" is not expected here.
╵

szemek:
@wout-konings you need to upgrade to version v1.12.0 or newer
https://github.com/grafana/terraform-provider-grafana/blob/master/CHANGELOG.md

PR #213, opened 2021-06-11
Prepare For Grafana Cloud Resources

trotttrotttrott:
Refactors the provider metadata interface to be a new type with fields for multiple API clients. Fields include `gapi` for the Grafana API client and `smapi` for the [Synthetic Monitoring API client](https://github.com/grafana/synthetic-monitoring-api-go-client). This as well updates vendoring to include it. No Synthetic Monitoring resources are added yet as to reduce PR size - those will be a followup.

In addition, I've added some SDK v2 refactoring I incorporated while working on cloud resources:

* The provider is made to be [debuggable](https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html#support-for-debuggable-provider-binaries). The relevant code for this can be found in `main.go`.
* Implements [new acceptance test driver](https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html#a-new-acceptance-testing-driver) with ProviderFactories. This eliminates our use of the deprecated field [`resource.TestCase. Providers`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.6.1/helper/resource#TestCase)

PR #215, opened 2021-06-15
Remove Vendor Directory

trotttrotttrott:
Managing the vendor directory has been cumbersome. It results in unnecessarily large changesets in pull requests (https://github.com/grafana/terraform-provider-grafana/pull/213 for example) and creates extra work for contributors.We should be able to reliably build dependencies with Go modules. Also looking at other major providers, it is not common to commit vendor.


inkel (approved):
Thank you!!!

Issue #199, opened 2021-05-17
Increase Statefile Size 


sham-github:
Hi Team, 

we are using Grafana provider to create our dashboards but getting below error after creating a significant number of dashboards. 

**Error: rpc error: code = ResourceExhausted... 
Error: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (4307026 vs. 4194304)**

we are using Azure RM as backend, and we don't see any compress feature for terraform statefile in our case of Azure which is there for AWS/Consul. 

This is a must needed feature since Grafana dashboards are supposed to be large and if we are managing by terraform then the provider also should work accordingly.

julienduchesne:
I don't believe we can increase the statefile size. That `ResourceExhausted` error is probably an internal Terraform error from communicating between the main CLI and the plugin (provider)

However, a `store_dashboard_sha256` provider option was added in v1.20.0 that, if toggled, will only save a sha256 of the dashboard's content to state instead of the whole json. I will therefore consider this issue completed

PR #218, opened 2021-06-17
Save dashboard md5sum instead of full config json in tfstate

fgouteroux:
Hi,

In some cases we could have bigger grafana dashboard, that would cause:
- the unability to review changes inside (even in git when we have thousand of lines)
- a bigger statefile size and some errors like https://github.com/grafana/terraform-provider-grafana/issues/199

I would like to add the possibility to save only the md5sum of the config json, if the option `config_json_md5` is set to `true`.

Also notice that we can go back to the full config json if we disable this parameter by default to `false`, so there is no breaking change.

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -30,6 +30,10 @@ local pipeline(name, trigger) = {
     {
       name: 'grafana',
       image: grafana,
+      environment: {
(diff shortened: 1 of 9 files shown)


fgouteroux:
Better with sha256

Issue #206, opened 2021-05-27
New feature: team sync/external group resource


ekpdt:
Suggesting a new feature to support managing team sync via Terraform.

The existing public API suffices: https://grafana.com/docs/grafana/latest/http_api/external_group_sync/

Proposed example resource block:
```
resource "grafana_external_group" "fooExternalGroup" {
  team_id = grafana_team.foo.id
  groups = [ "cn=editors,ou=groups,dc=grafana,dc=org", "cn=viewers,ou=groups,dc=grafana,dc=org"]
}
```
Or alternatively an external group resource for each group.

ekpdt:
Closed by #221

PR #221, opened 2021-06-18
add support of team external group

fgouteroux:
Hello,

I'm using another custom provider to manage grafana resources with TF. 
Since I see some updates on this provider, I'm interested to use it. 

Also as I use some grafana enterprise features, like the team sync, so I want to add the support of team external group (I see the issue https://github.com/grafana/terraform-provider-grafana/issues/206)

Need the PR https://github.com/grafana/grafana-api-golang-client/pull/29.


grafana/resource_team_external_group.go
```diff
@@ -0,0 +1,165 @@
...
+				Description: "The Team ID",
+			},
+
+			"groups": {
+				Type:     schema.TypeSet,
+				Required: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				Description: `The team external groups list`,
```
inkel: Nit: no need to use backticks for this string, please use just regular `"`.

grafana/resource_team_external_group.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func ReadGroups(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*client).gapi
+	teamID, _ := strconv.ParseInt(d.Id(), 10, 64)
+	teamGroups, err := client.TeamGroups(teamID)
+	if err != nil {
+		return err
+	}
+	groupSlice := []string{}
```
inkel: Can we rename this to `groupIDs` instead?

While at it, consider initialising this variable as `groupIDs := make([]string, 0, len(teamGroups)` to reduce the number of allocations that are needed.

grafana/resource_team_external_group.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func groupChanges(d *schema.ResourceData) ([]string, []string) {
+	// Get the lists of team groups read in from Grafana state (old) and configured (new)
+	state, config := d.GetChange("groups")
+
+	stateGroups := make([]string, state.(*schema.Set).Len())
+	for _, v := range state.(*schema.Set).List() {
+		stateGroups = append(stateGroups, v.(string))
+	}
```
inkel: Here and below when you declare the slice it will already have `len` empty elements, so when you append to them you will end up with twice the size of the slice, as shown in [here](https://play.golang.org/p/2tuFhXP7Vni).

There are two ways you can fix this: if you want to use append, then initialise the slice as `stateGroups := make([]string, 0, state.(*schema.Set).Len())`, so it has the right capacity but no elements assigned. Or range with an index and access the destination elements in the slice directly:

```suggestion
	stateGroups := make([]string, state.(*schema.Set).Len())
	for i, v := range state.(*schema.Set).List() {
		stateGroups[i] = v.(string)
	}
```

grafana/resource_team_external_group.go
```diff
@@ -0,0 +1,165 @@
...
+			removeGroups = append(removeGroups, group)
+		}
+	}
+
+	return addGroups, removeGroups
+}
+
+// SliceFind takes a slice and looks for an element in it. If found it will
+// return it's key, otherwise it will return -1 and a bool of false.
+func SliceFind(slice []string, val string) (int, bool) {
```
inkel: You don't seem to ever be using the returned `int`, consider making this simple returning a `bool` instead.

inkel (changes requested):

fgouteroux:
Thanks for your review, I applied your requested changes

inkel:
@fgouteroux thanks for the updates! Please note that CI is failing. I've tried the changes locally and was able to reproduce:

```
$ go build
# github.com/grafana/terraform-provider-grafana/grafana
grafana/resource_team_external_group.go:83:27: client.TeamGroups undefined (type *gapi.Client has no field or method TeamGroups)
grafana/resource_team_external_group.go:104:16: client.NewTeamGroup undefined (type *gapi.Client has no field or method NewTeamGroup)
grafana/resource_team_external_group.go:111:16: client.DeleteTeamGroup undefined (type *gapi.Client has no field or method DeleteTeamGroup)
```

Perhaps you're missing updating a dependency?

fgouteroux:
Yes this PR require the merge of https://github.com/grafana/grafana-api-golang-client/pull/29 and the update of https://github.com/grafana/terraform-provider-grafana/blob/master/go.mod#L6

grafana/resource_team_external_group_test.go
```diff
@@ -0,0 +1,62 @@
+package grafana
```
trotttrotttrott, line 3:
```suggestion
// +build enterprise

package grafana
```

We've started adding this build flag to enterprise resources. We don't yet have CI running against an enterprise instance so tests will fail.

grafana/resource_team_external_group_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"testing"
+
+	gapi "github.com/grafana/grafana-api-golang-client"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccTeam_External_Groups(t *testing.T) {
```
trotttrotttrott:
```suggestion
func TestAccTeamExternalGroups(t *testing.T) {
```

Minor nit, but I don't think the underscores are necessary.

grafana/resource_team_external_group.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	return diag.Diagnostics{}
+}
+
+func ReadTeamExternalGroup(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	if err := ReadGroups(d, meta); err != nil {
+		return diag.FromErr(err)
+	}
+	return nil
```
trotttrotttrott:
```suggestion
	return diag.Diagnostics{}
```
Trivial, but I suppose for the sake of consistency.

grafana/resource_team_external_group.go
```diff
@@ -0,0 +1,164 @@
...
+		if item == val {
+			return true
+		}
+	}
+	return false
+}
+
+func RemoveIndex(s []string, index int) []string {
+	return append(s[:index], s[index+1:]...)
+}
```
trotttrotttrott: This doesn't seem to be called. Can we remove it?

grafana/resource_team_external_group.go
```diff
@@ -0,0 +1,164 @@
...
+	d.SetId(strconv.FormatInt(int64(teamID), 10))
+	if err := UpdateGroups(d, meta); err != nil {
+		return diag.FromErr(err)
+	}
+
+	return diag.Diagnostics{}
+}
+
+func ReadTeamExternalGroup(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	if err := ReadGroups(d, meta); err != nil {
```
trotttrotttrott: This is the only time we call the `ReadGroups` function. It seems to me it would be better to move its logic here and collapse it into a single function.

fgouteroux:
@inkel CI is failing but I'm unable to reproduce the error. In my environment, it's working. 
Could you have more details about this failure ?

````
$ TF_ACC=1 go test ./grafana -run TestAccTeamExternalGroup_basic -v
=== RUN   TestAccTeamExternalGroup_basic
--- PASS: TestAccTeamExternalGroup_basic (5.17s)
PASS
ok  	github.com/grafana/terraform-provider-grafana/grafana	5.185s
```

grafana/resource_team_external_group.go
```diff
@@ -0,0 +1,165 @@
...
+	groupIDs := make([]string, 0, len(teamGroups))
+	for _, teamGroup := range teamGroups {
+		groupIDs = append(groupIDs, teamGroup.GroupID)
+	}
+	d.Set("groups", groupIDs)
+
+	return nil
+}
+
+func UpdateGroups(d *schema.ResourceData, meta interface{}) error {
```
trotttrotttrott: Seems to me this logic can move to `UpdateTeamExternalGroup`. It's pretty confusing having both `UpdateGroups` and `UpdateTeamExternalGroup`. And it appears you don't really need `DeleteTeamExternalGroup` since it's the same as `UpdateTeamExternalGroup`. You can use `UpdateTeamExternalGroup` for both `UpdateContext` and `DeleteContext`. The only difference it the `d.SetId("")` part, but you don't actually need it because its called automatically assuming there's no errors. It's often included only for explicitness.

grafana/resource_team_external_group.go
```diff
@@ -0,0 +1,165 @@
...
+// SliceFind takes a slice and looks for an element in it. If found it will
+// return true, otherwise false.
+func SliceFind(slice []string, val string) bool {
+	for _, item := range slice {
+		if item == val {
+			return true
+		}
+	}
+	return false
+}
```
trotttrotttrott: I'm not thrilled with having helper/convenience functions hiding in resources like `SliceFind` is here. What do you think about using it as a variable within this function? cc/ @inkel

```suggestion
	contains := func(slice []string, val string) bool {
		for _, item := range slice {
			if item == val {
				return true
			}
		}
		return false
	}

	addGroups := []string{}
	for _, group := range configGroups {
		if !contains(stateGroups, group) {
			addGroups = append(addGroups, group)
		}
	}

	removeGroups := []string{}
	for _, group := range stateGroups {
		if !contains(configGroups, group) {
			removeGroups = append(removeGroups, group)
		}
	}

	return addGroups, removeGroups
}
```

trotttrotttrott (changes requested):

grafana/resource_team_external_group.go
inkel: I definitely agree with that sentiment. I also found the `stateGroups` and `configGroups` variable names not very clear, but I haven't come up yet with better names.

fgouteroux: I agree, also I renamed variables names to currentGroups and desiredGroups.

grafana/resource_team_external_group.go
fgouteroux: I agree but the `UpdateGroups` function is also called in `CreateTeamExternalGroup`, so I renamed it to `ManageGroups`.

fgouteroux:
Thanks for your review @trotttrotttrott, I applied your suggested changes. Let me know if something else needed.

grafana/resource_team_external_group.go
```diff
@@ -0,0 +1,141 @@
...
+
+func UpdateTeamExternalGroup(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	if err := ManageGroups(d, meta); err != nil {
+		return diag.FromErr(err)
+	}
+
+	return diag.Diagnostics{}
+}
+
+func ManageGroups(d *schema.ResourceData, meta interface{}) error {
```
trotttrotttrott:
```suggestion
func manageTeamExternalGroup(d *schema.ResourceData, meta interface{}) error {
```

Sort of a nitpick, but since the entire provider is in the same package there tends to be pretty verbose function names. And they typically include the entire resource or data source name to avoid ambiguity.

trotttrotttrott: Another thing to note is that none of the function names in the provider really need to begin with a capital letter as if to suggest they should be exported. It's sort of an inherited convention to only start a function name with a capital letter if it's referenced directly by the resource (`CreateContext`, `DeleteContext`, etc).

grafana/resource_team_external_group.go
```diff
@@ -0,0 +1,141 @@
...
+		err := client.DeleteTeamGroup(teamID, group)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func groupChanges(d *schema.ResourceData) ([]string, []string) {
```
trotttrotttrott:
```suggestion
func groupChangesTeamExternalGroup(d *schema.ResourceData) ([]string, []string) {
```

trotttrotttrott (commented):
Sorry for all the back and forth, but I think this is all I've got left. Just some nitpicks about function names.

fgouteroux:
@trotttrotttrott thanks for the explanations

inkel (approved):
LGTM! Thank you @fgouteroux for your contributions.

I'll leave it up to @trotttrotttrott for a final review.

trotttrotttrott (approved):
Thanks @fgouteroux!

PR #222, opened 2021-06-18
Save dashboard sha256sum instead of full config json in tfstate

fgouteroux:
Hi,

In some cases we could have bigger grafana dashboard, that would cause:
- the unability to review changes inside (even in git when we have thousand of lines)
- a bigger statefile size and some errors like #199

I would like to add the possibility to save only the sha256sum of the config json, if the option config_json_sha256 is set to true.

Also notice that we can go back to the full config json if we disable this parameter by default to false, so there is no breaking change.


fgouteroux:
no feedback on it ?

justinTM:
oh this would be huge. thank you. since GitLab jobs have 4MB printout limit, i've just been redirecting the entire Apply to a file because we're at 300k lines of JSON and this would be much better

grafana/resource_dashboard.go
```diff
@@ -301,5 +319,11 @@ func normalizeDashboardConfigJSON(config interface{}) string {
 	delete(dashboardJSON, "id")
 	delete(dashboardJSON, "version")
 	j, _ := json.Marshal(dashboardJSON)
-	return string(j)
+	sha256_store := os.Getenv("GRAFANA_CONFIG_JSON_SHA256")
+	if sha256_store == "yes" {
+		configHash := sha256.Sum256([]byte(j))
```
inkel: `j` already is a `[]byte`:
```suggestion
		configHash := sha256.Sum256(j)
```

inkel (commented):
As far as the changes go it LGTM. Let me discuss it with the rest of the @grafana/terraform-provider team.

julienduchesne (dismissed):
Sorry for the delay, I actually thought I had commented this but I never clicked the button 😓 

I don't think setting an env variable (the `os.Setenv`) and using it in the code below is the right approach. Here's what I think should be done:

1. With a `DefaultFunc`, if `GRAFANA_CONFIG_JSON_SHA256` is set to a truthy value (using `strconv.ParseBool`), then the default is true, otherwise it's false
2. In the code, only use the value of `config_json_sha256`, at this point, the env variable is irrelevant, it serves as a way to set this globally, but it shouldn't be used in the resources themselves

What this allows is to be able to set this to false and have it true in a single resource or vice versa

fgouteroux:
> Sorry for the delay, I actually thought I had commented this but I never clicked the button 😓
> 
> I don't think setting an env variable (the `os.Setenv`) and using it in the code below is the right approach. Here's what I think should be done:
> 
> 1. With a `DefaultFunc`, if `GRAFANA_CONFIG_JSON_SHA256` is set to a truthy value (using `strconv.ParseBool`), then the default is true, otherwise it's false
> 2. In the code, only use the value of `config_json_sha256`, at this point, the env variable is irrelevant, it serves as a way to set this globally, but it shouldn't be used in the resources themselves
> 
> What this allows is to be able to set this to false and have it true in a single resource or vice versa

@julienduchesne  Thanks for the review. I agree with you that setting env var should be not used in the code but this env var is used in the StateFunc: normalizeDashboardConfigJSON  of config_json input. I don't know how to get the value of config_json_sha256 input in this StateFunc.

julienduchesne:
Good point. It doesn't give you the resource data. Then there's no way to handle it on a resource basis so I'd make this a provider setting since it has to be a global setting. Currently, setting the value to true for a single dashboard will enable it for all, this could be confusing

And again, setting a global var on provider init is probably preferable to using os.Getenv at various points (this allows configuration from both a provider option or an env var)

Second thing for this to get merged, is that I'd like a test

fgouteroux:
I agree, but there is something not clear for me. How to give the provider value 'store_dashboard_sha256' into the state function of the variable 'config_json' in resource_dahboard.go needed for the diff ?

julienduchesne:
> I agree, but there is something not clear for me. How to give the provider value 'store_dashboard_sha256' into the state function of the variable 'config_json' in resource_dahboard.go needed for the diff ?

I'd set it as a global var on provider startup. We don't have many options here unfortunately. At least, it'll be clear that it's global

fgouteroux:
@julienduchesne I'm not sure that this is what you mean with global var on provider startup. Am I on the good way or not ?

julienduchesne:
> @julienduchesne I'm not sure that this is what you mean with global var on provider startup. Am I on the good way or not ?

Can you make that a regular var: `var storeDashboardSHA256 bool` in `provider.go` instead of an env var? That way you won't have to set/parse an envvar

fgouteroux:
> > @julienduchesne I'm not sure that this is what you mean with global var on provider startup. Am I on the good way or not ?
> 
> Can you make that a regular var: `var storeDashboardSHA256 bool` in `provider.go` instead of an env var? That way you won't have to set/parse an envvar

yes, it 's more cleaner...

julienduchesne:
Alright, this LGTM. Can you add an acceptance test for this?

fgouteroux:
Is that test is enough ?

grafana/resource_dashboard_test.go
```diff
@@ -275,3 +276,33 @@ func Test_normalizeDashboardConfigJSON(t *testing.T) {
 		})
 	}
 }
+
+func Test_storeDashboardSHA256(t *testing.T) {
+	type args struct {
+		config interface{}
+	}
+
+	storeDashboardSHA256 = true
```
julienduchesne: You'll need to revert that after the test with a defer

julienduchesne (commented):
I think this needs a full acceptance test that uses the actual terraform test framework. This is a major change of behavior of the dashboard resource and we need to check various things like that there isn't drift right after an apply. It should apply the resource and check that the saved state is a sha256, also test an update

Also, I suspect that there is no test that checks that currently, we are saving valid json. We should validate that also in one of the existing tests

fgouteroux:
@julienduchesne I added the acc test with create/update but  it works only with setting env var GRAFANA_STORE_DASHBOARD_SHA256.  I try to add this without success
```
// testAccPreCheckSHA256 verifies required provider testing configuration. It should
// be present in TestAccDashboard_isSHA256_stored acceptance test.
//
// These verifications and configuration are preferred at this level to prevent
// provider developers from experiencing less clear errors for every test.
func testAccPreCheckSHA256(t *testing.T) {
	for _, e := range testAccPreCheckEnv {
		if v := os.Getenv(e); v == "" {
			t.Fatal(e + " must be set for acceptance tests")
		}
	}
	testAccProviderConfigure.Do(func() {
		// Since we are outside the scope of the Terraform configuration we must
		// call Configure() to properly initialize the provider configuration.
		raw := map[string]interface{}{
			"store_dashboard_sha256": true,
		}
		err := testAccProvider.Configure(context.Background(), terraform.NewResourceConfigRaw(raw))
		if err != nil {
			t.Fatal(err)
		}
	})
}
```

grafana/resource_dashboard_test.go
```diff
@@ -15,58 +16,74 @@ func TestAccDashboard_basic(t *testing.T) {
...
-			{
-				// Importing matches the state of the previous step.
-				ResourceName:            "grafana_dashboard.test",
-				ImportState:             true,
-				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"message"},
-			},
-		},
-	})
+	for _, useSHA256 := range []bool{false, true} {
```
julienduchesne, line 19: Review without whitespace changes. I reused the basic test but when it's a sha256, the expected state is different

julienduchesne (approved):
I fixed the tests. This LGTM!

fgouteroux:
> I fixed the tests. This LGTM!

@julienduchesne  many thanks

inkel:
I'm reviewing, give me a couple minutes.

grafana/resource_dashboard.go
```diff
@@ -211,12 +213,17 @@ func ReadDashboard(ctx context.Context, d *schema.ResourceData, meta interface{}
 		return diag.FromErr(err)
 	}
 
+	configJSON := d.Get("config_json").(string)
+
+	// Check if the SHA256 of configJSON is stored in the state
+	isSHA256, _ := regexp.MatchString(`[A-Fa-f0-9]{64}`, configJSON)
```
inkel: I'm not keen on compiling the regular expression every time this is evaluated. Can you cache this at the package level and then call the `MatchString` function on that?

Also, while at it, can we change the regular expression to have start and end anchors, as in `^[A-Fa-f0-9]{64}$`?

I've made a quick benchmark and the performance, both in memory and speed, is too much to ignore:

```
$ gotest -run=XXX -benchmem -bench='^BenchmarkSHA$' -benchtime=1000000x
goos: darwin
goarch: amd64
pkg: github.com/inkel/misc-go-benchmarks
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkSHA/sha256/uncached-12                  1000000              4495 ns/op            6999 B/op         23 allocs/op
BenchmarkSHA/sha256/cached-12                    1000000               603.8 ns/op             0 B/op          0 allocs/op
BenchmarkSHA/sha256/cached+anchors-12            1000000               613.0 ns/op             0 B/op          0 allocs/op
BenchmarkSHA/json64/uncached-12                  1000000             13037 ns/op            6985 B/op         23 allocs/op
BenchmarkSHA/json64/cached-12                    1000000              9241 ns/op               0 B/op          0 allocs/op
BenchmarkSHA/json64/cached+anchors-12            1000000                40.24 ns/op            0 B/op          0 allocs/op
BenchmarkSHA/json/uncached-12                    1000000              7295 ns/op            7000 B/op         23 allocs/op
BenchmarkSHA/json/cached-12                      1000000              3172 ns/op               0 B/op          0 allocs/op
BenchmarkSHA/json/cached+anchors-12              1000000                39.25 ns/op            0 B/op          0 allocs/op
PASS
ok      github.com/inkel/misc-go-benchmarks     39.038s
```

Here, `json64` refers to a string with a length of 64 characters, same as a SHA256 hash. Note particularly the case of passing a string that is not a SHA256 hash while the regular expression is cached and includes anchors.

inkel (changes requested):
I've a minor request, once that is addressed I'll be happy to re-review and approve.

grafana/resource_dashboard.go
```diff
@@ -340,5 +344,16 @@ func normalizeDashboardConfigJSON(config interface{}) string {
...
+	if storeDashboardSHA256 {
+		configHash := sha256.Sum256(j)
+		return fmt.Sprintf("%x", configHash[:])
+	} else {
+		return string(j)
+	}
+}
+
+func MatchStringSHA256(str string) bool {
+	match, _ := regexp.MatchString(`^[A-Fa-f0-9]{64}$`, str)
```
julienduchesne: By caching at the package level, I think what @inkel meant was to compile the regex once like this:
https://github.com/grafana/terraform-provider-grafana/blob/master/grafana/provider.go#L24-L28

So you could add sha256Regexp to that list and use that pre-compiled regex when matching

inkel: Yes, that's correct. Sorry I wasn't clear enough! 🙏

fgouteroux: yes, I actually didn't understand..ok fixed

grafana/resource_dashboard_test.go
inkel: @fgouteroux have you had a chance to look at this comment?

julienduchesne: It wasn't a comment for him. Just a general "if you review this test, review it without whitespace". I wrote that test 😄

inkel: Aaaahhh ok, my bad, then.

So, just to see if I got this straight: the test fails?

julienduchesne: No, the test passes. It just shows lots of changes because indentation changes

inkel (approved):
Great work, @fgouteroux! Thank you for your contribution.

fgouteroux:
> Great work, @fgouteroux! Thank you for your contribution.

@inkel @julienduchesne Thanks for your tips/feedback. I'm happy to see it merged ;-)

PR #225, opened 2021-06-22
fix!: make alert_notification settings nonsensitive

### LINKED ISSUE #220: Make settings in Alert Notification Resource non-sensitive
As sensitive settings can now (https://github.com/grafana/terraform-provider-grafana/commit/a8c02746a75f937772c0120e86eee581ed2ecbbc) be put in `secure_settings`, I presume `settings` no longer has to be marked _sensitive_. Doing so would make the terraform plan much easier to read, as any changes within the settings would be displayed.

thor:
Fixes #220.

As secure_settings should be used for sensitive settings, this marks the
alert_notification resource as insensitive by removing the sensitive
flag.

Suggest bumping the major version in order to avoid accidental "reveals"
of settings data that is sensitive.


trotttrotttrott (approved):
Thanks!

Issue #227, opened 2021-06-24
Leverage "subcategory" For Organizing Docs


trotttrotttrott:
We'd like to leverage [subcategories](https://www.terraform.io/docs/registry/providers/docs.html#subcategories) to improve documentation navigation on the [Terraform registry](https://registry.terraform.io/providers/grafana/grafana/latest/docs). I'd like to propose `Grafana`, `Enterprise`, and `Cloud` as our initial subcategories. This would produce the following navigation hierarchy:

* Grafana
  * Resources
    * grafana_alert_notification
    * grafana_dashboard
    * grafana_dashboard_permission
    * grafana_data_source
    * ...
* Enterprise
  * Resources
    * grafana_builtin_role_assignment
    * grafana_role
* Cloud
  * Resources
    * grafana_synthetic_monitoring_check
    * grafana_synthetic_monitoring_probe
  * Data Sources
    * grafana_synthetic_monitoring_probe
    * grafana_synthetic_monitoring_probes

Issue #226, opened 2021-06-24
CI for Cloud Acceptance Tests


trotttrotttrott:
We started using the `cloud` build tags for acceptance test files that need to run against an instance of Grafana Cloud. These tests should be run in CI

PR #228, opened 2021-06-24
Synthetic Monitoring Probe and Check Resources and Data Sources

### LINKED ISSUE #198: Add Synthetic monitoring resource
Hi. Since we're adding more and more checks in Synthetic monitoring, it would be very helpful to have those resources in Terraform; as well as the Grafana Alerts.

Thanks

trotttrotttrott:
Adds resources:

* grafana_synthetic_monitoring_probe
* grafana_synthetic_monitoring_check

Add data sources:

* grafana_synthetic_monitoring_probe (single probe by name)
* grafana_synthetic_monitoring_probes (map of all probes)

### Testing

A "cloud" build tag was added to the top of the Synthetic Monitoring `_test.go` files (`// +build cloud`). Therefore, CI in its current state will not run them. I've created an issue to capture the effort of enabling CI for these tests as a followup: https://github.com/grafana/terraform-provider-grafana/issues/226.

### Documentation

Documentation was generated with `go generate`. The resources have custom templates to allow the inclusion of multiple examples. I've created another issue to followup to use subcategories to group cloud resources together in navigation: https://github.com/grafana/terraform-provider-grafana/issues/227.

### adapters.go

The addition of this file was inspired by https://github.com/paultyng/terraform-provider-unifi/blob/8601203944690676671871390d938ed5238a7f08/internal/provider/strings.go. It's a collection of functions for converting types. These should be useful for making resources and data sources more succinct.

### schema.go

This was inspired by https://github.com/hashicorp/terraform-provider-google/blob/eb82ee95846c03976d7df3fbbb965f0ea8106d2b/google/datasource_helpers.go. These functions allow us to reuse resource schemas for data sources. This is instead of duplicating them. It's intended for general schema utilities like this that we can use throughout the provider.

Closes https://github.com/grafana/terraform-provider-grafana/issues/198


inkel (approved):
LGTM! One thing though: `examples/resources/grafana_synthetic_monitoring_probe/basic.tf` and `examples/resources/grafana_synthetic_monitoring_probe/complex.tf` look very similar, and neither actually looks complex. Is this because we require to have both?

captncraig:
Would it be possible to have the probe resource return the api token for the probe on creation? That would let us store it in vault or anywhere else to automatically provision the probe end-to-end. I realize that may be hard to do on subsequent reads, but maybe it doesn't need to?

trotttrotttrott:
Merging blocked due to tests failing. https://github.com/grafana/terraform-provider-grafana/pull/231 should resolve this.

inkel (approved):
Thanks for renaming those tests. Changes LGTM, and #231 is approved, so it's only missing a rebase and it should be good to merge.

PR #231, opened 2021-06-25
Fix Acceptance Tests: terraform-json 0.12.0

trotttrotttrott:
All acceptance tests started failing today with `unsupported state format version: expected "0.1", got "0.2"`. Updating `terraform-json` fixes this - though I don't fully understand why.

It has something to do with the [Terraform 1.0.1 release today](https://github.com/hashicorp/terraform/blob/v1.0/CHANGELOG.md#101-june-24-2021) and presumably this particular pull request: https://github.com/hashicorp/terraform/pull/28889.

https://github.com/hashicorp/terraform/pull/28889/files#diff-aec0f9962c6764cbde3325886b9b81650e1007d96d7693a9dcf0df9b53f09d2eL25-R25

Another strange finding was that tests would pass without updating `terraform-json` if Terraform <= 1.0.0 was in the path of the acceptance test runtime. If `terraform` was not in the path or if Terraform 1.0.1 was in the path, these failures would occur.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/grafana/grafana-api-golang-client v0.0.0-20210622232726-d2fbe458a0cf
 	github.com/grafana/synthetic-monitoring-api-go-client v0.0.0-20210608014215-abcc28348a76
 	github.com/hashicorp/go-cleanhttp v0.5.2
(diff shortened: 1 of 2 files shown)


inkel (approved):
Thanks for the update.

Issue #146, opened 2020-12-07
Error creating teams with token


ichasco:
Hi there,

I am getting problems when creating resources with tokens. Terraform gives me a error but it works (I don't know why)

### Terraform Version
```
Terraform v0.13.4
+ provider registry.terraform.io/grafana/grafana v1.7.0
+ provider registry.terraform.io/hashicorp/vault v2.16.0
```

### Affected Resource(s)

- grafana_team

### Terraform Configuration Files
Provider
```hcl

## ADMIN

data "vault_generic_secret" "grafana_admin" {
  path = "devops/Grafana/Users/Admin"
}

provider "grafana" {
  alias = "admin"
  url  = "https://grafana.example.com"
  auth = "admin:${data.vault_generic_secret.grafana_admin.data["Password"]}"

}

## MAIN ORG

data "vault_generic_secret" "grafana_token" {
  path = "devops/Grafana/Tokens"
}

provider "grafana" {
  url  = "https://grafana.example.com"
  auth = data.vault_generic_secret.grafana_token.data["Token-Terraform"]
}

```

Create ORG

```hcl
resource "grafana_organization" "mainorg" {
  provider    = grafana.admin
  name        = "MainOrg"
  admin_user  = "admin"
}
```
```
terraform apply -target=grafana_organization.mainorg
```

Create the token manually in created org with Admin permission

Create Team
```hcl
resource "grafana_team" "devops" {
  name    = "devops"
}

resource "grafana_team_preferences" "team_preferences" {
  team_id           = grafana_team.devops.id
  theme             = "dark"
  timezone          = "browser"
}
```

```
terraform apply
```

### Debug Output

Create ORG output
```
data.vault_approle_auth_backend_role_id.role: Refreshing state... [id=auth/approle/role/grafana-devops/role-id]
vault_approle_auth_backend_role_secret_id.id: Refreshing state... [id=backend=approle::role=grafana-devops::accessor=XXXXXXXXXXX]
vault_approle_auth_backend_login.login: Refreshing state... [id=XXXXXXXXXXXXXXXX]
data.vault_generic_secret.grafana_admin: Refreshing state... [id=devops/Grafana/Users/Admin]
data.vault_generic_secret.grafana_token: Refreshing state...
grafana_organization.mainorg: Refreshing state... [id=2]
```

Create Team output

```
grafana_team.devops: Creating...
Error: status: 403, body: {"message":"Permission denied"}

[...]

PR #244, opened 2021-07-19
Team Resource Refactoring

trotttrotttrott:
Depends on https://github.com/grafana/grafana-api-golang-client/pull/31.

* Use OrgUsersCurrent (GET /api/org/users) when reconciling team members.
  GET /api/users is not accessible on Grafana Cloud.
* Sets team_id on create. It's been left as null.
* Read only sets email if it's not a blank string. null vs. "" makes
  it appear that team resources are changed outside of Terraform.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/grafana/terraform-provider-grafana
 go 1.16
 
 require (
-	github.com/grafana/grafana-api-golang-client v0.0.0-20210719231312-127ddfdfd452
+	github.com/grafana/grafana-api-golang-client v0.0.0-20210720012848-3049c6914b86
 	github.com/grafana/synthetic-monitoring-agent v0.0.24
 	github.com/grafana/synthetic-monitoring-api-go-client v0.0.2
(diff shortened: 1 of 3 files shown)


cinaglia (approved):
Changes look reasonable to me.

cinaglia:
From a glance, it looks like this will likely solve https://github.com/grafana/terraform-provider-grafana/issues/146 as well.

cinaglia:
One thing that just came to mind. Will things continue to work as expected for users that rely on basic auth? I _think_ it will as long as they're specifying `org_id`.

trotttrotttrott:
> One thing that just came to mind. Will things continue to work as expected for users that rely on basic auth? I think it will as long as they're specifying org_id.

@cinaglia the provider requires `org_id` so there will always be a "current org" - should be all good 👍

https://github.com/grafana/terraform-provider-grafana/blob/23363ad5616e16dbc9bbd028cafa7423579cfcd4/grafana/provider.go#L49-L54

Also `TestAccTeam_Members` would fail if basic auth stopped working since we use `admin:admin` for our acceptance tests in CI.

Issue #246, opened 2021-07-20
Plan crashes when undetermined resource used in templated JSON


stevenproctor:
### Terraform Version
* Terraform: 1.0.2
* Terraform Grafana Provider: 1.12.0
* Grafana: v6.6.1 (21bf8b71bc)

### Affected Resource(s)
* grafana_dashboard

### Simple HCL
```hcl
terraform {
  required_version = "~>1.0.0"
  experiments      = [module_variable_optional_attrs]

  required_providers {
    grafana = {
      source  = "grafana/grafana"
      version = "1.12.0"
    }
  }
}

provider "grafana" {
}


resource "grafana_dashboard" "dashboard" {
  config_json = templatefile(
    "${path.module}/dashboard.json.tmpl",
    {
      environment     = "test-env"
      dashboard_title = "proctor-test-dashboard-error"
      uid             = "${random_uuid.dashboard_uid.id}-tf"
    }
  )
}

resource "random_uuid" "dashboard_uid" {
  keepers = {
    environment     = "test-env"
    dashboard_title = "proctor-test-dashboard-error"
  }
}
```

Simple JSON Template:
```json
{
  "annotations": {
    "list": [
      {
        "builtIn": 1,
        "datasource": "-- Grafana --",
        "enable": true,
        "hide": true,
        "iconColor": "rgba(0, 211, 255, 1)",
        "name": "Annotations & Alerts",
        "type": "dashboard"
      }
    ]
  },
  "editable": true,
  "graphTooltip": 0,
  "links": [],
  "refresh": "30s",
  "schemaVersion": 22,
  "style": "dark",
  "tags": [],
  "templating": {
  },
  "time": {
    "from": "now-30d",
    "to": "now"
  },
  "timepicker": {
    "refresh_intervals": [
      "5s",
      "10s",
      "30s",
      "1m",
      "5m",
      "15m",
      "30m",
      "1h",
      "2h",
      "1d"
    ]
  },
  "timezone": "",
  "title": "${ dashboard_title } (${ environment })",
  "uid": "${ uid }"
}
```

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.
```
╷
│ Error: Plugin did not respond
│ 
│   with module.grafana_dashboard.grafana_dashboard.dashboard,
│   on .terraform/modules/grafana_dashboard/modules/grafana-dashboard/main.tf line 25, in resource "grafana_dashboard" "dashboard":
[...]

PR #248, opened 2021-07-26
Crash when config_json will be known after apply #246

jnahelou:
Following discussion in #246

I have the same issue in acc tests with a computed value in config_json attribute like :
```
# main.tf
  config_json = <<EOD
{
  "title": "Terraform Acceptance Test",
  "id": ${substr(uuid(), 0, 6)},
  "uid": "basic",
  "version": 34
}
EOD
}
```

Despite it's a new plan (without state file), Terraform call the StateFunc `normalizeDashboardConfigJSON` with the default value 
 `UnknownVariableValue` (74D93920-ED26-11E3-AC10-0800200C9A66) which cause the unmashal error.
https://github.com/hashicorp/terraform-plugin-sdk/blob/112e2164c381d80e8ada3170dac9a8a5db01079a/internal/configs/hcl2shim/values.go#L16

I suggest to return an error during unmashall instead of terminate the program using `panic()`, and then skip the normalizeDashboardConfigJSON if the `config_json` doesn't have the expected format.

To reproduce in tests, I created 2 dashboards (one with the minimal configuration and a second with a configuration attribute defined by the first one)

### DIFF
--- a/examples/resources/grafana_dashboard/_acc_computed.tf
+++ b/examples/resources/grafana_dashboard/_acc_computed.tf
@@ -1,0 +1,19 @@
+# Creating a dashboard with a random uid.
+# We'd like to ensure that using a computed configuration works.
+
+resource "grafana_dashboard" "test" {
+  config_json = <<EOD
+{
+  "title": "Terraform Acceptance Test"
+}
+EOD
+}
+
+resource "grafana_dashboard" "test-computed" {
+  config_json = <<EOD
+{
+  "title": "Terraform Acceptance Test Computed",
+	"tags": ["${grafana_dashboard.test.uid}"]
+}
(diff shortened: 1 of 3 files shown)


flokli:
@jnahelou I tried to give this patch a try, and I can confirm the

```
Error: rpc error: code = Canceled desc = context canceled



Error: rpc error: code = Unavailable desc = transport is closing



Error: rpc error: code = Unavailable desc = transport is closing

```

are gone :tada:!

I get a proper plan for the planned changes.

jnahelou:
I pushed my unit test for normalizeDashboardConfigJSON

trotttrotttrott (approved):
Thanks @jnahelou!

PR #255, opened 2021-08-01
Use golangci-lint in Drone Pipelines

trotttrotttrott:
Use [golangci-lint](https://golangci-lint.run/). This is similar to what we did in the [Grafana API client](https://github.com/grafana/grafana-api-golang-client/pull/4). It's [consistent with Grafana](https://github.com/grafana/grafana/blob/31903778aecf9b2c5362f9483ee6c2bae1eaa742/scripts/go/configs/.golangci.toml).


julienduchesne (approved):

grafana/resource_builtin_role_assignment.go
```diff
@@ -225,7 +224,7 @@ func createOrRemove(meta interface{}, name string, changes []RoleChange) error {
 			err = client.DeleteBuiltInRoleAssignment(br)
 		}
 		if err != nil {
-			return errors.New(fmt.Sprintf("Error with %s %v", name, err))
+			return fmt.Errorf("Error with %s %v", name, err)
```
inkel: Nit: when wrapping an `error` use `%w` instead of `%v`, so it works with `errors.Unwrap`.

grafana/resource_builtin_role_assignment.go
```diff
@@ -226,7 +224,7 @@ func createOrRemove(meta interface{}, name string, changes []RoleChange) error {
 			err = client.DeleteBuiltInRoleAssignment(br)
 		}
 		if err != nil {
-			return errors.New(fmt.Sprintf("Error with %s %v", name, err))
+			return fmt.Errorf("error with %s %v", name, err)
```
inkel: Nit: when wrapping an `error` use `%w` instead of `%v` so it works with `errors.Unwrap`
```suggestion
			return fmt.Errorf("error with %s %w", name, err)
```

inkel (approved):

Issue #243, opened 2021-07-15
Grafana Data Source for Elasticsearch type with es_version seems to have incompatible field type


parker-mar:
### Terraform Version
* Terraform: 1.0.2
* Terraform Grafana Provider: 1.13.1

### Grafana Version
8.0.5

### Affected Resource(s)
* grafana_data_source

### Terraform Configuration Files
```hcl
resource "grafana_data_source" "this" {
  name          = "my-elasticsearch"
  type          = "elasticsearch"
  database_name = "my-database"
  json_data {
    es_version = 70
  }
  ...
}
```

### Debug Output
https://gist.github.com/parker-mar/cdb84c799c87e1b1aee23253fc47c2c2

### Expected Behavior
Resource with es_version created/updated.

### Actual Behavior
Resource with es_version cannot be created/updated (because es_version seems to unmarshal input as string, and passes it along to an int64?).

### Steps to Reproduce
`terraform apply`

### Important Factoids
Under Grafana v8 https://grafana.com/docs/grafana/latest/whatsnew/whats-new-in-v8-0/#data-source-updates, it seems semver strings are now allowed. I wonder if the Grafana provider is incompatible with this.

> Use semver strings to identify Elasticsearch version
> 
> We changed how the configured Elasticsearch version is handled. You can now specify via provisioning the full semver string version of your instance (such as “7.12.1”) instead of the old version format based on numbers. There’s no manual intervention needed, the old options will be correctly recognized.

For more information, see https://grafana.com/docs/grafana/latest/datasources/elasticsearch/#elasticsearch-version.

vtorosyan:
Thanks for the issue. Indeed with Grafana 8.0+ the change has been introduced which uses semantic versions as string for Elasticsearch. While Grafana backend is compatible with numeric values, this is considered legacy and should not be used anymore, the recommendation is to use full semantic version in the terraform configuration (e.g. `7.0.0` instead of `70`).

[This PR](https://github.com/grafana/terraform-provider-grafana/pull/267) should fix the problem.

julienduchesne:
Should be fixed in v1.14.0+

PR #267, opened 2021-09-10
Update Elasticsearch version to be string in grafana_data_source resource

vtorosyan:
### Context

Grafana 8.0+ switched to use semantic versions instead of numeric values. The numeric usage is considered legacy and should not be used anymore and semantic versions are represented as string values. This PR updates the `es_version` attribute in the `json_data` for `grafana_data_source` resource.

Note that this change is only relevant for Grafana 8.0+ versions. Prior versions to 8.0 must still use the provider version which supports `es_version` as a numeric value.

### Impact

All `grafana_data_source` resources which are provisioning Elasticsearch datasource need to changed to use full semantic version of Elasticsearch. 

For example, instead of the following:

```
resource "grafana_data_source" "elasticsearch" {
  name = "elasticsearch"
  type = "elasticsearch"
  json_data {
    es_version = 70
  }
}
```
the following should be used

```
resource "grafana_data_source" "elasticsearch" {
  name = "elasticsearch"
  type = "elasticsearch"
  json_data {
    es_version = "7.0.0"
  }
}
```

### Fixes

https://github.com/grafana/terraform-provider-grafana/issues/243

### Note for reviewers

- Relevant [change in the grafana-api-go-client](https://github.com/grafana/grafana-api-golang-client/pull/34) is a prerequisite for this PR.
- There is an internal [escalation](https://github.com/grafana/support-escalations/issues/841) for this problem, so getting this fixed has some level of urgency.

### DIFF
--- a/docs/resources/data_source.md
+++ b/docs/resources/data_source.md
@@ -110,7 +110,7 @@ Optional:
 - **default_project** (String) (Stackdriver) The default project for the data source.
 - **default_region** (String) (CloudWatch) The default region for the data source.
 - **encrypt** (String) (MSSQL) Connection SSL encryption handling: 'disable', 'false' or 'true'.
-- **es_version** (Number) (Elasticsearch) Elasticsearch version as a number (2/5/56/60/70).
+- **es_version** (String) (Elasticsearch) Elasticsearch semantic version (Grafana v8.0+).
 - **graphite_version** (String) (Graphite) Graphite version.
 - **http_method** (String) (Prometheus) HTTP method to use for making requests.
 - **interval** (String) (Elasticsearch) Index date time format. nil(No Pattern), 'Hourly', 'Daily', 'Weekly', 'Monthly' or 'Yearly'.
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/grafana/synthetic-monitoring-agent v0.0.24
 	github.com/grafana/synthetic-monitoring-api-go-client v0.0.2
 	github.com/hashicorp/go-cleanhttp v0.5.2
+	github.com/hashicorp/go-cty v1.4.1-0.20200414143053-d3edf31b6320
 	github.com/hashicorp/terraform-plugin-docs v0.4.0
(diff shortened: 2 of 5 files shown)


dsotirakis (commented):
LGTM - will also need to update `grafana-api-golang-client` first, your other PR that is, and then merge this, since the `golangci` check fails! Not clicking approve yet just to have than in mind!

inkel:
This is a huge breaking change. While I agree with the rationale behind it, I think we should instead mark this attribute as deprecated and add a new one; we could even take the opportunity to give it a better name, like `elasticsearch_version` (I never liked the `es` short form). Let's wait on @trotttrotttrott to chime in.

vtorosyan:
@inkel thanks for the feedback. I am generally supporting the idea of deprecating the attribute, the only challenge I see is that the current terraform plans using the deprecated attribute will still be broken and they would have to remove the attribute entirely for the plans to work, so the way I understand is that it's ultimately going to have the same breaking effect. 

Happy to move in any direction, and I'd also still not disregard the option of making this work from Grafana side (although the cost and risk is much higher of doing that). @trotttrotttrott would appreciate your thoughts on this.

trotttrotttrott:
I think we should proceed and accept this as a breaking change for on Grafana 7.x and earlier. We don't have a great story for backwards compatibility due to the difficultly of keeping up with all Grafana API changes. Supporting the latest major version is our priority.

The `es_version` name is consistent with the actual field in [jsonData](https://grafana.com/docs/grafana/latest/administration/provisioning/#json-data), `esVersion`. so we're going to want to keep it that way.

I've submitted a PR on this one to add a warning message if `es_version` is not a semantic version. I think this could alleviate some confusion - https://github.com/grafana/terraform-provider-grafana/pull/276.

vtorosyan:
Thanks @trotttrotttrott I've merged your PR, if there are any further feedback on that particular change, hope others can comment here.

So changes in this PR are ready to be double checked 👍

trotttrotttrott (approved):

PR #276, opened 2021-10-05
Warn If `es_version` is not a semantic version

trotttrotttrott:
https://github.com/grafana/terraform-provider-grafana/pull/267 introduces a subtle breaking change.

When merged the `es_version` field will no longer be valid for Grafana versions 7.x and earlier. This warning will at least provide an indication that it's intended for use with Grafana 8+ instead of silently failing.

For Grafana 8+ users, it will provide an indication that they must update this field for it to actually work.

Example execution plan output:

```
╷
│ Warning: Expected semantic version
│
│   with grafana_data_source.elasticsearch,
│   on datasource.tf line 7, in resource "grafana_data_source" "elasticsearch":
│    7:     es_version                    = 70
│
│ As of Grafana 8.0, the Elasticsearch plugin expects es_version to be set as a semantic version (E.g. 7.0.0, 7.6.1).
│
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/grafana/synthetic-monitoring-agent v0.0.24
 	github.com/grafana/synthetic-monitoring-api-go-client v0.0.2
 	github.com/hashicorp/go-cleanhttp v0.5.2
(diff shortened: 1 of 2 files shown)


vtorosyan (approved):

vtorosyan:
Thanks @trotttrotttrott , totally makes sense.

PR #286, opened 2021-11-11
fix(resource_alert_notification): safeguard against a nil value for AlertNotification.SecureFields

### LINKED ISSUE #280: Go panic: interface conversion with secure_settings attribute during grafana_alert_notification ressource creation
### Terraform Version
* Terraform: 0.15.3
* Terraform Grafana Provider: 1.14.0
* Grafana: 7.4.5

### Affected Resource(s)
* grafana_alert_notification

```
terraform {
  required_version = "= 0.15.3"
  required_providers {
    grafana = {
      source = "grafana/grafana"
      version = "=1.14.0"
    }
  }

provider "grafana" {
  url  = var.grafana_url
  auth = "${var.grafana_auth_username}:${var.grafana_auth_password}"
  org_id = var.grafana_org_id
}

resource "grafana_alert_notification" "monitoring" {
  name          = "Monitoring notification channel"
  type          = "email"
  is_default    = false
  send_reminder = true
  frequency     = "24h"
  settings = {
    addresses   = var.grafana_addresses_gitlab_monitoring
    uploadImage = "true"
  }
}

```

### Panic Output
https://gist.github.com/mehdiElma/b4404e4ddb246d0633a33ea851a4bf45

### Expected Behavior
Create grafana_alert_notification with terraform and without any error.

### Actual Behavior
Create grafana_alert_notification with terraform failed because of a Go panic interface conversion error.

### Steps to Reproduce
1. `terraform init`
2. `terraform apply`

I think the issue is due to this line : https://github.com/grafana/terraform-provider-grafana/blob/8eb0a8453d3b6a0c4a8ea579905c77895867d06e/grafana/resource_alert_notification.go#L163

```
for k, v := range alertNotification.SecureFields.(map[string]interface{}) {
		boolVal, ok := v.(bool)
		switch {
		case ok && boolVal:
			secureSettings[k] = "true"
		case ok && !boolVal:
			secureSettings[k] = "false"
		default:
			secureSettings[k] = v
		}
}
```
It tryed to make a range on an attribute, alertNotification.SecureFields, that doesn't exist (nil value), to get data from secure_settings. The makeAlertNotification function return this :  
```
return &gapi.AlertNotification{
		ID:                    id,
		Name:                  d.Get("name").(string),
		Type:                  d.Get("type").(string),
		IsDefault:             d.Get("is_default").(bool),
		DisableResolveMessage: d.Get("disable_resolve_message").(bool),
		UID:                   d.Get("uid").(string),
		SendReminder:          sendReminder,
		Frequency:             frequency,
		Settings:              settings,
		SecureSettings:        secureSettings,
	}, err
```
There is no **SecureFields** attribute on AlertNotification, I think it should be **SecureSettings** instead.

I'm a newby in Go, can you help me please ?

ctrombley:
Resolves #280 

I haven't been able to reproduce this with an acceptance test, but my local instance of grafana is not returning a `secureFields` field when reading an alert notification resource.  This is crashing the provider during a plan phase:
https://gist.github.com/ctrombley/4cdfa01b110d85bc28c88890815351b0

This PR safeguards against this scenario with a type assertion test.  

- Grafana `7.1.0-beta-2`
- Provider `1.14.0`
- Terraform `0.15.5`

I also added some instructions to the README for filtering tests during a test run when using docker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,20 @@ GRAFANA_ORG_ID=1 \
 make testacc
 ```
 
+#### Limiting an acceptance test run to specific tests
+
+The tests run by a target can be filtered by specifying an additional `run` argument:
+
+```sh
+TESTARGS='--run=TestAccDataSource_basic' make testacc-docker
+```
+
+The value passed to `run` supports regular expressions:
+
+```sh
+TESTARGS='--run=TestAccDataSource_.*' make testacc-docker
+```
+
 #### Running enterprise tests
 
 To run tests for resources which are available only for Grafana Enterprise, running instance of Grafana Enterprise is required.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,6 +12,7 @@ services:
       - GRAFANA_URL=http://grafana:3000
       - GRAFANA_AUTH=admin:admin
       - GRAFANA_ORG_ID=1
+      - TESTARGS=${TESTARGS}
     working_dir: /go/src/github.com/grafana/terraform-provider-grafana
     volumes:
       - .:/go/src/github.com/grafana/terraform-provider-grafana
--- a/grafana/resource_alert_notification.go
+++ b/grafana/resource_alert_notification.go
@@ -160,15 +160,17 @@ func ReadAlertNotification(ctx context.Context, d *schema.ResourceData, meta int
 	}
 	secureSettings := map[string]interface{}{}
 
-	for k, v := range alertNotification.SecureFields.(map[string]interface{}) {
-		boolVal, ok := v.(bool)
-		switch {
-		case ok && boolVal:
-			secureSettings[k] = "true"
-		case ok && !boolVal:
-			secureSettings[k] = "false"
-		default:
-			secureSettings[k] = v
+	if f, ok := alertNotification.SecureFields.(map[string]interface{}); ok {
+		for k, v := range f {
+			boolVal, ok := v.(bool)
+			switch {
+			case ok && boolVal:
+				secureSettings[k] = "true"
+			case ok && !boolVal:
+				secureSettings[k] = "false"
+			default:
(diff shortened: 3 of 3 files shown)


inkel (commented):
🤔 interesting. I wonder why it isn't responding with that field, as the [API documentation](https://grafana.com/docs/grafana/latest/http_api/alerting_notification_channels/#get-notification-channel-by-id) mentions it. At first I thought it might have been because you are using an older Grafana version, but [docs for 7.1](https://grafana.com/docs/grafana/v7.1/http_api/alerting_notification_channels/#get-notification-channel-by-id) also show the field.

Changes LGTM, although if the field is supposed to be returned then I think it's ok to panic due to trying accessing it. Let me see if I can reproduce before merging the changes.

RhynoVDS:
Hi all,
Any update on this? We are having the same issue when creating alert channels.
Thanks @ctrombley for fixing it.

RhynoVDS:
We upgraded our Grafana version from: 6.X to 8.3.1 which has resolved this issue for us.
Hope this helps others.
Thank you all for your hard work.

inkel:
Hi! I've been trying to reproduce this one without success. Note that the reported version of Grafana, [`7.1.0-beta-2`](https://github.com/grafana/grafana/releases/tag/v7.1.0-beta2) is from July 2020, and a beta while at it, so it's expected some issues might exist.

Given that updating Grafana to a later, non-beta, more supported version ([8.0.0](https://github.com/grafana/grafana/releases/tag/v8.0.0) was released in June 2021), I'm going to close this PR.

Note that also the Terraform version in used is pretty old, in fact Terraform is already at 1.x, so I suggest upgrading that one as well.

Thank you @ctrombley for the work you did on this, please let me know if you still have any issues after upgrading.

PR #289, opened 2021-11-17
Add `grafana_data_source_permission` resource

### LINKED ISSUE #205: New feature: datasource permissions
Suggesting a new feature to support managing datasource permissions resources via Terraform.

The existing public API suffices: https://grafana.com/docs/grafana/latest/http_api/datasource_permissions/

Proposed example resource block:

```
resource "grafana_datasource_permission" "fooPermissions" {
  datasource_id = grafana_datasource.foo.id
  permissions {
    team_id    = grafana_team.team.id
    permission = "Query" (optional, default "Query", options: Query)
  }
  permissions {
    user_id    = grafana_user.user.id
  }
}
```

julienduchesne:
Requires https://github.com/grafana/grafana-api-golang-client/pull/37. Once that's merged, I'll tag the golang client, I think. Using a commit is very confusing

As requested here: https://github.com/grafana/terraform-provider-grafana/issues/205


grafana/resource_datasource_permission.go
```diff
@@ -0,0 +1,173 @@
...
+						},
+						"user_id": {
+							Type:        schema.TypeInt,
+							Optional:    true,
+							Default:     0,
+							Description: "ID of the user to manage permissions for.",
+						},
+						"permission": {
+							Type:         schema.TypeString,
+							Optional:     true,
```
inkel: Even when we have a default value, which in addition is the only possible value, I'd rather have this as a required attribute. Personally I like it explicit better than implicit. WDYT?

grafana/resource_datasource_permission.go
```diff
@@ -0,0 +1,229 @@
...
+
+	return nil
+}
+
+func mapDatasourcePermissionStringToInt64(permission string) int64 {
+	if permission == "Query" {
+		return 1
+	}
+	return -1
+}
```
inkel: What happens if we map the string to `-1`? Will the API request fail? If so, maybe we should catch this earlier and let the user know the value couldn't be mapped 🤔

inkel: Also: as in the case of the PR in the API client, if we defined this as a custom enum, then perhaps we could avoid these mappings and conversions if we define them there.

grafana/resource_datasource_permission.go
```diff
@@ -0,0 +1,229 @@
...
+	count := 0
+	for _, permission := range response.Permissions {
+		permissionItem := make(map[string]interface{})
+		permissionItem["team_id"] = permission.TeamID
+		permissionItem["user_id"] = permission.UserID
+		permissionItem["permission"] = mapDatasourcePermissionInt64ToString(permission.Permission)
+
+		permissionItems[count] = permissionItem
+		count++
+	}
```
inkel: There's no need for the `count` variable, you can make use of the index returned by `range`:
```suggestion
	permissionItems := make([]interface{}, len(response.Permissions))
	for i, permission := range response.Permissions {
		permissionItem := make(map[string]interface{})
		permissionItem["team_id"] = permission.TeamID
		permissionItem["user_id"] = permission.UserID
		permissionItem["permission"] = mapDatasourcePermissionInt64ToString(permission.Permission)

		permissionItems[i] = permissionItem
	}
```

inkel (approved):
Just left a couple questions/suggestions, but it's good for merging!

grafana/resource_datasource_permission.go
julienduchesne: ah, oops. This is what I get for copy pasting from other resources

grafana/resource_datasource_permission.go
julienduchesne: Yeah, I made this an error. Definitely more robust. I should stop relying on the rest of the code and think a bit...

inkel: Nah, you're doing great! <3

inkel (approved):
:shipit:

PR #290, opened 2021-11-17
Add support for importing probe auth_token

mem:
Currently we can import an existing probe by doing:

	terraform import address probe_id

This stores the probe in the state, but the corresponding auth_token is
null.

In order to import the existing auth_token, add support for this form:

	terraform import address probe_id[:auth_token]

The auth_token is optional in order to avoid breaking existing uses.

Signed-off-by: Marcelo E. Magallon <marcelo.magallon@grafana.com>

### DIFF
--- a/grafana/resource_synthetic_monitoring_probe_helper_test.go
+++ b/grafana/resource_synthetic_monitoring_probe_helper_test.go
@@ -1,0 +1,72 @@
+package grafana
+
+import (
+	"context"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func TestImportProbeStateWithToken(t *testing.T) {
+	testcases := map[string]struct {
+		input             string
+		expectError       bool
+		expectedID        string
+		expectedAuthToken string
+	}{
+		"valid id, no auth_token": {
+			input:             "1",
+			expectError:       false,
+			expectedID:        "1",
+			expectedAuthToken: "",
+		},
+		"valid id, valid auth_token": {
+			input:             "1:aGVsbG8=",
+			expectError:       false,
+			expectedID:        "1",
+			expectedAuthToken: "aGVsbG8=",
+		},
+		"valid id, invalid auth_token": {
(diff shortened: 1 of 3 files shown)


grafana/resource_synthetic_monitoring_probe_helper_test.go
```diff
@@ -0,0 +1,72 @@
...
+				if tc.expectedID != res[0].Id() {
+					t.Fatalf("expecting id %q, got %q", tc.expectedID, res[0].Id())
+				}
+
+				if tc.expectedAuthToken != "" {
+					output, ok := res[0].GetOk("auth_token")
+					if !ok {
+						t.Fatalf("expecting auth_token to be set")
+					} else if str, ok := output.(string); !ok || str != tc.expectedAuthToken {
+						t.Fatalf("expecting auth_token %q, got %q", tc.expectedAuthToken, str)
```
inkel: How about:
```suggestion
						t.Fatalf("expecting auth_token to be string %q, got %#v", tc.expectedAuthToken, output)
```

My reasoning is that the condition could fail because `output` couldn't be casted as a `string` as well.

grafana/resource_synthetic_monitoring_probe_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build cloud
```
inkel: Do we really need this change for this PR to work?

inkel (commented):
Overall looks good to me, I just have a few questions.

grafana/resource_synthetic_monitoring_probe_test.go
mem: no, thanks for noticing. That's probably a side-effect of me opening the file looking for a place to add the tests.

grafana/resource_synthetic_monitoring_probe_helper_test.go
mem: Make sense, thanks!

inkel (approved):

PR #291, opened 2021-11-19
Add new datasource: `grafana_user`

julienduchesne:
This will be used to add cloud tests. Since we can't create new users in the cloud (They are managed with grafana.com)


docs/index.md
```diff
@@ -32,6 +32,7 @@ provider "grafana" {
 - **ca_cert** (String) Certificate CA bundle to use to verify the Grafana server's certificate. May alternatively be set via the `GRAFANA_CA_CERT` environment variable.
 - **insecure_skip_verify** (Boolean) Skip TLS certificate verification. May alternatively be set via the `GRAFANA_INSECURE_SKIP_VERIFY` environment variable.
 - **org_id** (Number) The organization id to operate on within grafana. May alternatively be set via the `GRAFANA_ORG_ID` environment variable.
+- **retries** (Number) The amount of retries to use for Grafana API calls. May alternatively be set via the `GRAFANA_RETRIES` environment variable.
```
inkel, line 35: Is this related to the scope of the PR? Or should it have been a separated PR?

docs/data-sources/user.md
```diff
@@ -0,0 +1,58 @@
...
+  does not currently work with API Tokens. You must use basic auth.
+---
+
+# grafana_user (Data Source)
+
+* [Official documentation](https://grafana.com/docs/grafana/latest/manage-users/server-admin/server-admin-manage-users/)
+* [HTTP API](https://grafana.com/docs/grafana/latest/http_api/user/)
+
+This resource uses Grafana's admin APIs for creating and updating users which
+does not currently work with API Tokens. You must use basic auth.
```
inkel, line 17: Can we check somehow in the code that you're using Basic Auth and return a specific error or will it just fail?

inkel (approved):
Changes LGTM, my comments do not block this from merging in any way.

docs/data-sources/user.md
julienduchesne: Hmm I copied this from the `grafana_user` resource. Let me check if it's still relevant

docs/index.md
julienduchesne: It was from another merge. Forgot to run `go generate` in that other PR

docs/data-sources/user.md
julienduchesne: Hmm, it's still relevant but I think I'd rather rely on the API 403 (in case it ever changes)

PR #296, opened 2021-11-23
Check for docs drift

julienduchesne:
`go generate` needs to run to generate the docs but this is an often forgotten step
This should check that we don't push anything to master that hasn't had the docs generated


inkel (approved):
❤️ thanks! :shipit:

PR #297, opened 2021-11-24
Update the grafana version to 8.2.5 in tests

julienduchesne:
The `TestAccDashboard_folder` test would fail locally and I was wondering why
The `folder` variable in that test was never set and so the ID of that variable was always 0.
The `CheckDestroy` was always checking that the folder with ID 0 was destroyed
Something must have changed between 8.0.3 and 8.2.5 that affected folder IDs and that made the test fail

So the solution was adding the `testAccFolderCheckExists` from the `grafana_folder` tests. That function sets the folder variable after checking if the folder exists


inkel (approved):
Nice catch.

Issue #59, opened 2019-05-12
Allow for org_id field in resource definitions


bcampoli:
I think a basic feature that is missing in this provider is the ability to specify to which org you would like a resource to belong. I understand this is in part due to the construction of the Grafana API. The Grafana API is not RESTful, since it maintains an internal state variable of the `current_org` between API calls. This is makes implementing this feature slightly more difficult.

This is a crucial feature since internally, each Grafana API object is implicitly linked to an `org_id`. Therefore, in order to align the Terraform Provider with the data-model used by Grafana, the resources should also have an `org_id` field.

### Affected Resource(s)
Please list the resources as a list, for example:
- All resources except for `grafana_organization`

### Expected Behavior

The below example is just done with `grafana_data_source` but applies to every resource in the provider.

Given a sample resource configuration like this:
```
provider "grafana" {
  url  = "http://localhost:3000/"
  auth = "admin:admin"
}


resource "grafana_organization" "org1" {
  name         = "test-provider1"
  admin_user   = "admin"
}

resource "grafana_organization" "org2" {
  name         = "test-provider2"
  admin_user   = "admin"
}

resource "grafana_data_source" "metrics" {
  type          = "graphite"
  name          = "myapp-metrics-test"
  org_id        = "${grafana_organization.org1.id}"  
}

resource "grafana_data_source" "metrics2" {
  type          = "graphite"
  name          = "myapp-metrics-test2"
  org_id        = "${grafana_organization.org2.id}"  
}

resource "grafana_data_source" "metrics3" {
  type          = "graphite"
  name          = "myapp-metrics-test3"
  org_id        = "${grafana_organization.org1.id}"  
}

resource "grafana_data_source" "metrics4" {
  type          = "graphite"
  name          = "myapp-metrics-test4"
  org_id        = "${grafana_organization.org2.id}"  
}
```

We should expect for 2 datasources to be created in 2 different orgs, like so:
![image](https://user-images.githubusercontent.com/29064205/57585481-c46f2900-74b6-11e9-93df-5de426881dd1.png)

and

![image](https://user-images.githubusercontent.com/29064205/57585493-e8cb0580-74b6-11e9-8e16-a94997328778.png)

### Actual Behavior
An error because these fields are not available in the current version of the provider

### References
Related issues: https://github.com/terraform-providers/terraform-provider-grafana/issues/31#issuecomment-491044905

bcampoli:
### Pull Request

[...]

PR #303, opened 2021-12-08
add org_id to the resource grafana_user

vnaina:
Currently, this version does not allow the resource "grafana_user" to assign the user to an organization.
This commit adds this feature.

### DIFF
--- a/docs/resources/user.md
+++ b/docs/resources/user.md
@@ -42,6 +42,7 @@ resource "grafana_user" "staff" {
 - **is_admin** (Boolean) Whether to make user an admin. Defaults to `false`.
 - **login** (String) The username for the Grafana user.
 - **name** (String) The display name for the Grafana user.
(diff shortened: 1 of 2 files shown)


julienduchesne:
The idea seems good to me. Though:
1. We should also get this information when we read the user
2. It should also be tested (adding the attribute to verified attributes in the acceptance tests)

To make it clear, it is ok to have the Org ID here because users is an admin resource. Explained here: https://github.com/grafana/terraform-provider-grafana/issues/59#issuecomment-996935572

inkel:
Also please note that the documentation is out of date, please run `go generate` to update the docs.

julienduchesne:
I'll close this because it's been inactive for so long. Managing an org's users is already possible using the `grafana_organization` resource: https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/organization

PR #306, opened 2021-12-10
add message to dashboard creation

justinTM:
i'm not sure why message was left out before but i have a feeling there's a good reason and i'll find out soon :)

in parallel, i am updating the golang api to include Message in the Dashboard struct as well: https://github.com/grafana/grafana-api-golang-client/pull/40


julienduchesne (approved):
LGTM! Thank you for your contribution!

PR #311, opened 2021-12-17
Skip tests instead of using build tags

julienduchesne:
The current way of working is annoying, I have to remove the build tags to make my IDE work
Instead, we can skip the tests that we don't want to run. It will write a small message for everyone of these but at least, it's very clear what's happening (why tests aren't running)


inkel (approved):
Thanks for taking care!

PR #313, opened 2021-12-18
Testing: Consistent uid and id regexes

julienduchesne:
Centralized them to a single var. Also, most of them were invalid, checking for only a single char within the string


inkel (approved):
Love it, much simpler and elegant.

PR #319, opened 2021-12-21
Fix folder ID documentation

### LINKED ISSUE #262: Difference between grafana provider docs an implementation
### Terraform Version
* Terraform: 0.14.5
* Terraform Grafana Provider: latest

### Affected Resource(s)
* grafana_folder

### Expected Behavior
According to the grafana provider docs for the [grafana_folder](https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/folder) the resource supports an "id" parameter to force the folder id and getting consistent folder urls. 

### Actual Behavior
The parameter does not exist and it fails to plan. Here https://github.com/grafana/terraform-provider-grafana/blob/master/grafana/resource_folder.go#L29-L42 you can see that the parameter is missing 

One of the two is correct, but i'm not sure which one. I'd rather the docs be because i'm using this feature for dashboards already
and I wanted to start using it for folders.

inkel:
By adding the ID as a computed attribute it will show as a read-only attribute instead of the misleading comment that it might be possible to assign a value. This documentation seem to be generated automatically by the `tfplugindocs` tool, so we might need to either fix that or changing the way we generate the documentation.

### DIFF
--- a/docs/resources/folder.md
+++ b/docs/resources/folder.md
@@ -31,12 +31,9 @@ resource "grafana_dashboard" "metrics" {
 
 - **title** (String) The title of the folder.
 
-### Optional
-
-- **id** (String) The ID of this resource.
-
 ### Read-Only
 
+- **id** (String) unique identifier for this folder.
 - **uid** (String) Unique identifier.
 
 ## Import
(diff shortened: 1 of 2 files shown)


julienduchesne (approved):
LGTM modulo CI passing 
`go generate` seems to be out of date :)

inkel:
🤦🏽 thanks! Pushing the updated docs, will merge afterwards.

Issue #234, opened 2021-06-29
Alert Notifications secure_settings passing bool when api expects string


saltxwater:
### Terraform Version
* Terraform: 0.14.6
* Terraform Grafana Provider: 1.12.0
* Grafana: 7.5.7

### Affected Resource(s)
* grafana_alert_notification

### Terraform Configuration Files
```hcl
terraform {
  backend "local" {
    path = "localterraform.tfstate"
  }
  required_providers {
    grafana = {
      source = "grafana/grafana"
      version = "~> 1.12.0"
    }
  }
}

provider "grafana" {
  url = "<HOST>"
  auth = "<API KEY>"
}

resource "grafana_alert_notification" "slack" {
  name                    = "Slack"
  type                    = "slack"
  uid                     = "myUid"

  settings = {
    "url"         = "My Url"
  }

  secure_settings = {
    "url" = true
  }

}

```

### Debug Output
grafana_alert_notification.slack: Creating...

Error: status: 400, body: [{"classification":"DeserializationError","message":"json: cannot unmarshal bool into Go struct field CreateAlertNotificationCommand.secureSettings of type string"}]

  on notification-channels.tf line 2, in resource "grafana_alert_notification" "slack":
   2: resource "grafana_alert_notification" "slack" {

https://gist.github.com/saltxwater/ee3577a0e7449f021a8361d6cc20780d

### Expected Behavior
The Alert Notification to be created

### Actual Behavior
The request to grafana failed and no alert notification was created

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
As demonstrated here this fails when creating a new resource if you specify secure_settings. Linked to this:
- If you don't specify secure_settings and the notification has been saved in the grafana ui then secureFields will have { "url": true }. Which causes terraform to see the resource as always modified even after subsequent applies. You then can't add the secure_setting as you get the same error.
- _The grafana api for alert notifications is inconsistent; when reading secureFields it is a bool but expects strings for create and update: https://github.com/grafana/grafana/blob/main/pkg/models/alert_notifications.go **Edit: This is not correct - secureFields is correct to be bool but it is doing the create on secureSettings which IS meant to be strings**_

saltxwater:
Digging into this some more:
I'm a bit confused about what the format of secure_settings should be.

If you create the resource WITHOUT specifying secure_settings then the resource is created fine in grafana.
If you then go into the grafana ui, view the notification, and hit save then the PUT request is:
[...]

PR #322, opened 2021-12-22
Make secure_settings always pass values as string

inkel:
In #234 we have found that the secure settings for alert notifications must always be passed as a `map[string]string`, however for the case of `true` and `false` we were converting the values to booleans, which made Grafana HTTP API fail with a 400 Bad Request.

This change only removes the boolean conversions, so values are always passed as strings. A better fix that also involves the underlying API client would be to make this field a `map[string]string` instead of `map[string]interface`.

### DIFF
--- a/grafana/resource_alert_notification.go
+++ b/grafana/resource_alert_notification.go
@@ -224,14 +224,10 @@ func makeAlertNotification(_ context.Context, d *schema.ResourceData) (*gapi.Ale
 	secureSettings := map[string]interface{}{}
 	for k, v := range d.Get("secure_settings").(map[string]interface{}) {
 		strVal, ok := v.(string)
-		switch {
-		case ok && strVal == "true":
-			secureSettings[k] = true
-		case ok && strVal == "false":
-			secureSettings[k] = false
-		default:
-			secureSettings[k] = v
+		if !ok {
+			return nil, errors.New("secure_settings must be a map of string")
 		}
+		secureSettings[k] = strVal
 	}
 
 	sendReminder := d.Get("send_reminder").(bool)
--- a/grafana/resource_alert_notification_test.go
+++ b/grafana/resource_alert_notification_test.go
@@ -186,6 +186,9 @@ resource "grafana_alert_notification" "test" {
 			"uploadImage" = "false"
 			"autoResolve" = "true"
 		}
+	secure_settings = {
+		 "foo" = "true"
+	}
 }
 `
 


grafana/resource_alert_notification.go
```diff
@@ -224,14 +224,10 @@ func makeAlertNotification(_ context.Context, d *schema.ResourceData) (*gapi.Ale
...
-			secureSettings[k] = true
-		case ok && strVal == "false":
-			secureSettings[k] = false
-		default:
-			secureSettings[k] = v
+		if !ok {
+			return nil, errors.New("secure_settings must be a map of string")
 		}
+		secureSettings[k] = strVal
 	}
```
inkel, line 231: I've tried different alternatives and `v` can always be casted into a string, no matter what we pass before (e.g. if we defined the value as a list then Terraform will identify it's an invalid value and fail before reaching this line). Because of this, we could have replaced all this code by the following and rely on Go's compiler to raise an error if it cannot cast the value to a string, however, this error is raised as a `panic`, so I'd rather have a more controlled error returned to the user.

```go
secureSettings := make(map[string]interface{})
for k, v := range d.Get("secure_settings").(map[string]interface{}) {
    secureSettings[k] = v.(string)
}
```

julienduchesne (commented):
So, this only seems to be the case for `secure_settings`, right? Because we have functional tests that have `settings` with booleans in them. Could we have functional tests for secure settings as well?

inkel:
Correct. I've tried passing booleans in `settings` and that works correctly, so the tests are reliable 😸 

I've tried a regression test for `secure_settings` but couldn't come up with a good one, if we have

```terraform
resource "grafana_alert_notification" "foo" {
  # …
  secure_settings = {
    "foo" = true
  }
}
```

with the changes in this PR, this won't be passed anymore as a boolean but rather as a string, so it always work. Same if we had an integer or any other basic kind of value.

Now, if we were to have something like:

```terraform
resource "grafana_alert_notification" "foo" {
  # …
  secure_settings = {
    "foo" = [true, false]
  }
}
```

then Terraform will catch this earlier than us and fail with:

```
│ Error: Incorrect attribute value type
│
│   on issue_234.tf line 11, in resource "grafana_alert_notification" "issue234":
│   11:   secure_settings = {
│   12:     "url" = [true, false]
│   13:   }
│
│ Inappropriate value for attribute "secure_settings": element "url": string required.
```

julienduchesne:
Yep, but I mean testing that we don't revert the code in the future.
 Adding 
```
secure_settings = {
		"foo" = "true"
	}
```
to `testAccAlertNotificationConfig_basic` makes the tests fail:

```
=== RUN   TestAccAlertNotification_basic
    resource_alert_notification_test.go:20: Step 1/1 error: Error running apply: exit status 1
        
        Error: status: 400, body: [{"classification":"DeserializationError","message":"json: cannot unmarshal bool into Go struct field CreateAlertNotificationCommand.secureSettings of type string"}]
        
          with grafana_alert_notification.test,
          on terraform_plugin_test.tf line 2, in resource "grafana_alert_notification" "test":
           2: resource "grafana_alert_notification" "test" {
```


With your fix, it doesn't fail anymore and we'll be tested for this case in the future

inkel:
Oh, I see what you mean. Yeah, I can do that.

julienduchesne (approved):
LGTM!

PR #332, opened 2022-01-06
Synthetic Monitoring: Allow filtering out deprecated probes

julienduchesne:
I'd like to only use the current ones instead of the whole list


docs/data-sources/synthetic_monitoring_probes.md
```diff
@@ -14,13 +14,18 @@ Data source for retrieving all probes.
...
+  filter_deprecated = true
+}
 ```
 
 <!-- schema generated by tfplugindocs -->
 ## Schema
 
 ### Optional
 
+- **filter_deprecated** (Boolean) If true, only probes that are not deprecated will be returned. Defaults to `false`.
```
electron0zero: I would lean to keeping it to `true` by default. why? ideally we should not default to scheduling checks to deprecated probes.
```suggestion
- **filter_deprecated** (Boolean) If true, only probes that are not deprecated will be returned. Defaults to `true`.
```

grafana/data_source_synthetic_monitoring_probes.go
```diff
@@ -12,6 +12,12 @@ func dataSourceSyntheticMonitoringProbes() *schema.Resource {
 		Description: "Data source for retrieving all probes.",
 		ReadContext: dataSourceSyntheticMonitoringProbesRead,
 		Schema: map[string]*schema.Schema{
+			"filter_deprecated": {
+				Type:        schema.TypeBool,
+				Description: "If true, only probes that are not deprecated will be returned.",
+				Optional:    true,
+				Default:     false,
```
electron0zero:
```suggestion
				Default:     true,
```

docs/data-sources/synthetic_monitoring_probes.md
electron0zero: deprecated probes will eventually be removed and turned off.

electron0zero (approved):
other then default comment, rest lgtm 👍🏼

docs/data-sources/synthetic_monitoring_probes.md
julienduchesne: Yeah, I was worried about unexpected changes to users of the provider but on the other hand, when they are removed, users will have changes in any case.

👍 will default to true

Issue #294, opened 2021-11-23
CI for Enterprise Acceptance Tests


julienduchesne:
We started using the enterprise build tags for acceptance test files that need to run against an instance of Grafana Enterprise. These tests should be run in CI

PR #337, opened 2022-01-13
Bug fix: Add global field when updating a role.

vtorosyan:
When updating a fine-grained access control role, Grafana checks if the role at question has global scope or not.
If global field is not specified in the role definition, Grafana will always assume that the role is not global,
which will result to failure when updating global roles.

**Note for the reviewers**

The role update is a Grafana Enterprise feature, [these code lines](https://github.com/grafana/grafana-enterprise/blob/main/src/pkg/extensions/accesscontrol/database/database.go#L219-L221) are the guardians which will cause the failure without the `global` flat set correctly.

### DIFF
--- a/grafana/resource_role.go
+++ b/grafana/resource_role.go
@@ -182,6 +182,7 @@ func UpdateRole(ctx context.Context, d *schema.ResourceData, meta interface{}) d
 		r := gapi.Role{
 			UID:         d.Id(),
 			Name:        d.Get("name").(string),
+			Global:      d.Get("global").(bool),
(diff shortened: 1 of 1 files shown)


julienduchesne (approved):
Seems like the current tests should already cover this case but I actually don't know how to run them 😬 . I really have to get https://github.com/grafana/terraform-provider-grafana/issues/294 done

Jguer (approved):

Jguer:
Have run the tests, the one that covers this fails in the main branch, passes in this one :+1:

PR #339, opened 2022-01-14
SM: Use new `GetCheck` and `GetProbe` functions

julienduchesne:
Added in client 0.4.0: https://github.com/grafana/terraform-provider-grafana/pull/338
Makes code a bit cleaner


inkel (approved):
Nice! Definitely easier to read.

PR #352, opened 2022-01-25
Add importer to alert notification

RhynoVDS:
**Use Case**

We have more than 200 alert notifications which are used by hundreds of alert rules. We need the ability to import these existing alert notifications. Destroying and recreating is not an option as we would have to manually update hundreds of alert rules.


julienduchesne:
👋  Can you add an `import.sh` file in `examples/resources/grafana_alert_notification` (see other resources for examples). Then run `go generate` in the root of the repo

This will add the import statement to the docs

RhynoVDS:
thank you, I have made the change as requested.

julienduchesne (approved):
LGTM! Thank you for your contribution

Issue #342, opened 2022-01-20
Add terraform resource for Grafana Stacks


NasAmin:
### Problem
Grafana Cloud provides a [Stacks API](https://grafana.com/docs/grafana-cloud/reference/cloud-api/#stacks).
Customers using Grafana Cloud and want to manage multiple stacks/instances of Grafana currently rely on a manual process using the Grafana admin console. This is not sustainable for a scaling organisation.

Providing a terraform provider for creating new Grafana stacks will help users automate stack creation and closing the loop on manual processes.

### Proposal
Since there is already an API available, the following is proposed:

* Add support for the [Stacks API](https://grafana.com/docs/grafana-cloud/reference/cloud-api/#stacks) to the [Grafana API client for golang](https://github.com/grafana/grafana-api-golang-client)
* Use the above client library in this terraform provider and add a `grafana_cloud_stack` resource as well as a `grafana_cloud_stack` data source

ℹ️   I am happy to contribute

inkel:
@NasAmin that's a great idea! Feel free to submit a PR for it 😺

NasAmin:
@inkel Thanks. What would be the best way for testing any terraform changes locally for the [Stacks API](https://grafana.com/docs/grafana-cloud/reference/cloud-api/#stacks)? The README is not very clear on how to test the CloudAPI changes.

I am guessing I'll have to run acceptance tests against a real account on grafana.com?

julienduchesne:
> I am guessing I'll have to run acceptance tests against a real account on grafana.com?

Yes, that's how I would do it. I can add a `GRAFANA_CLOUD_TOKEN` into our CI so you could run tests there if you like. We already have a test accounts we run tests in

NasAmin:
@julienduchesne Thanks. 
That won't work on the fork, once I open the PR on the upstream repo, it will then use the `GRAFANA_CLOUD_TOKEN` correct?

julienduchesne:
> @julienduchesne Thanks. That won't work on the fork right, once I open the PR on the upstream repo, it will then use the `GRAFANA_CLOUD_TOKEN` correct?

Yes. You're free to work in a draft PR to test your changes though!

NasAmin:
@julienduchesne @inkel I have opened a [PR](https://github.com/grafana/grafana-api-golang-client/pull/53) on the grafana-api-golang-client repo. Please review 🙇

julienduchesne:
[...]

PR #354, opened 2022-01-25
Add Grafana Cloud stack resource and DataSource

NasAmin:
**Closes** #342 

🗒️  This PR is using a new version of the client api based on this [PR](https://github.com/grafana/grafana-api-golang-client/pull/53). 

### Resource
This PR introduces a `grafana_cloud_stack` resource. It does the following:
* Create a new stack
* Update an existing stack. Only stack `name`, `slug` and `description` can be updated.
* Changing `region_slug` with a valid value will result in `ForceNew`. This will remove the existing stack and create a new one in the desired region.
* Destroy will delete the stack from the target Grafana Cloud instance.
#### Validation
* Supports `eu`, `us` or `au` for region. The provider contains validation logic. Any other region will return a validation error like this if region was set to `ch`:
```
❯ terraform validate
╷
│ Error: region 'ch' is not supported. Only 'eu', 'us' and `au` are currently supported
│ 
│   with grafana_cloud_stack.test,
│   on grafana-stack.tf line 21, in resource "grafana_cloud_stack" "test":
│   21:   region_slug = "ch"
```

### Data Source
This PR also introduces a `grafana_cloud_stack` terraform data source.

### Examples Usage
```hcl

# Provider - url must be set to https://grafana.com
provider "grafana" {
  url = "https://grafana.com"

  # export TF_VAR_grafana_auth=...
  auth = var.grafana_auth
}

# Create a grafana stack
resource "grafana_cloud_stack" "test" {
  name   = "myteststack"
  slug   = "myteststack"
  region_slug = "eu"
  description = "here is a test description"
}

# Get an existing stack as a data source
data "grafana_cloud_stack" "test" {
  # slug of the desired stack
  slug = grafana_cloud_stack.test.slug
}
```


grafana/resource_stack.go
```diff
@@ -0,0 +1,274 @@
...
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Status of the Alertmanager instance configured for this stack.",
+			},
+		},
+	}
+}
+
+func CreateStack(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*client).gapi
```
julienduchesne: Since https://github.com/grafana/terraform-provider-grafana/pull/358, you can use
```suggestion
	client := meta.(*client).gcloudapi
```

julienduchesne (commented):
We've added support for a dedicated Grafana Cloud client. You can remove your changes to the .drone.yml and it should work

NasAmin:
> Details

@julienduchesne Thanks ever so much for creating a new client for us. I have applied the suggested changes but the tests are now failing with a `401`. Is the  `GRAFANA_CLOUD_API_KEY`  not set in the pipeline correctly?

Again I can use the provider locally and it is performing all the CRUD as expected.

julienduchesne:
I've opened https://github.com/grafana/terraform-provider-grafana/pull/364 to fix the secret. In the meantime, you can merge that branch into yours. Some tests are failing due to an invalid stack slug

grafana/resource_stack.go
```diff
@@ -0,0 +1,274 @@
...
+				Description: `Region slug to assign to this stack.
+Chaning region will destroy the existing stack and create a new one in the desired region`,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+					value := v.(string)
+					// Only acceptable regions are eu and us
+					if value != "eu" && value != "us" {
+						errors = append(errors, fmt.Errorf("region '%s' is not supported. Only 'eu' and 'us' are currently supported", value))
+					}
+					return
+				},
```
julienduchesne: You can use the `StringInSlice` helper for this
```suggestion
				ValidateFunc: validation.StringInSlice([]string{"au", "eu", "us"}, false),
```

grafana/resource_stack.go
```diff
@@ -0,0 +1,274 @@
...
+				Optional:    true,
+				Description: "Description of stack.",
+			},
+			"slug": {
+				Type:     schema.TypeString,
+				Required: true,
+				Description: `
+Subdomain that the Grafana instance will be available at (i.e. setting slug to “<stack_slug>” will make the instance
+available at “https://<stack_slug>.grafana.net".`,
+			},
```
julienduchesne: Since slugs can only support alphanumeric chars, you can have a validation at plan time

```suggestion
var stackSlugRegex = regexp.MustCompile("^[a-z0-9]+$")

.....

			"slug": {
				Type:     schema.TypeString,
				Required: true,
				Description: `
Subdomain that the Grafana instance will be available at (i.e. setting slug to “<stack_slug>” will make the instance
available at “https://<stack_slug>.grafana.net".`,
			},
			"slug": {
				Type:     schema.TypeString,
				Required: true,
				Description: `
Subdomain that the Grafana instance will be available at (i.e. setting slug to “<stack_slug>” will make the instance
available at “https://<stack_slug>.grafana.net".`,
				ValidateFunc: validation.StringMatch(stackSlugRegex, "must be a lowercase alphanumeric string"),
			},
```

grafana/resource_stack.go
```diff
@@ -0,0 +1,274 @@
...
+			"region_slug": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Description: `Region slug to assign to this stack.
+Chaning region will destroy the existing stack and create a new one in the desired region`,
```
julienduchesne:
```suggestion
Changing region will destroy the existing stack and create a new one in the desired region`,
```

NasAmin:
@julienduchesne thanks a lot for adding the secret for Cloud api.
I am now seeing errors like this:
Does the underlying API contain any additional limitations e.g. length of the slug?
Because the `slug` is used as part of the stack url, it seems like that url is already in use?
```

│ Error: status: 409, body: {
│   "code": "Conflict",
│   "message": "That url is not available"
│ }
│ 
│   with grafana_cloud_stack.test,
│   on grafana-stack.tf line 19, in resource "grafana_cloud_stack" "test":
│   19: resource "grafana_cloud_stack" "test" {
```

julienduchesne:
> @julienduchesne thanks a lot for adding the secret for Cloud api. I am now seeing errors like this: Does the underlying API contain any additional limitations e.g. length of the slug? Because the `slug` is used as part of the stack url, it seems like that url is already in use?
> 
> ```
> 
> │ Error: status: 409, body: {
> │   "code": "Conflict",
> │   "message": "That url is not available"
> │ }
> │ 
> │   with grafana_cloud_stack.test,
> │   on grafana-stack.tf line 19, in resource "grafana_cloud_stack" "test":
> │   19: resource "grafana_cloud_stack" "test" {
> ```

Probably that a test stack wasn't deleted from a previous run. I see the following:
- gcloudstacktest
- grafanacloudstacktest

Perhaps we should check that they exist and if they do, deleted them as a precondition for tests?

NasAmin:
> precondition

A precondition sounds sensible but what if the slug is used by someone else in another org? We won't be able to delete such stacks right?

I created the second stack i.e. `gcloudstacktest` locally to test it works. I then destroyed it. But it still seems to be lingering behind the scenes. This is probably due to DNS propagation taking time.

julienduchesne:
> > precondition
> 
> 
> 
> A precondition sounds sensible but what if the slug is used by someone else in another org? We won't be able to delete such stacks right?
> 
> 
> 
> I created the second stack i.e. `gcloudstacktest` locally to test it works. I then destroyed it. But it still seems to be lingering behind the scenes. This is probably due to DNS propagation taking time.

The test could add a random int suffix to the stack name. The precondition could list the stacks and delete the ones with the right prefix

NasAmin:
> The test could add a random int suffix to the stack name. The precondition could list the stacks and delete the ones with the right prefix

Thanks, a random suffix sounds interesting.
The pre-condition for checking if the stack exists sounds sensible. However, it won't help here because the stack actually doesn't exist and the API will say it doesn't. But it sounds like the DNS entry behind the stack url isn't getting cleaned up correctly. Not sure how to resolve that bit.

Maybe the underlying API needs to be reviewed to see how it does the cleanup?

julienduchesne:
I think the DNS entries do get cleaned up. The stacks still exist on Grafana.com. Due to a run that was cancelled before the delete happened maybe? 
image

NasAmin:
@julienduchesne thank you for the info. I have now implemented generating a random name for the stack. However, I am now seeing this error:
```

        
        Error: status: 403, body: {
          "code": "Forbidden",
          "message": "Hosted instance limit reached"
        }
        
          with grafana_cloud_stack.test,
          on terraform_plugin_test.tf line 2, in resource "grafana_cloud_stack" "test":
           2: 	resource "grafana_cloud_stack" "test" {
```

Apologies for all the chatter 🙇

julienduchesne:
> @julienduchesne thank you for the info. I have now implemented generating a random name for the stack. However, I am now seeing this error:
> 
> ```
> 
>         
>         Error: status: 403, body: {
>           "code": "Forbidden",
>           "message": "Hosted instance limit reached"
>         }
>         
>           with grafana_cloud_stack.test,
>           on terraform_plugin_test.tf line 2, in resource "grafana_cloud_stack" "test":
>            2: 	resource "grafana_cloud_stack" "test" {
> ```
> 
> Apologies for all the chatter 🙇

Yes, because the stacks aren't being cleaned up. This should fail on your own account as well. For the moment, I've cleaned up our test account

NasAmin:
@julienduchesne I have tested these changes when running the tests locally. I am not seeing the errors seen on the latest build locally.

Your guidance will be much appreciated as I am not sure how to debug the current issues.

julienduchesne:
> @julienduchesne I have tested these changes when running the tests locally. I am not seeing the errors seen on the latest build locally.
> 
> 
> 
> Your guidance will be much appreciated as I am not sure how to debug the current issues.
> 
> 

👀 I'll take a look today or tomorrow and try running them locally with the CI creds

julienduchesne:
Hey @NasAmin. I fixed the tests here: https://github.com/grafana/terraform-provider-grafana/pull/368/commits/d43ad4a34421113d72831f56f81d2a1934fc8d98
See this build: https://drone.grafana.net/grafana/terraform-provider-grafana/624/3/2

There were a few issues that I fixed:
- Added a pre-cleanup to delete test instances from previous failed runs
- Wrong client was being used for the datasource
- Added sleep time at the end of tests. It's unfortunate but I think the stacks were just not fully provisioned when being deleted and that caused the failures

Also, I released v0.2.6 of the client, so you'll be able to pull that. So you can cherry-pick my commit, update the client. Then, I'll do a final review and we can push this through!

NasAmin:
> There were a few issues that I fixed:
> 
> * Added a pre-cleanup to delete test instances from previous failed runs
> * Wrong client was being used for the datasource
> * Added sleep time at the end of tests. It's unfortunate but I think the stacks were just not fully provisioned when being deleted and that caused the failures
> 
> Also, I released v0.2.6 of the client, so you'll be able to pull that. So you can cherry-pick my commit, update the client. Then, I'll do a final review and we can push this through!

@julienduchesne Thanks a lot for diagnosing and fixing the issues. 
I have applied your suggested changes and the [build now passes](https://drone.grafana.net/grafana/terraform-provider-grafana/633/3/2). 🎉 🎉

grafana/resource_stack.go
```diff
@@ -0,0 +1,271 @@
...
+)
+
+var stackSlugRegex = regexp.MustCompile("^[a-z][a-z0-9]+$")
+
+func ResourceStack() *schema.Resource {
+	return &schema.Resource{
+
+		Description: `
+* [Official documentation](https://grafana.com/docs/grafana-cloud/reference/cloud-api/#stacks/)
+* [HTTP API](https://grafana.com/docs/grafana/latest/http_api/stack/)
```
dohnto: This link returns 404 for me.

NasAmin: @dohnto Thank you! I put that as a placeholder. What would be the right link? (I couldn't find API docs for stacks apart from the first link).

NasAmin: @julienduchesne Shall I remove the `HTTP API` link given that are no docs for it currently?

julienduchesne: Yes

julienduchesne (approved):
Alright. Other than that dead link. This LGTM! Great work @NasAmin!

NasAmin:
> Alright. Other than that dead link. This LGTM! Great work @NasAmin!

Splendid! I have removed the broken link. 

Thank you so much for all the help and your patience 👍 

I'll keep an eye on for this to be merged. We can then start building our internal workflows 🚀

julienduchesne:
> > Alright. Other than that dead link. This LGTM! Great work @NasAmin!
> 
> Splendid! I have removed the broken link.
> 
> Thank yo so much for all the help and your patience 👍
> 
> I'll keep an on for this to be merged. We can then start building our internal workflows 🚀

Merging this right away 😄 . I'll be doing a release once the library panel feature is complete: https://github.com/grafana/terraform-provider-grafana/pull/360. This should be imminent

justinTM:
@julienduchesne the holdup with the data source Library Panel is that I'm getting "404 not found" errors in my tests. My question:

When the test plans and applies the **resource** , and I want to reference this newly-created resource in the **datasource**, will it 404 if I include both of these Terraform blocks in the same file/test? I assumed I could immediately grab the library panel (by name or UID) at the same time and Terraform would do the dependency work to make it happen but I've looked the code over and it's exactly the same for Datasource read resource as it is for the Resource read resource.

PR #358, opened 2022-01-27
Add client for Grafana Cloud

julienduchesne:
This allows users to use at least 1 of Grafana, Grafana Cloud or Synthetic Monitoring

For now, the client is from https://github.com/grafana/grafana-api-golang-client. We can put cloud resources there for now. Because in any case, clients will all shift around when we get a Grafana client generated from OpenAPI

TODO: Add validations on resources to indicate which token is required. For example, someone using a cloud resource without setting a cloud token should get a clear error


julienduchesne:
This would be used to enable https://github.com/grafana/terraform-provider-grafana/pull/354

julienduchesne:
Ignore whitespace in your review!

inkel (approved):
Great work!

PR #331, opened 2022-01-06
Add Library Panel Resource

justinTM:
This will be my first substantial Go code, and it also relies on some yet-nonexistent grafana go API additions for library panels

PR #365, opened 2022-01-31
Add function to check semver for OSS tests

julienduchesne:
Will be used for library panels: https://github.com/grafana/terraform-provider-grafana/pull/331. This is a feature from v8 so the tests on v7 shouldn't run
This line can be added to the test: `CheckOSSTestsSemver(t, ">=8.0.0")`
If a feature is deprecated from Grafana in the future, we can add a lower than semver


julienduchesne:
If you go into the first commit build: https://drone.grafana.net/grafana/terraform-provider-grafana/558, you can see that I skipped a test for v7

inkel (approved):
Great feature to have! Thanks

PR #369, opened 2022-02-02
Update resource_builtin_role_assignment resource to ignore assignments defined in the server side

### LINKED ISSUE #362: BuiltInRole deploy not idempotent
**Terraform Version**
Terraform v0.14.4
Terraform Grafana Provider 1.12.0
Grafana Enterprise v8.3.5

**Affected Resource(s)**
grafana_builtin_role_assignment

**Terraform config**
```
      "grafana_builtin_role_assignment": {
         "editor": {
            "builtin_role": "Editor",
            "roles": [
               {
                  "global": false,
                  "uid": "${grafana_role.report_viewer.uid}"
               }
            ]
         }
      },
      "grafana_role": {
         "report_viewer": {
            "description": "Can view reports",
            "name": "Report Viewer",
            "permissions": [
               {
                  "action": "reports:read",
                  "scope": "reports:*"
               }
            ],
            "version": 1
         }
      }
```

**Expected Behavior**
`terraform deploy` should create the role + assignment to Editor on first deploy, and make no changes on second deploy.

**Actual Behavior**
As expected on first deploy, but on second deploy all built-in-role-assignments for Editor are destroyed, including the default assignments, in our case the assignment of datasources:explore to Editor, and only the built-in role assignment specified in our config is recreated.

vtorosyan:
## Context

When Grafana starts, by default it creates built-in role assignments for fixed roles. Additionally, users can create built-in role assignments via API. In these scenarios, when the resource is not properly imported, terraform will try to destroy what is in server side, but not in the configuration.

The change ensures that this does not happen as it can cause unintended behaviour for users.

## Implementation details

The only workaround, with the given resources is to do in advance check to see if the fetched assignments has been ever in the state, or is present in the configuration, otherwise ignore it.

## Testing

I have manually tested the following scenarios:
- Having built-in role assignments in server, and creating a new assignment in terraform
- Updating the above created assignments after it's first time creation
- Removing roles from the above assignments
- Destroying entire resource -> in this scenario what has not been defined in the config, won't be destroyed from the backend.

Additionally, the newly added test covers the scenario mentioned in the issue.

## Related issues

Fixes https://github.com/grafana/terraform-provider-grafana/issues/362

### DIFF
--- a/grafana/resource_builtin_role_assignment.go
+++ b/grafana/resource_builtin_role_assignment.go
@@ -108,9 +108,9 @@ func ReadBuiltInRole(ctx context.Context, d *schema.ResourceData, meta interface
 	}
 	roles := make([]interface{}, 0)
 	for _, br := range brRole {
-		// It is possible on the server side there are roles assigned to the built-in role which were never in Terraform state,
-		// and are not now in the current configuration. To prevent unintended behaviour, such as destroying remove resources,
-		// the following check ensures to only consider roles which are either in the state or in the config.
+		// It is possible that in the server side there are roles assigned to the built-in role which were never in Terraform state,
+		// and are not in the current configuration. The following check ensures that we only consider roles which are either in the state or in the config.
+		// This prevents unintended behaviour, such as destroying the assignments which were never managed by Terraform.
 		_, isInState := stateRoles[br.UID]
 		_, isInConfig := configRoles[br.UID]
 		if !isInState && !isInConfig {
--- a/grafana/resource_builtin_role_assignment.go
+++ b/grafana/resource_builtin_role_assignment.go
@@ -102,8 +102,20 @@ func ReadBuiltInRole(ctx context.Context, d *schema.ResourceData, meta interface
 		return nil
 	}
 
+	stateRoles, configRoles, err := collectRoles(d)
+	if err != nil {
+		return diag.FromErr(err)
+	}
 	roles := make([]interface{}, 0)
 	for _, br := range brRole {
+		// It is possible on the server side there are roles assigned to the built-in role which were never in Terraform state,
+		// and are not now in the current configuration. To prevent unintended behaviour, such as destroying remove resources,
+		// the following check ensures to only consider roles which are either in the state or in the config.
+		_, isInState := stateRoles[br.UID]
+		_, isInConfig := configRoles[br.UID]
+		if !isInState && !isInConfig {
+			continue
+		}
 		rm := map[string]interface{}{
 			"uid":    br.UID,
 			"global": br.Global,
--- a/grafana/resource_builtin_role_assignment_test.go
+++ b/grafana/resource_builtin_role_assignment_test.go
@@ -48,6 +48,147 @@ func TestAccBuiltInRoleAssignment(t *testing.T) {
 	})
 }
 
+func TestAccBuiltInRoleAssignmentUpdate(t *testing.T) {
+	CheckEnterpriseTestsEnabled(t)
+
+	var br gapi.BuiltInRoleAssignment
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: testAccProviderFactories,
+		CheckDestroy:      testAccBuiltInRoleAssignmentCheckDestroy(&br),
+		Steps: []resource.TestStep{
+			{
+				PreConfig: func() {
+					err := prepareDefaultAssignments()
+					if err != nil {
+						t.Errorf("error when creating built-in role ssignments %s", err)
+					}
+				},
+				Config: builtInRoleAssignmentUpdatePreConfig,
+				Check: resource.ComposeTestCheckFunc(
+					testAccBuiltInRoleAssignmentCheckExists("grafana_builtin_role_assignment.test_builtin_assignment"),
+					resource.TestCheckResourceAttr(
+						"grafana_builtin_role_assignment.test_builtin_assignment", "builtin_role", "Viewer",
+					),
+					resource.TestCheckResourceAttr(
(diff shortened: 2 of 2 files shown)


grafana/resource_builtin_role_assignment_test.go
```diff
@@ -48,6 +48,147 @@ func TestAccBuiltInRoleAssignment(t *testing.T) {
...
 
+func TestAccBuiltInRoleAssignmentUpdate(t *testing.T) {
+	CheckEnterpriseTestsEnabled(t)
+
+	var br gapi.BuiltInRoleAssignment
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: testAccProviderFactories,
+		CheckDestroy:      testAccBuiltInRoleAssignmentCheckDestroy(&br),
```
julienduchesne: Should the `CheckDestroy` function also check that the default assignements weren't destroyed?

julienduchesne (approved):
The solution seems like a good one to me and it's well tested. I can't test it on my side but it LGTM!

grafana/resource_builtin_role_assignment_test.go
```diff
@@ -48,6 +48,147 @@ func TestAccBuiltInRoleAssignment(t *testing.T) {
...
+	if err != nil {
+		return fmt.Errorf("error creating role: %s", err)
+	}
+	_, err = client.NewBuiltInRoleAssignment(gapi.BuiltInRoleAssignment{
+		BuiltinRole: "Viewer",
+		RoleUID:     role1.UID,
+		Global:      false,
+	})
+	if err != nil {
+		return fmt.Errorf("error creating built-in role assigntment: %s", err)
```
Jguer:
```suggestion
		return fmt.Errorf("error creating built-in role assigntment: %w", err)
```

nit: just so error un/wrapping still succeeds

grafana/resource_builtin_role_assignment_test.go
```diff
@@ -48,6 +48,147 @@ func TestAccBuiltInRoleAssignment(t *testing.T) {
...
+	}
+	_, err = client.NewBuiltInRoleAssignment(gapi.BuiltInRoleAssignment{
+		BuiltinRole: "Viewer",
+		RoleUID:     role2.UID,
+		Global:      false,
+	})
+	if err != nil {
+		return fmt.Errorf("error creating built-in role assigntment: %s", err)
```
Jguer:
```suggestion
		return fmt.Errorf("error creating built-in role assigntment: %w", err)
```

nit: just so error un/wrapping still succeeds

grafana/resource_builtin_role_assignment_test.go
```diff
@@ -48,6 +48,147 @@ func TestAccBuiltInRoleAssignment(t *testing.T) {
...
+			},
+		},
+	}
+	role1, err := client.NewRole(r1)
+	if err != nil {
+		return fmt.Errorf("error creating role: %s", err)
+	}
+	role2, err := client.NewRole(r2)
+	if err != nil {
+		return fmt.Errorf("error creating role: %s", err)
```
Jguer:
```suggestion
		return fmt.Errorf("error creating role: %w", err)
```

nit: just so error un/wrapping still succeeds

grafana/resource_builtin_role_assignment_test.go
```diff
@@ -48,6 +48,147 @@ func TestAccBuiltInRoleAssignment(t *testing.T) {
...
+		Global:  true,
+		Permissions: []gapi.Permission{
+			{
+				Action: "users:create",
+			},
+		},
+	}
+	role1, err := client.NewRole(r1)
+	if err != nil {
+		return fmt.Errorf("error creating role: %s", err)
```
Jguer:
```suggestion
		return fmt.Errorf("error creating role: %w", err)
```

nit: just so error un/wrapping still succeeds

Jguer (approved):

grafana/resource_builtin_role_assignment_test.go
vtorosyan: Really good catch. I actually just realized that `testAccBuiltInRoleAssignmentCheckDestroy` checks nothing, as the pointer to `&br` never updates to the created built-in role assignment, so we always pass an empty struct to the method, which returns success on it's turn as there is indeed no assignment for empty built-in role. 

Working on a fix.

vtorosyan: 5566c8b4691e45c25d4b1383e155229156eda19f

grafana/resource_builtin_role_assignment.go
```diff
@@ -102,8 +102,20 @@ func ReadBuiltInRole(ctx context.Context, d *schema.ResourceData, meta interface
...
 	}
 
+	stateRoles, configRoles, err := collectRoles(d)
+	if err != nil {
+		return diag.FromErr(err)
+	}
 	roles := make([]interface{}, 0)
 	for _, br := range brRole {
+		// It is possible on the server side there are roles assigned to the built-in role which were never in Terraform state,
+		// and are not now in the current configuration. To prevent unintended behaviour, such as destroying remove resources,
```
inkel: I'm not sure I understand what _destroying remove resources_ means, could you rephrase this? Thank you.

inkel (commented):
I've just left a minor request. My initial review looks good, but I want to give it another round to fully understand what are we testing here.

Thank you so much for your hard work!

grafana/resource_builtin_role_assignment.go
vtorosyan: Good point, I can see it was unclear and wrong. 396491caf72db7241154901dd00ce33d83069d9d is this a bit better?

inkel: Yup, much better, thank you!

PR #360, opened 2022-01-28
add data_source_library_panel

justinTM:

PR #370, opened 2022-02-03
add data_source_library_panel

justinTM:
squashed commits


justinTM:
@julienduchesne i think it's ready for review.

the holdup was writing new features to the normalize function to suppress diffs for dashboards with connected library panels. Grafana will populate a significant amount of metadata, but only requires `name` and `uid` for each library panel, so i remove all except these inside that function to avoid diffs for that use case.

additionally, `id` is now removed from each panel due to diffs as well. The part I'm worried about is the fact the normalize function doesn't return any error (suppresses them instead) so potentially erratic JSON changes **could** result in the future, but probably only if the Grafana dashboard schema changes. I assume because the normalize dashboard JSON function is used as the StateFunc for the dashboard resource, it can't return an error? not sure.

other than that I think it's finally reached a state where users can make library panels and connect them to dashboards, as well as ingest user-created (outside of Terraform) dashboards; the two use cases I have been trying to accomplish for the past month or so. Breath of relief! 

Cheers and thanks again for your help.

inkel:
@justinTM this is the same as #360 but rebased, is that correct?

examples/data-sources/grafana_library_panel/bad-ExactlyOneOf.tf
```diff
@@ -0,0 +1,14 @@
+resource "grafana_library_panel" "test_bad_inputs" {
+  name       = "test name"
+  model_json = jsonencode({
+    gridPos  = {
+      h      = 8
+      w      = 12 }
```
inkel: Can you run `terraform fmt` in this directory please? Thanks!

justinTM:
basically yeah! then i added the final example terraform code:

make a library panel, create a dashboard with it (Grafana makes the connection between the two), get the dashboard ID from the library panel resource after dashboard creation, then finally retrieve the dashboard by ID with datasource to make sure it all works as expected.

go.mod
```diff
@@ -12,4 +12,7 @@ require (
 	github.com/hashicorp/go-cty v1.4.1-0.20200414143053-d3edf31b6320
 	github.com/hashicorp/terraform-plugin-docs v0.5.1
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.10.1
+	github.com/tidwall/sjson v1.2.4
 )
+
+replace github.com/grafana/grafana-api-golang-client v0.2.6 => github.com/justinTM/grafana-api-golang-client v0.2.2-0.20220203033248-b89448800226
```
inkel: Is this actually required? I've just checked your fork the `grafana-api-golang-client` and GitHub says it's actually behind the official one.

justinTM:
@inkel since i truly don't know Go well enough to test locally, I have been really leaning on the cloud CI and so i have a ton of commits /shrug

grafana/data_source_library_panel_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+func TestAccDatasourceLibraryPanel(t *testing.T) {
+	CheckOSSTestsEnabled(t)
+	CheckOSSTestsSemver(t, ">=8.0.0")
+
+	var panel gapi.LibraryPanel
+	// var dashboard gapi.Dashboard
+	checks := []resource.TestCheckFunc{
+		testAccLibraryPanelCheckExists("grafana_library_panel.test", &panel),
+		// testAccDashboardCheckExists("grafana_dashboard.test", &dashboard),
```
inkel: Please remove this line if it will be commented out.

grafana/resource_library_panel.go
```diff
@@ -162,6 +172,18 @@ func ReadLibraryPanel(ctx context.Context, d *schema.ResourceData, meta interfac
...
+	connections, err := client.LibraryPanelConnections(uid)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	dashboardIds := []int64{}
+	for _, connection := range *connections {
+		dashboardIds = append(dashboardIds, connection.DashboardID)
+	}
+	// // return diag.Errorf("%#v", dashboardIds)
```
inkel: Please remove commented out lines 😸

grafana/resource_library_panel.go
```diff
@@ -162,6 +172,18 @@ func ReadLibraryPanel(ctx context.Context, d *schema.ResourceData, meta interfac
 	d.Set("created", panel.Meta.Created.String())
 	d.Set("updated", panel.Meta.Updated.String())
 
+	connections, err := client.LibraryPanelConnections(uid)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	dashboardIds := []int64{}
```
inkel: Small performance tip: given that we know the final length of this slice, please use the following:
```suggestion
	dashboardIds := make)[]int64, 0, len(*connections))
```

This will reduce the number of allocations Go has to make to create this slice.

inkel (changes requested):
First of all, thank you **so much** for all your contributions.

I'm requesting some changes here, mostly small things and nothing really blocking from getting this merged, just making the code more consistent with existing codebase and minor minor best practices.

I still need to evaluate how I feel about introducing the `github.com/tidwall/sjson` dependency, I'll be re-reviewing this again later today.

inkel:
> @inkel since i truly don't know Go well enough to test locally, I have been really leaning on the cloud CI and so i have a ton of commits /shrug

Don't worry about it! For not knowing Go well enough you're producing lots of good code in this repository 😸 

I don't mind having lots of commits, with time you'll master that art.

go.mod
justinTM: https://github.com/grafana/grafana-api-golang-client/pull/55

this PR is because i opened a PR to fix the API docs to fix a 404 error with library elements by name returning array instead of single object. I will resolve Julien's review comments there (forgot about them!) and then tackle these here.

examples/data-sources/grafana_library_panel/bad-ExactlyOneOf.tf
justinTM: should I align these `=` signs or leave as the formatter suggests? (see below)
image

grafana/data_source_library_panel_test.go
justinTM: https://github.com/grafana/terraform-provider-grafana/pull/370/commits/038acea6f902be6d643957d3940d691cba8f33a0

examples/data-sources/grafana_library_panel/bad-ExactlyOneOf.tf
inkel: Just run `terraform fmt examples/data-sources/grafana_library_panel/` and it will make the pertinent changes. In #372 I'm doing the same for all the other directories containing `.tf` files.

grafana/resource_library_panel.go
justinTM: https://github.com/grafana/terraform-provider-grafana/pull/370/commits/c0a3f5b67f02b9a99ad593b4749fba258a9d44e3

examples/data-sources/grafana_library_panel/bad-ExactlyOneOf.tf
justinTM: https://github.com/grafana/terraform-provider-grafana/pull/370/commits/89cfdaeeb6adc38568c06e2b6ea36c7381c95188

grafana/resource_library_panel.go
justinTM: :) https://github.com/grafana/terraform-provider-grafana/pull/370/commits/7e5c9c247ea5059268d153beb3f3f33eb73311a1

go.mod
inkel: I see 🤔 

Let's focus on that one first and then I'll come back to this one.

julienduchesne (approved):
LGTM! I updated the client to the released tag, and made a small change to avoid adding a dependency but it's ready to go IMO now!

Great work @justinTM !

PR #371, opened 2022-02-03
add example/test for dashboard with connected library panel

justinTM:


justinTM:
see https://github.com/grafana/terraform-provider-grafana/pull/370

Issue #348, opened 2022-01-24
Resource to control the new style of alerting provided in Grafana 8.X


logan-bobo:
Currently, the only resource for managing alert notifications is `grafana_alert_notification`. In the newer version of grafana, this provider does not look to have a way to manage contact points - (https://grafana.com/docs/grafana/latest/alerting/unified-alerting/contact-points/). Happy to actually pick this up but may need some initial direction. Would love to contribute.

julienduchesne:
Alerting is tracked in this issue: https://github.com/grafana/terraform-provider-grafana/issues/57

PR #381, opened 2022-02-07
Remove changelog entries for v1.14.0+

julienduchesne:
The changelog renderer doesn't know the difference between a closed issue and a resolved one so it adds confusing things to the log
Ex: we currently have `Resource to control the new style of alerting provided in Grafana 8.X #348` in the v1.19.0 changelog but that's only going to be released with Grafana 8.5
The release notes builder that comes in Github Releases produces a much nicer result so let's only use that
It also simplifies the release process, no need for a Github token


inkel (approved):
Good call!

PR #385, opened 2022-02-10
Fix `grafana_team` members reordering

### LINKED ISSUE #384: `grafana_team` `members` ordering causes terraform changes
### Terraform Version
* Terraform: 1.1.5
* Terraform Grafana Provider: 1.19.0
* Grafana: v8.3.3 (30bb7a93ca)

### Affected Resource(s)
* grafana_team

```hcl
locals {
  testers = {
    "A"  = "a@example.com"
    "B" = "b@example.com"
    "C"  = "c@example.com"
  }
}

resource "grafana_user" "testers" {
  for_each = local.testers

  name     = each.key
  email    = each.value
  login    = each.value
  password = "PASSWORD"
  is_admin = false
}

resource "grafana_team" "testers" {
  name = "Testing team"

  # Re-ordering the members here would cause a terraform change that is not persisted.
  # i.e. The initial order will be kept even if we `terraform apply` a different order.
  members = [
    local.testers["A"],
    local.testers["B"],
    local.testers["C"],
  ]

  depends_on = [grafana_user.testers]
}

```

### Expected Behavior
If we reorder `members`, e.g.

```hcl
  members = [
    local.testers["C"],
    local.testers["A"],
    local.testers["B"],
  ]
```

`terraform apply` should apply and save the new ordering. Subsequent `terraform apply` should show no changes.

### Actual Behavior

There is always a change as follows,

```
  # grafana_team.testers will be updated in-place
  ~ resource "grafana_team" "testers" {
        id      = "3"
      ~ members = [
          + "c@example.com",
            "a@example.com",
            "b@example.com",
          - "c@example.com",
        ]
        name    = "Testing team"
        # (1 unchanged attribute hidden)
    }
```

julienduchesne:
It's managed as a set by the Grafana API (and client) but we have it as a list in the resource
This means that whenever the ordering change, it's not being pushed to the API and we have a drift (see issue for more info)

This makes the members attribute a set. Migration is not an issue since both lists and sets are represented as JSON lists in the state (I tested it)

Closes https://github.com/grafana/terraform-provider-grafana/issues/384

### DIFF
--- a/docs/resources/team.md
+++ b/docs/resources/team.md
@@ -34,7 +34,7 @@ resource "grafana_team" "test-team" {
 
 - **email** (String) An email address for the team.
 - **id** (String) The ID of this resource.
-- **members** (List of String) A list of email addresses corresponding to users who should be given membership
+- **members** (Set of String) A set of email addresses corresponding to users who should be given membership
 to the team. Note: users specified here must already exist in Grafana.
 
 ### Read-Only
--- a/grafana/resource_team.go
+++ b/grafana/resource_team.go
@@ -35,7 +35,6 @@ func ResourceTeam() *schema.Resource {
 * [Official documentation](https://grafana.com/docs/grafana/latest/manage-users/manage-teams/)
 * [HTTP API](https://grafana.com/docs/grafana/latest/http_api/team/)
 `,
-
 		CreateContext: CreateTeam,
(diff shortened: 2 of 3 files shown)


inkel (approved):
Thank you!

PR #392, opened 2022-02-16
Fix missing `org_id` of resource `grafana_organization`

pallxk:
## Test terraform config

```tf
terraform {
  required_providers {
    grafana = {
      source = "grafana/grafana"
    }
  }
}

resource "grafana_organization" "org_2" {
  name = "org_2"
}

output "org_id" {
  value = grafana_organization.org_2.org_id
}
```

## Before

```sh
$ terraform output
╷
│ Warning: No outputs found
│ 
│ The state file either has no outputs defined, or all the defined outputs are empty. Please define an output in your configuration with the `output` keyword and run `terraform refresh` for it to become available. If you are
│ using interpolation, please verify the interpolated value is not empty. You can use the `terraform console` command to assist.
╵
```

## After

```sh
$ terraform output
org_id = 2
```

### DIFF
--- a/grafana/resource_organization.go
+++ b/grafana/resource_organization.go
@@ -155,6 +155,7 @@ func ReadOrganization(ctx context.Context, d *schema.ResourceData, meta interfac
 	if err != nil {
 		return diag.FromErr(err)
 	}
+	d.Set("org_id", resp.ID)
 	d.Set("name", resp.Name)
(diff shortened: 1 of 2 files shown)


julienduchesne (approved):
Thanks for your contribution @pallxk! I also added a test which lead me to find another bug 😄

pallxk:
Great to have things tested! :smiley:

PR #408, opened 2022-03-04
Rename stack resources to include `Cloud`

julienduchesne:
I'll be adding some other cloud resources today, this will make the repo more tidy


inkel (approved):
Oohh this is a nice idea!

PR #410, opened 2022-03-04
Update Drone pipelines

julienduchesne:
- Add concurrency limit to the cloud pipeline. It unfortunately doesn't work when multiple versions run at the same time so let's restrict that
- Update Grafana versions used in pipelines (Added 8.4.x and dropped 8.0.x, I think 4 minor versions is enough 😄 )

### DIFF
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -1,4 +1,4 @@
-local grafanaVersions = ['8.3.3', '8.2.7', '8.1.8', '8.0.7', '7.5.12'];
+local grafanaVersions = ['8.4.3', '8.3.5', '8.2.7', '8.1.8', '7.5.15'];
 local images = {
   go: 'golang:1.16',
   lint: 'golangci/golangci-lint',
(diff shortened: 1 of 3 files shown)


inkel (approved):

inkel:
> * Update Grafana versions used in pipelines (Added 8.4.x and dropped 8.0.x, I think 4 minor versions is enough 😄 )

Agreed on this one! But I wonder if it wouldn't be possible to fetch this versions automatically instead of hard-coding them, given that Grafana has a new release quite often 😸

julienduchesne:
> > * Update Grafana versions used in pipelines (Added 8.4.x and dropped 8.0.x, I think 4 minor versions is enough 😄 )
> 
> Agreed on this one! But I wonder if it wouldn't be possible to fetch this versions automatically instead of hard-coding them, given that Grafana has a new release quite often 😸

Yeah, that'd be nice. Unfortunately, we can't generate pipelines on the fly BUT we could have a cron that updates the list of versions in the yaml file. I'll open an issue

PR #413, opened 2022-03-07
Separate Cloud and Cloud instance tests

julienduchesne:
As I am adding more resources and options for Cloud, I'm running into this annoying thing
I want to make sure that the tests do NOT use a grafana auth token because I'm generating one from the Cloud API
However, the `cloud` tests that we currently have include both API and "Pro features" tests and therefore, they need a Cloud token for some tests and a Grafana auth for others

This PR separates the API and instance tests into two different make targets and pipelines

### DIFF
--- a/grafana/provider_test.go
+++ b/grafana/provider_test.go
@@ -244,15 +244,15 @@ func IsUnitTest(t *testing.T) {
 func CheckOSSTestsEnabled(t *testing.T) {
 	t.Helper()
 
+	if !accTestsEnabled(t, "TF_ACC_OSS") {
+		t.Skip("TF_ACC_OSS must be set to a truthy value for OSS acceptance tests")
+	}
+
 	checkEnvVarsSet(t, []string{
 		"GRAFANA_URL",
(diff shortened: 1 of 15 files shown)


grafana/provider_test.go
```diff
@@ -248,6 +223,16 @@ func accTestsEnabled(t *testing.T, envVarName string) bool {
 	return enabled
 }
 
+func checkEnvVarsSet(t *testing.T, envVars []string) {
```
inkel: We could make this accept variadic arguments instead, and then you just call it like `checkEnvVarsSet(t, "FOO", "BAR", "QUUX")`
```suggestion
func checkEnvVarsSet(t *testing.T, envVars ...string) {
```

WDYT?

inkel (approved):
LGTM! I've left a suggestion, but feel free to ignore.

grafana/provider_test.go
julienduchesne: Yes, that's much cleaner, thanks

PR #414, opened 2022-03-07
Upgrade to go 1.17

julienduchesne:
Also, removed `ioutil` references which is deprecated as of 1.16
https://pkg.go.dev/io/ioutil
> As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.


jvrplmlmn (approved):
👍

inkel (approved):
THANK YOU!

PR #421, opened 2022-03-08
Add readiness check for cloud stacks

julienduchesne:
The API doesn't expose any indication that a stack is actually ready
There is a `status` field on the API response (GETing the stack) but it shows "active" immediately after POST

So this PR adds a readiness check using the created Grafana instance URL. Whenever the Grafana instance returns a 200 (from the login screen), it's ready
This check can also be disabled by setting `wait_for_readiness` to false on the resource

### DIFF
--- a/grafana/resource_cloud_stack.go
+++ b/grafana/resource_cloud_stack.go
@@ -2,14 +2,19 @@ package grafana
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"log"
+	"net/http"
 	"path"
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	gapi "github.com/grafana/grafana-api-golang-client"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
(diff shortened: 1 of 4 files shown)


grafana/resource_cloud_stack.go
```diff
@@ -71,6 +76,12 @@ Changing region will destroy the existing stack and create a new one in the desi
 				Optional:    true,
 				Description: "Custom URL for the Grafana instance. Must have a CNAME setup to point to `.grafana.net` before creating the stack",
 			},
+			"wait_for_readiness": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     true,
+				Description: "Whether to wait for readiness of the stack after creating it. The check is a simple GET request to the stack URL (Grafana instance).",
```
dohnto:
```suggestion
				Description: "Whether to wait for readiness of the stack after creating it. The check is a simple HEAD request to the stack URL (Grafana instance).",
```

If I read the implementation correctly, it is implemented using HEAD not GET.

dohnto (changes requested):

grafana/resource_cloud_stack.go
```diff
@@ -268,3 +287,28 @@ func FlattenStack(d *schema.ResourceData, stack gapi.Stack) {
...
+		}
+		defer resp.Body.Close()
+		if resp.StatusCode != 200 {
+			return resource.RetryableError(errors.New("stack is not ready yet"))
+		}
+
+		return nil
+	})
+	if err != nil {
+		return diag.FromErr(fmt.Errorf("error waiting for stack to be ready: %s", err))
```
dohnto: I haven't used the diag errors, but isn't `%w` always more useful?
```suggestion
		return diag.FromErr(fmt.Errorf("error waiting for stack to be ready: %w", err))
```

grafana/resource_cloud_stack.go
```diff
@@ -268,3 +287,28 @@ func FlattenStack(d *schema.ResourceData, stack gapi.Stack) {
...
 }
+
+// waitForStackReadiness retries until the stack is ready, verified by querying the Grafana URL
+func waitForStackReadiness(ctx context.Context, d *schema.ResourceData) diag.Diagnostics {
+	if wait := d.Get("wait_for_readiness").(bool); !wait {
+		return nil
+	}
+
+	err := resource.RetryContext(ctx, 2*time.Minute, func() *resource.RetryError {
+		resp, err := http.Head(d.Get("url").(string))
```
dohnto: I think you should utilize the given parent `ctx` context to fulfills possible timeouts/cancellations during the HTTP request.

julienduchesne:
Thanks for the review @dohnto. All fixed 👍

dohnto (approved):

PR #424, opened 2022-03-08
Add resource to install cloud plugin

matthewnolf:
Adds a new resource `grafana_cloud_plugin_installation` that can be used to install plugins to Grafana Cloud instances.

Example usage:
```hcl
    resource "grafana_cloud_plugin_installation" "test-installation" {
        stack_slug = "mystack"
        slug       = "aws-datasource-provisioner-app"
        version    = "1.7.0"
    }
```

Depends on https://github.com/grafana/grafana-api-golang-client/pull/68 being merged


grafana/resource_cloud_plugin.go
```diff
@@ -0,0 +1,89 @@
...
+			"slug": {
+				Description: "Slug of the plugin to be installed.",
+				Type:        schema.TypeString,
+				Optional:    false,
+			},
+			"version": {
+				Description: "Version of the plugin to be installed.",
+				Type:        schema.TypeString,
+				Optional:    false,
+			},
```
julienduchesne: These three attributes should be `ForceNew: true`
Also, instead of `Optional: false`, you can just set `Required: true`. Not sure if it's equivalent but it's the pattern most providers use

examples/resources/grafana_cloud_plugin/resource.tf
```diff
@@ -0,0 +1,5 @@
+resource "grafana_cloud_plugin_installation" "test" {
```
julienduchesne: The name of the folder has to match the resource name, otherwise it doesn't show up in the docs

templates/index.md.tmpl
```diff
@@ -88,3 +88,6 @@ Finally, go back to your stack page, and go to "Details" for Loki to get
 ```http
 https://grafana.com/orgs/<org-slug>/hosted-logs/<logs-instance-id>
 ```
+
+### Installing a plugin
```
julienduchesne: For a resource like this, the resource page is sufficient. No need for it to be on the index. My own resources were like "glue" for different provider instances, so that's why I put them there

grafana/resource_cloud_plugin.go
```diff
@@ -0,0 +1,94 @@
...
+
+	d.SetId(strconv.Itoa(installation.ID))
+
+	return nil
+}
+
+func resourceCloudPluginInstallationRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*client).gcloudapi
+
+	stackSlug := d.Get("stack_slug").(string)
```
julienduchesne: A read should always start from `d.Id()` (not attributes set by the user). That's how the importer will work. So your ID would could be `<stack_slug>/<plugin_slug>`. It's easier for the user to figure out and it allows you to get the entire information afterwards

grafana/resource_cloud_plugin_test.go
```diff
@@ -0,0 +1,82 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGrafanaCloudPluginInstallation(slug, pluginSlug, pluginVersion),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCloudPluginInstallationCheckExists("grafana_cloud_plugin_installation.test-installation", slug, pluginSlug),
+					resource.TestCheckResourceAttrSet("grafana_cloud_plugin_installation.test-installation", "id"),
+					resource.TestCheckResourceAttr("grafana_cloud_plugin_installation.test-installation", "stack_slug", slug),
+					resource.TestCheckResourceAttr("grafana_cloud_plugin_installation.test-installation", "slug", "aws-datasource-provisioner-app"),
+					resource.TestCheckResourceAttr("grafana_cloud_plugin_installation.test-installation", "version", "1.7.0")),
+			},
```
julienduchesne, line 29: You have an importer so an import test would be nice. Example: https://github.com/grafana/terraform-provider-grafana/blob/master/grafana/resource_folder_test.go#L41-L45

examples/resources/grafana_cloud_plugin/resource.tf
julienduchesne: You should also place an `import.sh` script beside this file, and it will also show up in the docs

grafana/resource_cloud_plugin.go
```diff
@@ -0,0 +1,94 @@
...
+	pluginSlug := d.Get("slug").(string)
+
+	installation, err := client.GetCloudPluginInstallation(stackSlug, pluginSlug)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	_ = d.Set("slug", installation.PluginSlug)
+	_ = d.Set("version", installation.Version)
```
julienduchesne: On a 404, the Id should be nulled. Like this: https://github.com/grafana/terraform-provider-grafana/pull/430. This ensures that if someone removes a plugin outside of Terraform, it will be put back. Also, this one is tricky so a test is nice here (like in the PR)

julienduchesne (approved):

PR #428, opened 2022-03-09
Add `grafana_synthetic_monitoring_installation` resource

julienduchesne:
This is used to bridge the gap between creating a cloud stack and using SM
See the README for an example!


docs/index.md
```diff
@@ -91,6 +91,72 @@ resource "grafana_folder" "my_folder" {
...
+provider "grafana" {
+  alias         = "cloud"
+  cloud_api_key = "<my-api-key>"
+}
+
+resource "grafana_cloud_stack" "sm_stack" {
+  provider = grafana.cloud
+
+  name        = "<StackName>"
+  slug        = "<stackslug>"
```
inkel: Nit:
```suggestion
  slug        = "<StackSlug>"
```

inkel (approved):
Minor comment, but not blocking.

docs/index.md
julienduchesne: I was trying to show the format of each element but I guess that that wasn't clear 😆 . I'll fix it

PR #431, opened 2022-03-12
Cloud Stack: Use same schema for data and resource

julienduchesne:
Also:
- Some refactoring to the functions to do this (use the same functions for SM probe)
- Rename resource functions to be public (to be standard across all datasources and resources)

### DIFF
--- a/grafana/schema.go
+++ b/grafana/schema.go
@@ -38,7 +38,7 @@ func cloneResourceSchemaForDatasource(r *schema.Resource, updates map[string]*sc
 			delete(clone, k)
 		} else {
 			clone[k] = v
-		}		
+		}
 	}
 	return clone
 }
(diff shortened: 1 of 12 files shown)


grafana/schema.go
```diff
@@ -20,62 +20,24 @@ func schemaDiffFloat32(k, old string, nw string, d *schema.ResourceData) bool {
...
-func fixDatasourceSchemaFlags(schema map[string]*schema.Schema, required bool, keys ...string) {
-	for _, v := range keys {
-		schema[v].Computed = false
-		schema[v].Optional = !required
-		schema[v].Required = required
+	for k, v := range updates {
+		clone[k] = v
+		if v == nil {
+			delete(clone, k)
+		}
```
inkel: Small nit: how about this instead? I was confused at first when I read this.
```suggestion
		if v == nil {
			delete(clone, k)
		} else {
			clone[k] = v
		}		
```

inkel (approved):
Just a small comment.

PR #437, opened 2022-03-16
Allow removing already deleted users from an organization

### LINKED ISSUE #427: grafana_organization should silently ignore user deletion when the user does not exist any more in Grafana.
Hello,

In the same terraform code, we have two loops that creates all referenced users in Grafana and add them to different organizations.

- The code calls a module that creates a single user in a for_each loop of users. 
- The module fetches their email from LDAP and return them as an output, which is used as an input to the organization module
- The organization module creates one organization, and admins/editors/viewers are built from the users module.

This allow us to create a dependency and Terraform waits upon all users creation before trying to update the organization.

Upon removal, when there's multiple changes, the user is deleted first, and the organization update fails because it can't find its ID any more. The message is not appropriate either:

`Error: error adding user user@domain. User does not exist in Grafana` (when it is a removal).

We could switch to use `create_users` option of the organization resource, but the users are not deleted from Grafana with this option. After taking a glimpse look at the code, a solution could be:

[resource_organization.go#L304-L317](https://github.com/grafana/terraform-provider-grafana/blob/5c25203575953382707f610b06638e6ff903d43e/grafana/resource_organization.go#L304-L317)
```
	for _, change := range changes {
		id, ok := gUserMap[change.User.Email]
		if !ok && !create && change["Type"] != "Remove" {
			return nil, fmt.Errorf("error adding user %s. User does not exist in Grafana", change.User.Email)
		}
		if !ok && create {
			id, err = createUser(meta, change.User.Email)
			if err != nil {
				return nil, err
			}
		}
		if ok || change["Type"] != "Remove" {
			change.User.ID = id
			output = append(output, change)
		}
	}
```
I haven't code in GO for a while, I'm sorry if this proposal is broken, but the logic is here.

If you have any other solution, let me know.

julienduchesne:
It is currently impossible to remove a user from a org if it has been deleted. This PR fixes that

Fixes https://github.com/grafana/terraform-provider-grafana/issues/427

### DIFF
--- a/grafana/resource_organization.go
+++ b/grafana/resource_organization.go
@@ -289,6 +289,10 @@ func addIdsToChanges(d *schema.ResourceData, meta interface{}, changes []UserCha
 	create := d.Get("create_users").(bool)
 	for _, change := range changes {
 		id, ok := gUserMap[change.User.Email]
+		if !ok && change.Type == Remove {
+			log.Printf("[WARN] can't remove user %s from organization %s because it no longer exists in grafana", change.User.Email, d.Id())
+			continue
+		}
 		if !ok && !create {
 			return nil, fmt.Errorf("error adding user %s. User does not exist in Grafana", change.User.Email)
 		}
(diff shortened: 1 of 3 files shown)


inkel (approved):
Thank you!

PR #449, opened 2022-03-31
Cloud Stack: Fix read on existing slacks

### LINKED ISSUE #448: Error: json: cannot unmarshal number 0.00011808529102399268 into Go struct field for grafana_cloud_stack resource
### Terraform Version
* Terraform: 1.1.7
* Terraform Grafana Provider: 1.21.0
* Grafana: Grafana Cloud

### Affected Resource(s)
* grafana_cloud_stack

```hcl
provider "grafana" {
  alias         = "cloud"
  cloud_api_key = data.sops_file.secrets.data["grafana_cloud_org_key"]
}

resource "grafana_cloud_stack" "my_stack" {
  provider    = grafana.cloud
  name        = "teststack"
  slug        = "teststack"
  region_slug = "eu"
}

```

### Error

The resource is properly created, and i can run a plan or apply after, but after some time/a couple of retries I get this error and plan will break. It starts happening randomly (or so I think ATM), but once it happens, will continue to happen. I have only been able to bypass this by removing the resources from my state file, manually destroying the stack on grafana cloud, and applying again

```
╷
│ Error: json: cannot unmarshal number 0.00011808529102399268 into Go struct field Stack.hlInstanceCurrentUsage of type int
│ 
│   with grafana_cloud_stack.my_stack,
│   on providers.tf line 62, in resource "grafana_cloud_stack" "my_stack":
│   62: resource "grafana_cloud_stack" "my_stack" {
│ 

```

julienduchesne:
Fixes https://github.com/grafana/terraform-provider-grafana/issues/448
Bumping the client to https://github.com/grafana/grafana-api-golang-client/pull/76 does it

Also, I found an import issue on the cloud stacks while testing this. I added an import test

### DIFF
--- a/grafana/resource_cloud_stack.go
+++ b/grafana/resource_cloud_stack.go
@@ -80,6 +80,8 @@ Changing region will destroy the existing stack and create a new one in the desi
 				Optional:    true,
 				Default:     true,
 				Description: "Whether to wait for readiness of the stack after creating it. The check is a HEAD request to the stack URL (Grafana instance).",
+				// // Always suppress diff, we don't care about this field, it is only used at creation time.
+				// DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool { return true },
 			},
 			"org_id": {
 				Type:        schema.TypeInt,
--- a/grafana/resource_cloud_stack.go
+++ b/grafana/resource_cloud_stack.go
@@ -82,7 +82,7 @@ Changing region will destroy the existing stack and create a new one in the desi
 				Description: "Whether to wait for readiness of the stack after creating it. The check is a HEAD request to the stack URL (Grafana instance).",
 				// Suppress the diff if the new value is "false" because this attribute is only used at creation-time
 				// If the diff is suppress for a "true" value, the attribute cannot be read at all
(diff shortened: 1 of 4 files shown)


grafana/resource_cloud_stack.go
```diff
@@ -80,6 +80,9 @@ Changing region will destroy the existing stack and create a new one in the desi
 				Optional:    true,
 				Default:     true,
 				Description: "Whether to wait for readiness of the stack after creating it. The check is a HEAD request to the stack URL (Grafana instance).",
+				// Suppress the diff if the new value is "false" because this attribute is only used at creation-time
+				// If the diff is suppress for a "true" value, the attribute cannot be read at all
+				DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool { return newValue == "false" },
```
inkel: IIRC it's possible not to name the parameters that we won't use, as:
```suggestion
				DiffSuppressFunc: func(_, _, newValue string, _ *schema.ResourceData) bool { return newValue == "false" },
```

Is this a nice feature or not? I can't say 😬

inkel (approved):
:shipit:

PR #456, opened 2022-04-05
Add Grafana Oncall support

Konstantinov-Innokentii:


grafana/provider.go
```diff
@@ -140,6 +141,21 @@ func Provider(version string) func() *schema.Provider {
...
+					Type:        schema.TypeString,
+					Optional:    true,
+					Sensitive:   true,
+					DefaultFunc: schema.EnvDefaultFunc("GRAFANA_AMIXR_ACCESS_TOKEN", nil),
+					Description: "An Amixr access token. May alternatively be set via the `GRAFANA_AMIXR_ACCESS_TOKEN` environment variable.",
+				},
+				"amixr_url": {
+					Type:         schema.TypeString,
+					Optional:     true,
+					DefaultFunc:  schema.EnvDefaultFunc("GRAFANA_AMIXR_URL", "https://synthetic-monitoring-api.grafana.net"),
```
julienduchesne: oops, probably not `https://synthetic-monitoring-api.grafana.net`

grafana/provider.go
```diff
@@ -140,6 +141,21 @@ func Provider(version string) func() *schema.Provider {
...
+					Description:  "Synthetic monitoring backend address. May alternatively be set via the `GRAFANA_SM_URL` environment variable.",
```
julienduchesne: Description is copied as well

grafana/provider.go
```diff
@@ -210,6 +242,8 @@ type client struct {
 	smURL string
 
 	mlapi *mlapi.Client
+
+	aapi *aapi.Client
```
julienduchesne: I'd call that `amixrAPI` even if that's a bit more verbose. I can read the others, but aapi is a bit harder

grafana/provider.go
```diff
@@ -344,6 +383,16 @@ func createSMClient(d *schema.ResourceData) (string, *smapi.Client) {
 	return smURL, smapi.NewClient(smURL, smToken, nil)
 }
 
+func createAClient(d *schema.ResourceData) (*aapi.Client, error) {
```
julienduchesne: Same here
```suggestion
func createAmixrClient(d *schema.ResourceData) (*aapi.Client, error) {
```

grafana/resource_amixr_integration.go
```diff
@@ -0,0 +1,393 @@
...
+					},
+				},
+				MaxItems:    1,
+				Description: "Jinja2 templates for Alert payload.",
+			},
+		},
+	}
+}
+
+func resourceIntegrationCreate(d *schema.ResourceData, m interface{}) error {
```
julienduchesne: I think these should be prefixed as well, to be more explicit when autocompleting

```suggestion
func resourceAmixrIntegrationCreate(d *schema.ResourceData, m interface{}) error {
```

grafana/provider.go
```diff
@@ -164,9 +180,8 @@ func Provider(version string) func() *schema.Provider {
 				"grafana_user":                    ResourceUser(),
 
 				// Cloud
-				"grafana_cloud_api_key":             ResourceCloudAPIKey(),
-				"grafana_cloud_plugin_installation": ResourceCloudPluginInstallation(),
-				"grafana_cloud_stack":               ResourceCloudStack(),
+				"grafana_cloud_api_key": ResourceCloudAPIKey(),
+				"grafana_cloud_stack":   ResourceCloudStack(),
```
julienduchesne: Oops, removed a cloud resource there

julienduchesne (commented):
This is coming along nicely!

grafana/provider.go
inkel: I would also suggest `ocapi` or `irmapi` instead.

grafana/adapters.go
```diff
@@ -39,3 +51,14 @@ func int32SliceToIntList(list []int32) []interface{} {
 func int32SliceToSet(src []int32) *schema.Set {
 	return schema.NewSet(schema.HashInt, int32SliceToIntList(src))
 }
+
+func listOfSetsToStringSlice(listSet []interface{}) [][]string {
+	var ret [][]string
+	if listSet == nil {
+		return ret
+	}
```
inkel: There's no need for this `if`: calling `range` on a `nil` slice is the same as calling it over an empty slice. What I suggest doing instead is allocating `ret` with the required size:
```suggestion
	ret := make([][]string, 0, len(listSet))
```

grafana/data_source_amixr_action.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"fmt"
+	amixrAPI "github.com/grafana/amixr-api-go-client"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func DataSourceAmixrAction() *schema.Resource {
+	return &schema.Resource{
+		Description: `
+* [HTTP API](https://grafana.com/docs/grafana-cloud/oncall/oncall-api-reference/outgoing_webhooks/)
```
inkel: I believe this is going to the wrong page, maybe this is the right URL?
```suggestion
* [HTTP API](https://grafana.com/docs/grafana-cloud/oncall/oncall-api-reference/)
```

grafana/provider.go
```diff
@@ -344,6 +385,16 @@ func createSMClient(d *schema.ResourceData) (string, *smapi.Client) {
...
 
+func createAmixrClient(d *schema.ResourceData) (*amixrAPI.Client, error) {
+	aToken := d.Get("amixr_access_token").(string)
+	base_url := d.Get("amixr_url").(string)
+	aclient, err := amixrAPI.New(base_url, aToken)
+	if err != nil {
+		return nil, err
+	}
+	return aclient, nil
+}
```
julienduchesne:
```suggestion
	return amixrAPI.New(base_url, aToken)
}
```
You can return immediately, the client will be ignored if there's an err

grafana/resource_amixr_route.go
```diff
@@ -0,0 +1,169 @@
...
+		},
+	}
+}
+
+func resourceAmixrRouteCreate(d *schema.ResourceData, m interface{}) error {
+	client := m.(*client).amixrAPI
+	if client == nil {
+		err := errors.New("amixr api client is not configured")
+		return err
+	}
```
julienduchesne: I have an idea to remove all of these blocks, but it's just sugar. We can wait until we migrate all of this to `...Context` functions

julienduchesne (approved):
I don't see any blockers to merge this. I think we can ship this and go incrementally from there!

grafana/data_source_amixr_escalation_chain.go
```diff
@@ -0,0 +1,57 @@
...
+			},
+		},
+	}
+}
+
+func dataSourceEscalationChainRead(d *schema.ResourceData, m interface{}) error {
+	client := m.(*client).amixrAPI
+	if client == nil {
+		err := errors.New("amixr api client is not configured")
+		return err
```
inkel: No need to create a new variable here:

```suggestion
		return errors.New("amixr api client is not configured")
```

grafana/data_source_amixr_escalation_chain.go
```diff
@@ -0,0 +1,57 @@
...
+	options := &amixrAPI.ListEscalationChainOptions{}
+	nameData := d.Get("name").(string)
+
+	options.Name = nameData
+
+	escalationChainsResponse, _, err := client.EscalationChains.ListEscalationChains(options)
+
+	if err != nil {
+		return err
+	}
```
inkel: Nit: error checking should be immediately after the action that might trigger an error is.
```suggestion
	escalationChainsResponse, _, err := client.EscalationChains.ListEscalationChains(options)
	if err != nil {
		return err
	}
```

grafana/data_source_amixr_schedule.go
```diff
@@ -0,0 +1,62 @@
...
+			},
+		},
+	}
+}
+
+func dataSourceAmixrScheduleRead(d *schema.ResourceData, m interface{}) error {
+	client := m.(*client).amixrAPI
+	if client == nil {
+		err := errors.New("amixr api client is not configured")
+		return err
```
inkel:
```suggestion
		return errors.New("amixr api client is not configured")
```

grafana/data_source_amixr_schedule.go
```diff
@@ -0,0 +1,62 @@
...
+	options := &amixrAPI.ListScheduleOptions{}
+	nameData := d.Get("name").(string)
+
+	options.Name = nameData
+
+	schedulesResponse, _, err := client.Schedules.ListSchedules(options)
+
+	if err != nil {
+		return err
+	}
```
inkel:
```suggestion
	schedulesResponse, _, err := client.Schedules.ListSchedules(options)
	if err != nil {
		return err
	}
```

grafana/data_source_amixr_schedule.go
inkel: I've found the same in other files, can you please check those as well?

inkel (changes requested):
I've just left a couple suggestions regarding styling. Also, can you format the example files? I've just checked and the following aren't properly formatted.

```
$ terraform fmt -check -recursive examples
examples/provider/provider-amixr.tf
examples/resources/grafana_amixr_escalation/resource.tf
examples/resources/grafana_amixr_on_call_shift/grafana_amixr_on_call_shift.tf
examples/resources/grafana_amixr_route/resource.tf
examples/resources/grafana_amixr_schedule/resource.tf
```

grafana/data_source_amixr_action.go
Konstantinov-Innokentii: The link is correct, action and outgoing webhook are the same

grafana/provider_test.go
```diff
@@ -85,7 +85,7 @@ func TestProviderConfigure(t *testing.T) {
 		{
 			name:        "no config",
 			env:         map[string]string{},
-			expectedErr: "\"auth\": one of `auth,cloud_api_key,sm_access_token` must be specified",
+			expectedErr: "\"auth\": one of `auth,cloud_api_key,oncall_access_token,sm_access_token` must\nbe specified",
```
inkel, line 88: I'm curious: where does this `\n` between `must` and `be` come from? Checking the [SDK code](https://github.com/hashicorp/terraform-plugin-sdk/blob/4681738a561387fb0b3aaa69aeb42231383634a0/helper/schema/schema.go#L1874) that raises this error doesn't seem to have a newline, maybe it's the `resource.UnitTest`? 🤔

inkel (approved):
Excellent work, @Konstantinov-Innokentii!!! Thanks for your patience with all the change requests ❤️ 

Let's wait for @julienduchesne for a final approval.

julienduchesne (approved):

PR #464, opened 2022-04-22
Fix folder resource read for Grafana 8.5

### LINKED ISSUE #463: Unable to create folders with grafana 8.5
### Terraform Version
* Terraform: 1.1.7
* Terraform Grafana Provider: 1.21.1
* Grafana: 8.5.0

### Affected Resource(s)

* grafana_dashboard
* grafana_folder

```
error: Provider produced inconsistent result after apply
│ 
│ When applying changes to grafana_folder.envoy, provider "provider[\"registry.terraform.io/grafana/grafana\"]" produced an unexpected new value: Root resource was present, but now absent.
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵```

```hcl
resource "grafana_folder" "envoy" {
  title = "Envoy"
}
```

julienduchesne:
Temporarily replace the read logic with a more hackish logic that works for every version (uses the "list" and "get by uid" endpoints which are more commonly used, and therefore not broken)

Fixes https://github.com/grafana/terraform-provider-grafana/issues/463

### DIFF
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -1,4 +1,4 @@
-local grafanaVersions = ['8.4.4', '8.3.7', '8.2.7', '8.1.8', '7.5.15'];
+local grafanaVersions = ['8.5.0', '8.4.7', '8.3.7', '8.2.7', '7.5.15'];
 local images = {
   go: 'golang:1.18',
   lint: 'golangci/golangci-lint',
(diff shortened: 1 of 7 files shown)


inkel (approved):
Ship it!

saikatharryc:
@julienduchesne @inkel  
There is an issue with versioning when used with multi-project architecture,
with one project when creating the folder for first time works.
another project, I'm trying to update the same, (No changes - I expect it to pass through),
it fails with `{“message”:“the folder has been changed by someone else”,“status”:“version-mismatch”}`

PR #474, opened 2022-05-05
Basic Roles - Remove builtin role assignments

Jguer:
Fixes https://github.com/grafana/grafana-enterprise/issues/3059

Built in role assignments will be deprecated soon, when basic roles go on by default, this should be merged


vtorosyan:
For the context, this resource is available only for enterprise users, we should be able to give in advance warning before making the change. 

I'd also recommend to wait until Grafana 9.0 is out and release this afterwards.

@Jguer with the remaining `grafana_role` resource, how will we support basic roles modifications? Or would you suggest not to support it at all and add it later?

Jguer:
@vtorosyan needs to be tested but as it is, it should be possible to do the permission updates using the `grafana_role` resource (might just need a few tweaks to ensure only update is performed on those)

vtorosyan (approved):
I have tested the changes and can confirm that `grafana_role` resource supports the functionality which was previously provided by the `grafana_builtin_assignments` resource. Note that since this is a breaking change, we will have the announcement properly with the release notes, including instructions of how to use the `grafana_role`.

Note that this PR https://github.com/grafana/terraform-provider-grafana/pull/495 needs to be merged too.

vtorosyan:
For future reference, to manipulate permissions of a [basic role](https://grafana.com/docs/grafana/next/enterprise/access-control/about-rbac/#basic-roles), the following steps are needed:

1. Define desired basic role in the Terraform configuration
2. Import the resource as [described here](https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/role#import)
3. Make sure `terraform plan` does not show any change after import
4. Modify permissions of the `grafana_role` resource
5. Apply the changes

PR #479, opened 2022-05-10
Fix field attribute

alekhrycaiko:
Corrects an error in documentation for the `grafana_oncall_schedule` resource. Imaginably others will get caught with this, so I went ahead and added a correction.


inkel (approved):
Good catch, thanks!

inkel (approved):

PR #482, opened 2022-05-12
feat: add data_source_organization

wywywywy:


grafana/data_source_organization.go
```diff
@@ -0,0 +1,42 @@
...
+	client := meta.(*client).gapi
+	name := d.Get("name").(string)
+	org, err := client.OrgByName(name)
+
+	if err != nil && err.Error() == "404 Not Found" {
+		return nil
+	}
+	if err != nil {
+		return diag.FromErr(err)
+	}
```
inkel: Couple of things here.

First, how about the following?
```suggestion
	if err != nil {
		if err.Error() == "404 Not Found" {
			return nil
		}

		return diag.FromErr(err)
	}
```

Second, but probably more important: I'm not 100% sure that in the case of a non existing organization the value of `err.Error()` is that string. A quick test shows me that the value is actually `status: 404, body: {"message":"Organization not found"}`.

This clearly raises an issue in the client library, as we should be returning sentinel errors in this case, but as it currently is, this code won't return `nil` when the organization isn't found.

inkel (changes requested):
Thank you for your contribution! 🎉 

I've left some comments on your changes, please let us know once they get addressed for another round of reviews.

inkel:
Can you also please rebase this branch with the latest `master` branch? There were some fixes for our CI pipeline that are needed for this before merging.

inkel:
🤔 it seems there are lots of conflicts after your rebase, are you on Windows by any chance? From what I can tell your rebase changed all the newlines and that's why so many files are reported as conflicting.

wywywywy:
> 🤔 it seems there are lots of conflicts after your rebase, are you on Windows by any chance? From what I can tell your rebase changed all the newlines and that's why so many files are reported as conflicting.

Yes sorry about that. I was testing rebasing from VS Code and made a proper mess... I should have stuck with the CLI 😞 

I'll try and fix it, but if I can't get it done then I'll close this PR and create a new one.

wywywywy:
Ok. Changes made as suggested.

As a side note, there is a problem with the docker test. I'll investigate more and submit another PR later.

wywywywy:
Tests are failing but it seems to be a transient problem. Can you rerun it if possible please.

inkel:
@wywywywy yup, re-ran the tests and now are passing. Thanks for your contribution!

inkel (approved):

PR #483, opened 2022-05-12
Upgrade indirect hc-install package

inkel:
This package is responsible for installing Terraform to perform acceptance tests. Since the announcement of the [HashiCorp Releases API](https://www.hashicorp.com/blog/announcing-the-hashicorp-releases-api) 3 days ago our CI is broken.

After further inspection, I've found that this package was indeed the culprit, and that we need to upgrade the package to 0.3.2 after https://github.com/hashicorp/hc-install/pull/57 got merged.


mildwonkey (approved):
LGTM!

inkel:
I've also pinned the linter to version 1.45 instead of the latest, we can always review that in another PR.

Duologic (approved):

PR #486, opened 2022-05-12
Fix broken docker-compose test

wywywywy:
There are 2 problems stopping the docker-compose tests from working.

1. The `GRAFANA_VERSION` environment variable is now mandatory to run tests since this PR but it was not passed into the docker container - #365

2. The test certificates are all out of date (expired 2022-04-25)


inkel (commented):
Thanks for looking into this. One question though: when do these new certificates expire?

wywywywy:
```
% openssl x509 -in testdata/grafana.crt -noout -enddate
notAfter=May 12 22:53:06 2023 GMT
```

inkel (approved):

PR #487, opened 2022-05-13
Generate certificates on the fly

inkel:
In #486 it became apparent that keeping our certificates up to date when testing using Docker (`make testacc-docker` and `make testacc-docker-tls`) can become a tedious task, albeit one that we don't have to run too often. Not only this, but also had a dependency on an external tool to create the dependencies.

This PR removes away the certificates from the repository, and the external tool dependency, and replaces the mechanism with a Go file that creates certificates with an expiration of 1 hour since the time of creation. This might be low, but the `testacc-docker` and `testacc-docker-tls` targets were also modified to generate the certificates before running the tests using Docker, thus the 1 hour expiration should be more than enough (unless of course the test suite runs for longer than 1 hour, in which case we will have a bigger issue than the certificates being expired :wink:)


inkel:
The code is pretty crude at the moment, that's why it's a draft. I'll try to refactor it into simpler and more readable code.

julienduchesne:
Code LGTM but running it gives me the following error: `x509: certificate is not valid for any names, but wanted to match mtls-proxy`

inkel:
🤔 funny. I'll look into it on Monday. I remember seeing something about mtls-proxy in an article.

inkel:
@julienduchesne can I ask you to add the following in line 75, right after `crt.IPAddresses`:

```go
crt.DNSNames = []string{"mtls-proxy"}
```

And then try again? 🙏🏽

julienduchesne:
> ```go
> crt.DNSNames = []string{"mtls-proxy"}
> ```

I tried it, with or without the modification and now I get:
```
Error: Post "https://admin:***@mtls-proxy:3001/api/folders": x509: certificate has expired or is not yet valid: current time 2022-05-17T13:18:00Z is after 2022-05-14T11:29:35Z
```

I even cleared all files in `testdata` and I still get the same thing when they are regenerated

inkel:
This is really odd, because the generator uses the current time and set the expiration to an hour later 🤔 

I could understand a difference in seconds or even minutes, but not days 🤔 🤔

julienduchesne:
@inkel, I pushed a commit. It all works now. Can you test and confirm?
Command: `TESTARGS="-run TestAccTeam_basic" make testacc-docker-tls`

julienduchesne (approved):

inkel:
```
=== RUN   TestAccTeam_basic
--- PASS: TestAccTeam_basic (8.93s)
PASS
ok      github.com/grafana/terraform-provider-grafana/grafana   8.945s
```

We're good to go! Thanks.

PR #501, opened 2022-05-24
Fix grafana oncall shifts docs

Konstantinov-Innokentii:
This PR fixes docs for oncall shift resource (Example of usage missing there). Also I cleaned makefile, because pushed there commands which I am using for my local setup.


julienduchesne (approved):
👍

PR #507, opened 2022-05-26
Add new datasource: Cloud IPs

### LINKED ISSUE #314: Make Grafana Source IPs allowlist accessible via provider
The data for https://grafana.com/docs/grafana-cloud/reference/allow-list/ should be available using the provider.

As an example: https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/ip_ranges, https://registry.terraform.io/providers/hashicorp/tfe/latest/docs/data-sources/ip_ranges

julienduchesne:
Simply the list of IPs to allow, from https://grafana.com/docs/grafana-cloud/reference/allow-list/
Closes https://github.com/grafana/terraform-provider-grafana/issues/314


inkel (approved):
I like it, but hit me thinking: is a premature optimization to fetch those URLs in parallel?

julienduchesne:
> I like it, but hit me thinking: is a premature optimization to fetch those URLs in parallel?

I'd say so 😄 . 5 light GET calls like this probably takes a few hundred ms and it's a datasource that will only ever be used once (not in a loop)

Issue #380, opened 2022-02-07
Allow to map roles directly to users, teams and service accounts


vtorosyan:
## Context

Grafana fine-grained access control allows now to assign roles directly to users and teams. We would need to update relevant terraform resources to support this functionality. 

## Implementation proposal

I can think of about three different approaches:

### First option - extend `grafana_user` and `grafana_team` resources

Extend [`grafana_user`](https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/role) and [`grafana_team`](https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/team) resources with an ability to assign roles in-place
 
```
resource "grafana_user" "staff" {
  email    = "staff.name@example.com"
  name     = "Staff Name"
  login    = "staff"
  password = "my-password"
  is_admin = false

 roles {
    uid  = "uid"
    global = false
  }

 roles {
    uid  = "seconduid"
    global = true
  }
}

resource "grafana_team" "test-team" {
  name  = "Test Team"
  email = "teamemail@example.com"
  members = [
    "viewer-01@example.com"
  ]

  roles {
    uid  = "uid"
  }
}
```
**Pros:**
- The roles are assigned in-place, this gives very clear idea of what a team or a user can do. 

**Cons:**
- Fine-grained access control is an enterprise feature. The `grafana_role` and `grafana_user` resources are supporting open source functionality. Mixing the two comes with a cost - maintainability, clarity and unmanageable expectations from users perspective are few of the things I can think of.
- When you have terraform modules and resources spread in different places, I can imagine it might be a hard job for terraform users to pass down necessary variables to couple roles to users/teams.

### Second option - create a new resource

We can create a new resource, which will be designated for adding roles to users and teams. 
```
resource "grafana_role_assignments" "assignments" {
 role_uid = "uid"

 users {
    user_id  = 1
    global = true
  }

 teams {
    team_id  = 1
  }
}
```
Pros:
- This follows the [grafana_folder_permission](https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/folder_permission) design somehow. So might be a pattern we want to keep consistent.
- You clearly see who has the given role assigned and control it in one place.
- You can have your teams and users defined in one place, roles in another and the grafana_role_assignments in another.
- This will allow to assign fixed roles to users and teams without a need of having the fixed role definition first in the terraform. Note that fixed roles are created by grafana backend and if users need to terraform it, they would need first to define it and import from the backend.

[...]

PR #522, opened 2022-06-15
Pause running the failing test for datasource_permission resource

vtorosyan:
### Context

With Grafana 9.0, the datasource permissions are by default enabled. On top of that, datasource permissions are implemented using the new access control system - RBAC. A side effect of this is that the endpoint which returns datasource permissions now returns additional permissions, which are computed on the fly by the RBAC system and are not actually persisted in the database.

### Problem

When the new endpoint returns additional permissions, the sample terraform configuration was not aware about it and was failing. Note that this is actually an issue not only with the test assertion, but an issue when working with the resource. So we are going to look at resolution for this specific case in addition to this (in other words, the actual fix is out of scope of this PR).

What is important to note is that while terraform plan will show that it is trying to remove the non-existing permissions in the config, actually it will have no impact, as the permissions are computed with the request, nothing is persisted.

### The fix

- Pause the test from running temporary, until the RBAC resources are ready to control access for DS permissions

### DIFF
--- a/docs/resources/data_source_permission.md
+++ b/docs/resources/data_source_permission.md
@@ -42,9 +42,6 @@ resource "grafana_data_source_permission" "fooPermissions" {
     user_id    = 3 // 3 is the admin user in cloud. It can't be queried
     permission = "Query"
   }
-  permissions {
-    permission = "Query"
-  }
 }
 ```
 
--- a/docs/resources/data_source_permission.md
+++ b/docs/resources/data_source_permission.md
@@ -42,7 +42,6 @@ resource "grafana_data_source_permission" "fooPermissions" {
     user_id    = 3 // 3 is the admin user in cloud. It can't be queried
     permission = "Query"
   }
-  // As the data_source_permission is using legacy implementation, this temporary fixes the issue when Grafana returns more on-the-fly computed permissions than provisioned.
   permissions {
     permission = "Query"
   }
--- a/docs/resources/data_source_permission.md
+++ b/docs/resources/data_source_permission.md
@@ -42,6 +42,10 @@ resource "grafana_data_source_permission" "fooPermissions" {
     user_id    = 3 // 3 is the admin user in cloud. It can't be queried
     permission = "Query"
   }
+  // As the data_source_permission is using legacy implementation, this temporary fixes the issue when Grafana returns more on-the-fly computed permissions than provisioned.
+  permissions {
+    permission = "Query"
+  }
 }
 ```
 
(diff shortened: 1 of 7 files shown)


vtorosyan:
Note there is another broken docs link with the new release, which is this commit https://github.com/grafana/terraform-provider-grafana/pull/522/commits/32df5ca7d9c6b6b3a284246222112422a1433984

grafana/resource_datasource_permission_test.go
```diff
@@ -64,15 +64,9 @@ func testAccDatasourcePermissionsCheckExists(rn string, datasourceID *int64) res
...
 		if err != nil {
-			return fmt.Errorf("Error getting datasource permissions %d: %s", *datasourceID, err)
-		}
-		if response.Enabled {
-			return fmt.Errorf("Datasource permissions %d still enabled", *datasourceID)
-		}
-		if len(response.Permissions) > 0 {
-			return fmt.Errorf("Permissions were not empty when expected")
+			return fmt.Errorf("error getting datasource permissions %d: %s", *datasourceID, err)
 		}
```
julienduchesne: This doesn't check anything anymore. We want to check that the user-provisioned permissions were removed, right?

docs/resources/data_source_permission.md
```diff
@@ -42,6 +42,9 @@ resource "grafana_data_source_permission" "fooPermissions" {
     user_id    = 3 // 3 is the admin user in cloud. It can't be queried
     permission = "Query"
   }
+  permissions {
+    permission = "Query"
+  }
```
julienduchesne: So, does this mean anyone is allowed to query? If so, and it's added by default, this resource is useless because the only available permission is `Query`

vtorosyan: Your assumption is correct, with the Grafana 9.0, when creating a new datasource in Grafana Enterprise (where datasource permissions are available), everyone by default is able to query as the datasource permissions are enabled now by default. 

In addition to that change, there is now Edit permission available next to query (which allows users to edit the datasource). That said, I agree that this datasource needs to be deprecated eventually and we should use the new role-based access control system for managing permissions for datasources. For now, I think this resource is still relevant as you can use it to assign permissions to teams and specific users, in case you take out Query permissions from Viewers, for example. For the way forward I would suggest the following:

- Short term: Add the Edit permission to the allowed list of permissions
- Mid term: Come up with a holistic approach for managing access for DS permissions using RBAC and once there is consensus, mark this resource as a deprecated and start working on the new alternative.

I will create relevant issues to track this.

vtorosyan: @julienduchesne after some research and talking to the team, we have decided to pursue the proper approach and introduce missing resources for RBAC ([this issue specifically](https://github.com/grafana/terraform-provider-grafana/issues/380)). Meanwhile, as you suggested, let's skip running this test.

julienduchesne (approved):

PR #523, opened 2022-06-15
Bump grafana-api-golang-client from 0.2.6 to 0.6.0 and fix incompatibilities

alexweav:
Hello! We're wanting to extend the terraform provider with some new types for unified alerting.

We have an open PR to add the first of these types to the golang client: https://github.com/grafana/grafana-api-golang-client/pull/86

But, the terraform provider still depends on 0.2.6 of the client, which is incompatible with the latest version that the PR is open against. This PR bumps the provider to the latest version of the golang client.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.16
 
 require (
 	github.com/Masterminds/semver/v3 v3.1.1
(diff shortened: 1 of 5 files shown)


alexweav:
Never mind, I just didn't sync >.>

PR #497, opened 2022-05-20
Add a deprecation message for grafana_builtin_role_assignment

vtorosyan:
We are going to remove the resource once the Grafana 9.0 is released. This PR adds a small deprecation message as an in advance warning.

PR #529, opened 2022-06-20
Remove grafana_builtin_role_assignment resource

vtorosyan:
This change is essentially a revert of [this revert](https://github.com/grafana/terraform-provider-grafana/pull/496). 

With Grafana 9.0 we have removed the built-in role assignments entirely and this resource has been [marked as deprecated ](https://github.com/grafana/terraform-provider-grafana/pull/497) with previous versions and we can safely remove the resource itself. 

Note that in addition we mentioned the breaking change in the What's new for Grafana 9.0.


Jguer (approved):

vtorosyan:
Closing this as we have decided to keep the resource as deprecated until few next versions, when we feel comfortable that the resource is not use anymore.

PR #532, opened 2022-06-22
Alerting: Add support for Unified Alerting message templates

alexweav:
Adds support for provisioning the message template resource in unified alerting.

Rel https://github.com/grafana/grafana/issues/36153


JohnnyQQQQ (approved):
LGTM, what are the semantics when you don't add a `define` tag? The server would create it with `define` tag and we save the new template value in the state?

alexweav:
Yes. The semantics are essentially identical to using the UI. There is no difference between using UI and using Terraform, I was very careful to preserve this.

PR #536, opened 2022-06-28
Add `Edit` permission to resource_datasource_permission

ishanjainn:
From the Grafana Instance, when adding permissions for a data source, it can be either `Query` or `Edit`. But when adding permissions to data source, due to the check, It can be only `Query`. 
Therefore adding `Edit` as an option.

### DIFF
--- a/docs/resources/data_source_permission.md
+++ b/docs/resources/data_source_permission.md
@@ -62,7 +62,7 @@ resource "grafana_data_source_permission" "fooPermissions" {
 
 Required:
 
-- `permission` (String) Permission to associate with item. Must be `Query`.
+- `permission` (String) Permission to associate with item. Must be `Query` or `Edit`.
 
 Optional:
 
--- a/docs/resources/data_source_permission.md
+++ b/docs/resources/data_source_permission.md
@@ -62,7 +62,7 @@ resource "grafana_data_source_permission" "fooPermissions" {
 
 Required:
 
-- `permission` (String) Permission to associate with item. Must be `Query` or `Edit`.
+- `permission` (String) Permission to associate with item. Must be one of `Query` or `Edit`.
 
 Optional:
 
--- a/grafana/resource_datasource_permission.go
+++ b/grafana/resource_datasource_permission.go
@@ -54,7 +54,7 @@ func ResourceDatasourcePermission() *schema.Resource {
 						"permission": {
 							Type:         schema.TypeString,
 							Required:     true,
-							ValidateFunc: validation.StringInSlice([]string{"Query","Edit"}, false),
+							ValidateFunc: validation.StringInSlice([]string{"Query", "Edit"}, false),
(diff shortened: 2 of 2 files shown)


inkel (approved):

inkel:
Note that you need to generate the docs for the CI to pass.

ishanjainn:
The CI pipeline is failing at `cloud api tests` with the below error. Could you help with this? @inkel 
```
testing_new.go:84: Error running post-test destroy, there may be dangling resources: exit status 1
        
        Error: status: 500, body: {
          "code": "Internal",
          "message": "Error deleting hosted instance"
        }
        
--- FAIL: TestAccGrafanaAuthKeyFromCloud (56.81s)
```

julienduchesne:
> The CI pipeline is failing at `cloud api tests` with the below error. Could you help with this? @inkel
> 
> ```
> testing_new.go:84: Error running post-test destroy, there may be dangling resources: exit status 1
>         
>         Error: status: 500, body: {
>           "code": "Internal",
>           "message": "Error deleting hosted instance"
>         }
>         
> --- FAIL: TestAccGrafanaAuthKeyFromCloud (56.81s)
> ```

It's a bit flaky. I should add retries around that

julienduchesne:
This resource is deprecated btw. I think it doesn't work with Grafana 9.0+. @vtorosyan can tell you more

ishanjainn:
Got it, For Grafana 9.0+, is there some other resource which can help control data source permissions? @vtorosyan @julienduchesne

vtorosyan:
As @julienduchesne mentioned, this resource at the moment is semi-broken. Grafana 9.0 comes with RBAC which is now fully supporting the datasource permissions also, and the underlying implementation is done in a way to support backward compatibility, which introduced an issue with this resource, particularly, every time you run terraform plan, regardless if there are changes or not, the resource would try to add additional "Query" permission (this is because Grafana API returns this additional one to support RBAC).

That said, we think that the right way to control permissions for datasources programmatically in the future should be by using [grafana_role](https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/role), and for this to work properly, we also need [this issue](https://github.com/grafana/terraform-provider-grafana/issues/380) completed, which allows to map roles to users, teams and service accounts. We are aiming to deliver this as soon as possible and chose not to invest in datasource_permissions resource anymore.

Note that there is another challenge with this implementation - the Edit type is _actually_ not support by Grafana API, it comes from RBAC, so even if we add the type here, it won't work unfortunately. 

@ishanjainn would you perhaps be able to share a bit more details about the specific need and perhaps we can find a way to support it , until we figure out the future of this?

vtorosyan:
I have opened a different issue to track the bug and progress https://github.com/grafana/terraform-provider-grafana/issues/542, hopefully this provides some additional details.

ishanjainn:
Awesome, Thanks for the context @vtorosyan. Ill close the PR since it isn't applicable now with Grafana 9+.

PR #537, opened 2022-06-28
Add retries on the gcloud client

julienduchesne:
We often get 500s on the tests (probably timing errors because we create and delete stacks pretty quick)
This should help


inkel (approved):
LGTM and I think it's ready to be merged, but how about if we add a default number of retries?

julienduchesne:
> LGTM and I think it's ready to be merged, but how about if we add a default number of retries?

It’s already 3 retries (same as the Grafana client)

PR #545, opened 2022-07-05
NOOP: Various small code formatting and shell script cleanup tasks

mdb:
Hello! Thanks for your work on on `terraform-provider-grafana`. This PR addresses a few small code cleanup opportunities I noticed in the source code, but should not impact functionality:

* removes trailing whitespace from `GNUMakefile`
* `.drone/wait-for-instance.sh` now uses consistent 2 space indentation
* `.drone/wait-for-instance.sh` now conforms to [shellcheck](https://github.com/koalaman/shellcheck) recommendations (specifically, it 1) uses double quotes surrounding variables to prevent globbing and wordsplitting if/when in put contains spaces, line feeds, or glob characters, and 2) no longer features an unnecessary `echo`

Thank you!

### DIFF
--- a/.drone/wait-for-instance.sh
+++ b/.drone/wait-for-instance.sh
@@ -3,7 +3,7 @@
 set -euxo pipefail
 
 getStatus() {
-  "$(curl -I -L -s -o /dev/null -w "%{http_code}" "${1}")"
+  curl -I -L -s -o /dev/null -w "%{http_code}" "${1}"
 }
 
 status=$(getStatus "${1}")
--- a/.drone/wait-for-instance.sh
+++ b/.drone/wait-for-instance.sh
@@ -3,17 +3,17 @@
 set -euxo pipefail
 
 getStatus() {
-  echo "$(curl -I -L -s -o /dev/null -w "%{http_code}" $1)"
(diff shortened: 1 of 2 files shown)


.drone/wait-for-instance.sh
```diff
@@ -3,17 +3,17 @@
 set -euxo pipefail
 
 getStatus() {
-    echo "$(curl -I -L -s -o /dev/null -w "%{http_code}" $1)"
```
mdb, line 6: I don't believe the `echo` nor the subshell is necessary.

.drone/wait-for-instance.sh
```diff
@@ -3,17 +3,17 @@
...
+  curl -I -L -s -o /dev/null -w "%{http_code}" "${1}"
```
mdb, line 6: Quoting the input prevents the script from breaking if ever the input contains spaces, line feeds, or glob characters.

.drone/wait-for-instance.sh
julienduchesne: hah, indeed, that's terrible 😆

julienduchesne (approved):

julienduchesne:
Thanks

Issue #480, opened 2022-05-12
Add grafana_organization datasource


adkinsjd:
It would be helpful to have a datasource which mirrors the grafana_organization resource.

wywywywy:
I've created a new PR that you might be interested in https://github.com/grafana/terraform-provider-grafana/pull/482

inkel:
PR #482 has been merged, thanks @wywywywy for creating it! This will be included in the next version of the provider, I'll let you know once that's done.

adkinsjd:
This is great! Ideally I would also be able to get a list of users (similar to the admins, editors, and viewers list in the resource). It looks like it wouldn't be too difficult to add; maybe I'll have time to give it a shot later in the week.

mdb:
> Ideally I would also be able to get a list of users (similar to the admins, editors, and viewers list in the resource).

@adkinsjd Does PR #551 address this issue in your mind?

adkinsjd:
Yes it does!

PR #551, opened 2022-07-06
`grafana_organization` data source returns member user data 

mdb:
This seeks to address issue #480.

Specifically, its implementation is inspired by the following comment from @adkinsjd:

https://github.com/grafana/terraform-provider-grafana/issues/480#issuecomment-1127156583

The generated docs included in the PR -- as well as my PR annotation comments -- provide additional implementation notes and reviewer considerations.

Thanks!


docs/data-sources/organization_users.md
```diff
@@ -0,0 +1,43 @@
...
+  create_users = true
+  admin_user   = "admin"
+  admins = [
+    "admin@example.com"
+  ]
+}
+
+data "grafana_organization_users" "from_name" {
+  organization_name = grafana_organization.test_org.name
+}
```
mdb: In theory, the already-existing `data.grafana_organization` could be enhanced to report user data, but I opted to make a separate, dedicated `grafana_organization_users` data source given that `GET /api/org/<id>` (used by `data.grafana_organization`) and `GET /api/org/<id>/users` are separate Grafana API endpoints. Do others agree?

grafana/data_source_organization_users.go
```diff
@@ -0,0 +1,62 @@
...
+* [HTTP API](https://grafana.com/docs/grafana/latest/http_api/org/)
+`,
+		ReadContext: dataSourceOrganizationUsersRead,
+		Schema: map[string]*schema.Schema{
+			"organization_name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The name of the Organization.",
+			},
+			"emails": {
```
mdb: I considered calling this `user_emails` and/or `user_email_addresses`, though `emails` seems the most terse, clear, and consistent with the `email` attribute used elsewhere by the API and provider. Is that reasonable?

docs/data-sources/organization_users.md
julienduchesne: I would make it a single resource. And mirror the attributes from the existing organization resource (admins, viewers, editors)

grafana/data_source_organization_users.go
```diff
@@ -0,0 +1,62 @@
...
+				Type:     schema.TypeSet,
+				Computed: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				Description: "A list of the Organization member users' email addresses.",
+			},
```
mdb: I considered enhancing the data source with a `users` property:

```golang
"users": {
  Type:     schema.TypeSet,
  Computed: true,
  Elem: &schema.Resource{
      "email": {
        Type:     schema.TypeString,
        Computed: true,
      },
      "login": {
        Type:     schema.TypeString,
        Computed: true,
      },
      "role": {
        Type:     schema.TypeString,
        Computed: true,
      },
      "id": {
        Type:     schema.TypeInt,
        Computed: true,
      },
    },
  },
```

Would others like to see that as part of this PR now? If so, should that replace or coincide with the existing `emails`?

grafana/data_source_organization_users_test.go
```diff
@@ -0,0 +1,61 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccDatasourceOrganizationUsersConfig(orgName, email),
+				Check:  resource.ComposeTestCheckFunc(checks...),
+			},
+		},
+	})
+}
+
+func testAccDatasourceOrganizationUsersConfig(orgName, email string) string {
```
mdb: I opted to use this configuration in tests -- rather than the `examples/data-sources/grafana_organization_users/data-source.tf` configuration -- so we can dynamically declare the `foo@example.com` user email address from within the tests, rather than needing to hard-code the expected email addresses across both the `examples/data-sources/grafana_organization_users/data-source.tf` and `grafana/data_source_organization_users_test.go` files.

Do others agree with this decision, or would you prefer I abandon `testAccDatasourceOrganizationUsersConfig` in favor of using a `examples/data-sources/grafana_organization_users/data-source.tf` configuration?

docs/data-sources/organization_users.md
mdb: > I would make it a single resource. And mirror the attributes from the existing organization resource (admins, viewers, editors)

@julienduchesne Thanks for the feedback! I'm amenable to that! But, before I get started, does it change your opinion to consider that...

1. That isn't how the [Grafana API itself](https://grafana.com/docs/grafana/latest/developers/http_api/org/#get-users-in-organization) models this data (`GET /api/org/<id>/users` is a separate endpoint & resource; the endpoint's associated JSON resource itself is simply a `[]OrgUser`)
2. At a glance, it seems other, existing `terraform-provider-grafana` data sources may err on the side of multiple, distinct, simple individual resources (for example, `data.grafana_team` returns no team membership/user attributes; instead, the `data.grafana_user` is used to fetch these attributes)

...or do you still favor a single `data.grafana_organization` data source with `admins`, `editors`, and `viewers` attributes?

julienduchesne: Better to mirror the existing organization resource IMO

mdb: @julienduchesne Sounds good and thanks for the feedback! Stay tuned...

mdb:
@julienduchesne I believe this is ready for re-review; thanks!

grafana/data_source_organization.go
```diff
@@ -38,6 +63,28 @@ func dataSourceOrganizationRead(ctx context.Context, d *schema.ResourceData, met
...
+	}
+
+	for _, user := range orgUsers {
+		role := fmt.Sprintf("%ss", strings.ToLower(user.Role))
+		userCollections[role] = append(userCollections[role], user.Email)
+	}
+
+	for roleCollection, emails := range userCollections {
+		if err := d.Set(roleCollection, emails); err != nil {
+			return diag.FromErr(fmt.Errorf("error setting %s: %v", roleCollection, err))
```
julienduchesne:
```suggestion
			return diag.Errorf("error setting %s: %v", roleCollection, err)
```

grafana/data_source_organization.go
```diff
@@ -38,6 +63,28 @@ func dataSourceOrganizationRead(ctx context.Context, d *schema.ResourceData, met
...
+		"editors": []string{},
+		"viewers": []string{},
+	}
+
+	for _, user := range orgUsers {
+		role := fmt.Sprintf("%ss", strings.ToLower(user.Role))
+		userCollections[role] = append(userCollections[role], user.Email)
+	}
+
+	for roleCollection, emails := range userCollections {
```
julienduchesne:
```suggestion
	for role, emails := range userCollections {
```

It's just a role, not a collection of it

julienduchesne (approved):
LGTM! Just some nits

grafana/data_source_organization.go
mdb: @julienduchesne I believe `roleCollection` refers to the key name of the `map[string][]string` and is plural (i.e. `admins`, `editors`, `viewers`). Nonethless, I'm happy to change this variable name if you prefer.

julienduchesne:
Thanks @mdb !

Issue #11, opened 2018-01-03
[Feature Request] Support Annotations


mdb:
In current implementation, it appears `terraform-provider-grafana` does not support Grafana [annotations](http://docs.grafana.org/reference/annotations/). It would be great to be able to annotate dashboards from a `terraform apply`, thus empowering Grafana dashboard visualization of events -- such as deployments -- and the events' impact on key metrics.

Note that [apparentlymart/go-grafana-api PR #6](https://github.com/apparentlymart/go-grafana-api/pull/6) seeks to add annotation support to `apparentlymart/go-grafana-api`, such that annotation support could then be addressed here in `terraform-provider-grafana`.

Thanks!

julienduchesne:
There's a typo on this line: https://github.com/nytm/go-grafana-api/blob/master/annotation.go#L41
`"/api/annotation"` should be `"/api/annotations"`. This would need a fix on the client's side

julienduchesne:
Hi @trotttrotttrott, I tried to tackle this but there's a typo here: https://github.com/grafana/grafana-api-golang-client/blob/master/annotation.go#L41. `"/api/annotation"` should be `"/api/annotations"`. However, I saw that that client is deprecated. Is the solution to do a PR to it anyways or is there another client that could be used?

mdb:
@julienduchesne To add a bit more context there, I attempted to fix this `/api/annotation` typo in [PR 57](https://github.com/nytm/go-grafana-api/pull/57/files) of the original `github.com/nytm/go-grafana-api` client package, which I believe was the original home of `github.com/grafana/grafana-api-golang-client`, but that PR was never reviewed or merged.

So, that's just to say, I have the same question: should bug fixes be PR'd against `github.com/grafana/grafana-api-golang-client`, or should `terraform-provider-grafana` use an alternative API client?

inkel:
> However, I saw that that client is deprecated. Is the solution to do a PR to it anyways or is there another client that could be used?

@julienduchesne despite the client being deprecated I'd say please do fix it, we have been adding some small fixes to it nevertheless.

mdb:
> despite the client being deprecated I'd say please do fix it, we have been adding some small fixes to it nevertheless.

@inkel I've created another PR fixing the `GET /api/annotations` URL path here: https://github.com/grafana/grafana-api-golang-client/pull/36

Thanks!

PR #558, opened 2022-07-13
add initial, basic grafana_annotation resource

mdb:
This seeks to address issue #11

Note that, because Terraform manages the CRUD lifecycle of resources declared in `*.tf` configuration, the `grafana_annotation` resource creates an annotation on initial `terraform apply`, then continuously updates that _same_ annotation (or deletes that same annotation if the `grafana_annotation` declaration is removed from `*.tf` configuration). This may be unexpected for those users who expect `grafana_annotation` to create fresh annotations with each `terraform apply` invocation and to leave in place those annotations created by preceding `terraform apply` invocations.


grafana/resource_annotation_test.go
```diff
@@ -0,0 +1,109 @@
...
+					testAccAnnotationCheckExists("grafana_annotation.test", &annotation),
+					resource.TestCheckResourceAttr("grafana_annotation.test", "text", testAccAnnotationUpdatedText),
+				),
+			},
+			{
+				// Importing matches the state of the previous step.
+				ResourceName:      "grafana_annotation.test",
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
```
mdb, line 49: Are there additional test cases worth adding?

grafana/resource_annotation.go
```diff
@@ -0,0 +1,174 @@
...
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Default:     false,
+				Description: "The ID of the dashboard on which to create the annotation.",
+			},
+
+			"panel_id": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Default:     false,
```
julienduchesne: Should be required
```suggestion
				Required:    true,
```

grafana/resource_annotation.go
```diff
@@ -0,0 +1,174 @@
...
+		ReadContext:   ReadAnnotation,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"dashboard_id": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Default:     false,
```
julienduchesne: Should be required
```suggestion
				Required:    true,
```

grafana/resource_annotation.go
```diff
@@ -0,0 +1,174 @@
...
+				Description: "The ID of the dashboard panel on which to create the annotation.",
+			},
+
+			"text": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     false,
```
julienduchesne: Should be required
```suggestion
				Required:    true,
```

docs/resources/annotation.md
```diff
@@ -0,0 +1,36 @@
...
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Optional
+
+- `dashboard_id` (Number) The ID of the dashboard on which to create the annotation. Defaults to `false`.
+- `panel_id` (Number) The ID of the dashboard panel on which to create the annotation. Defaults to `false`.
+- `tags` (Set of String) The tags to associate with the annotation.
+- `text` (String) The text to associate with the annotation. Defaults to `false`.
```
julienduchesne: I think having the time and timeEnd attributes is needed for a MVP. If we're managing annotations with Terraform, they're probably historical annotations, not ones that apply to right now

grafana/resource_annotation_test.go
```diff
@@ -0,0 +1,109 @@
...
+		if err != nil {
+			return err
+		}
+
+		if len(annotations) < 1 {
+			return nil
+		}
+
+		for _, annotation := range annotations {
+			if annotation.Text == testAccAnnotationInitialText || annotation.Text == testAccAnnotationUpdatedText {
```
julienduchesne: Since only this resource will ever create annotations and tests do not run in parallel, we could just fail whenever there's an annotation, no matter the text. It'd be future-test proof (no need to keep the test up-to-date)

grafana/resource_annotation_test.go
julienduchesne: Should be enough

julienduchesne (commented):
Great work. It's almost ready, just missing some attributes 🙂

grafana/resource_annotation.go
mdb: @julienduchesne Are you certain `dashboard_id` should be required? According to the [Grafana API docs](https://grafana.com/docs/grafana/latest/developers/http_api/annotations/#create-annotation):

> The `dashboardId` and `panelId` fields are optional. If they are not specified then an organization annotation is created and can be queried in any dashboard that adds the Grafana annotations data source.

grafana/resource_annotation.go
mdb: @julienduchesne As mentioned above, the [Grafana API docs](https://grafana.com/docs/grafana/latest/developers/http_api/annotations/#create-annotation) lead me to believe `panel_id` should be optional.

julienduchesne: Indeed, you're right. Then the default value should just be empty, not `false`. A boolean doesn't make much sense in this context. Also, 
- it probably needs to cause a recreate to change the `panelId` or the `dashboardId`
- I think panelId is required when you set dashboardId and vice-versa. You can use `RequiredWith` for this
- It needs a test case, we need to test that having an annotation without panelId and setting one does recreate the resource correctly

mdb: Thanks for the feedback. Stay tuned; I can cook all that up...

mdb: > I think panelId is required when you set dashboardId and vice-versa.

...and I do think we need to verify this, as this has not been my understanding, historically.

mdb: > it probably needs to cause a recreate to change the `panelId` or the `dashboardId`

@julienduchesne Is this to say you'd like to see a `ForceNew: true` associated with the `dashboard_id` and `panel_id` `*schema.Schema`? If so, I've added this.

> I think panelId is required when you set dashboardId and vice-versa.

@julienduchesne I don't believe this is true; I don't believe `panelId` and `dashboardId` are co-dependent, nor are either required. I believe I've proven this via an acceptance test configuration that uses `dashboard_id` without `panel_id`, and an acceptance test configuration that uses `panel_id` without a `dashboard_id`. Do you agree?

> It needs a test case, we need to test that having an annotation without panelId and setting one does recreate the resource correctly

@julienduchesne I believe I've added test cases covering these scenarios. Are there other test cases you'd like to see?

grafana/resource_annotation.go
mdb: I don't believe `dashboard_id` should be required.

julienduchesne (approved):
LGTM! Thanks!

PR #562, opened 2022-07-18
Alerting: Add support for Unified Alerting mute timings.

alexweav:
Adds support for mute timings. In draft, while a couple loose ends are tied up with the golang client.


go.mod
```diff
@@ -77,3 +77,5 @@ require (
 	google.golang.org/grpc v1.47.0 // indirect
 	google.golang.org/protobuf v1.28.0 // indirect
 )
+
+replace github.com/grafana/grafana-api-golang-client v0.9.0 => "../grafana-api-golang-client"
```
alexweav: Temporary, until changes get propagated.

grafana/resource_playlist.go
```diff
@@ -74,27 +72,28 @@ func ResourcePlaylist() *schema.Resource {
 }
 
 func CreatePlaylist(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-	client := meta.(*client).gapi
+	/*client := meta.(*client).gapi
```
alexweav: Playlist changes are temporary due to an unrelated breaking change in the golang client.

grafana/resource_playlist_test.go
```diff
@@ -177,7 +177,7 @@ func testAccPlaylistConfigBasic(name string) string {
 resource "grafana_playlist" "test" {
 	name     = %[1]q
 	interval = "5m"
-	
+
```
alexweav, line 180: Simple whitespace inconsistency that my editor kept wanting to fix.

grafana/resource_alerting_mute_timing.go
```diff
@@ -0,0 +1,307 @@
...
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "The name of the mute timing.",
+			},
+
+			"intervals": {
+				Type:        schema.TypeList,
```
alexweav, line 43: A list rather than a set here is a fairly nuanced decision. We need to reply on diff-suppression deeper in this file, but turns out that diff-suppression is completely broken in Terraform if there is a set anywhere on the path to that field.

This happens because TF represents sets inside the state object using the set's hash as a key, like:
```
intervals: map[string]interface{} {
    "hash-of-elem-1": <elem-1-data>,
    "hash-of-elem-2": <elem-2-data>,
}
```

Since the hash is the key, a change on any field no matter how deeply nested will cause the hash to be different. Terraform will think that this is a deletion and creation of the member at every layer, not an update.

When using diff-suppression, the arguments to the diff suppression hook end up being completely useless, it's all either empty strings or numbers representing the set's cardinality (weird, right?) This core reason is also coincidentally why the output of `tf plan` is so strange on sets, it shows every element as being changed rather than the one that actually got changed. Interesting issue! We just switch to lists to work around it.

JohnnyQQQQ: Nit: should we add this as a comment above the field?

grafana/resource_alerting_mute_timing.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+	return mt
+}
+
+func packIntervals(nts []gapi.TimeInterval) []interface{} {
+	if nts == nil {
+		return nil
+	}
+
+	intervals := make([]interface{}, 0)
```
JohnnyQQQQ:
```suggestion
	intervals := make([]interface{}, len(nts))
```

grafana/resource_alerting_mute_timing.go
```diff
@@ -0,0 +1,307 @@
...
+	return map[string]string{
+		"start": time.StartMinute,
+		"end":   time.EndMinute,
+	}
+}
+
+func unpackTimeRange(raw interface{}) gapi.TimeRange {
+	vals := raw.(map[string]interface{})
+	return gapi.TimeRange{
+		StartMinute: vals["start"].(string),
```
JohnnyQQQQ, line 308: Do we have a guarantee that this will be a string? I assume yes, as the type validation should run prior as part of the runtime.

JohnnyQQQQ (approved):
LGTM, added a few comments/questions.

grafana/resource_alerting_mute_timing.go
alexweav: Doesn't quite work because we use `append`. The array ends up 2x as long as expected with the first half all empty structs. Instead I did `make([]interface{}, 0, len(nts))` which has the desired effect, size is 0 but len(nts) element preallocated.

grafana/resource_alerting_mute_timing.go
alexweav: Yes, it's the same type as whatever we packed it into the state as. So the `pack` function directly above this one sort of mirrors it, and packs the other way around from `string` to `interface{}`.

This is pretty common practice in the provider, you'll see lots of type casting among all the types when reading from the data object. It's all over the place in nearly every file.

Issue #563, opened 2022-07-20
Add GNUmake command for running golangci-lint in local development


mdb:
Currently, `terraform-provider-grafana`'s `GNUmakefile` offers no easy way to locally reproduce the `lint` step that occurs in CI (https://github.com/grafana/terraform-provider-grafana/blob/master/.drone/drone.yml#L10-L13).

As a contributor, I'd like to be able to invoke `make golangci-lint` locally, in advance of opening a PR and invoking CI.

Thanks!

PR #564, opened 2022-07-20
add golangci-lint Make command

mdb:
Previously, `terraform-provider-grafana`'s `GNUmakefile`
offered no easy way to locally reproduce the `lint` step
that occurs in CI (https://github.com/grafana/terraform-provider-grafana/blob/master/.drone/drone.yml#L10-L13).

This seeks to fix that by addressing issue #563.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -46,5 +46,14 @@ drone:
 	drone lint .drone/drone.yml
 	drone sign --save grafana/terraform-provider-grafana .drone/drone.yml
 
+golangci-lint:
+	docker run \
+		--rm \
+		--interactive \
+		--tty \
+		--volume "$(shell pwd):/src" \
+		--workdir "/src" \
+		golangci/golangci-lint:v1.45 golangci-lint run ./...
+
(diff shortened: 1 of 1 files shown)


GNUmakefile
```diff
@@ -46,5 +46,14 @@ drone:
 	drone lint .drone/drone.yml
 	drone sign --save grafana/terraform-provider-grafana .drone/drone.yml
 
+golangci-lint:
+	docker run \
+		--rm \
+		--interactive \
+		--tty \
```
inkel: Are these needed for this?

inkel (commented):
Thank you for the change, @mdb! It's without a doubt a good addition. I've just left a minor comment, but otherwise I'm inclined to approve it once I've an answer 🙏🏽

GNUmakefile
mdb: @inkel - Good call; I don't believe those are necessary and I've removed them. Thanks!

inkel (approved):
LGTM, thanks for the changes!

PR #571, opened 2022-07-28
Update `sm_url` documentation

mem:
Clarify that the value _should_ be the same everywhere and specify where
to find it.

Signed-off-by: Marcelo E. Magallon <marcelo.magallon@grafana.com>

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -98,6 +98,7 @@ resource "grafana_folder" "my_folder" {
 provider "grafana" {
   alias         = "cloud"
   cloud_api_key = "<my-api-key>"
+  sm_url        = "<Synthetic Monitoring API URL>"
 }
 
 resource "grafana_cloud_stack" "sm_stack" {
@@ -130,6 +131,7 @@ resource "grafana_synthetic_monitoring_installation" "sm_stack" {
 provider "grafana" {
   alias           = "sm"
   sm_access_token = grafana_synthetic_monitoring_installation.sm_stack.sm_access_token
+  sm_url          = "<Synthetic Monitoring API URL>"
 }
 
 data "grafana_synthetic_monitoring_probes" "main" {
@@ -156,6 +158,12 @@ resource "grafana_synthetic_monitoring_check" "ping" {
   }
 }
 ```
+
+Please note the `sm_url` value is optional, but it must correspond with the
+value specified as the `region_slug` in the `grafana_cloud_stack` resource.
+Also, if the `sm_url` value is specified for one provider instance, it's almost
+certain that the same value should be specified for all provider instances.
+Currently the provider code DOES NOT enforce this.
 
 ### Managing Grafana OnCall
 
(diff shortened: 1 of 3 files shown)


docs/index.md
```diff
@@ -157,6 +159,12 @@ resource "grafana_synthetic_monitoring_check" "ping" {
 }
 ```
 
+Please note the `sm_url` value is optional, but it must correspond with the
+value specified as the `region_slug` in the `grafana_cloud_stack` resource.
+Also, if the `sm_url` value is specified for one provider instance, it's almost
+certain that the same value should be specified for all provider instances.
```
inkel: I'm not entirely sure that I understand what you mean with this and why is that.

mem: Thanks for looking at this.

I wasn't sure how to word this, so it's not a surprise that it's not clear.

If you look at the docs [here](https://registry.terraform.io/providers/grafana/grafana/latest/docs#installing-synthetic-monitoring-on-a-new-grafana-cloud-stack) you'll see that we tell users to instantiate the provider twice:

```
// Step 1: Create a stack
provider "grafana" {
  alias         = "cloud"
  cloud_api_key = "<my-api-key>"
}

...

// Step 3: Interact with Synthetic Monitoring
provider "grafana" {
  alias           = "sm"
  sm_access_token = grafana_synthetic_monitoring_installation.sm_stack.sm_access_token
}
...
```

the reason for this is that the first provider is used to create the Grafana stack and the second provider is used to interact with the Synthetic Monitoring API.

While it might be obvious that the second provider needs the `sm_url` setting, the first one also needs it because it's the one used to create the `grafana_synthetic_monitoring_installation` resource.

It could well be that it's not necessary to use two instances of the provider, because as far as I can see, the access token will be set on the SM installation resource after it's created.

But this is what the example is telling users to do...

inkel: Ah, I see what you mean now. 🤔 let me think a bit more about it, I believe your updates are correct… but I'd rather have a clearer overall documentation in the first place.

mem: I think I see the issue...

The token is set on the installation resource after it's created. The problem is that you need to set it as an attribute in the _provider_, so you need one provider to obtain the access token and a second provider to act on the SM resources. So you need to set the URL in the first provider so that the installation works and then you need to set the URL in the second provider so that the resources are created in the correct location. If you don't, you should see errors because the access token in the first provider won't be valid for the second provider.

So that's what I meant: you have to make sure that you are talking to the same SM API server in all the providers that deal with SM resources (unless you know _exactly_ what you are doing, of course, which is the reason I didn't want to assert that you MUST set the same URL in all the providers).

mdb: @mem With that in mind, would a phrasing such as the following satisfy your concern?

```
Note that when a Terraform configuration contains multiple provider instances managing SM resources associated with the same Grafana stack, specifying an explicit `sm_url` set to the same value for each provider ensures all providers interact with the same SM API.
```

inkel: @mdb yes, that reads better, thank you!

julienduchesne (approved):

Issue #411, opened 2022-03-04
Automatically update the list of Grafana versions


julienduchesne:
As mentioned here: https://github.com/grafana/terraform-provider-grafana/pull/410#issuecomment-1059223204, the list of Grafana versions maintained here: https://github.com/grafana/terraform-provider-grafana/blob/master/.drone/drone.jsonnet#L1 could be automatically generated so we don't have to look at Grafana release notes to find new releases

mdb:
> could be automatically generated 

@julienduchesne What source of data do you imagine driving the automation? And what are the business rules driving the list? For example, do you imagine using the GitHub releases API to fetch the latest Grafana minor version GitHub release for each major version release? Or are you imagining something else?

julienduchesne:
Following exploration of this issue and possible solutions by @mdb in https://github.com/grafana/terraform-provider-grafana/pull/572, I have decided not to pursue this issue further

PR #572, opened 2022-07-29
dynamically determine .drone.yml Grafana versions

mdb:
This seeks to address issue #411 by dynamically templating
the `.drone/drone.yml` file with the targeted Grafana versions,
as provided by the GitHub releases API.

It's currently in a draft state towards the goal of soliciting some early
feedback and conversation.

Note that...

* this currently assumes installation of the `gh` CLI ~, `jq`, `uniq`, and `sort`~, but
  perhaps should instead invoke a `docker run` such that `docker` is the only required
  dependency
* this currently populates the list of targeted Grafana versions with all those versions
  returned by the `repos/grafana/grafana/releases?per_page=100` endpoint. Rather than fetch the last 100 releases, should we paginate through and collect all releases?
* this currently invokes a shell script from `make drone`, though there are likely more
  `make`-idiomatic ways of accomplishing this without a shell script. Would those be preferable?


mdb:
@julienduchesne - I'm curious to get your early feedback on this draft PR. I can imagine a different directions to take this, but it's probably worth hearing more about your original vision before I over-invest in a misinterpretation.

julienduchesne:
> @julienduchesne - I'm curious to get your early feedback on this draft PR. I can imagine a different directions to take this, but it's probably worth hearing more about your original vision before I over-invest in a misinterpretation.

That issue was still a bit nebulous to me. I put it down as an idea but didn't put much thought into it.

It's a bit harder than what you're doing:
1. The idea was to have PRs to auto-update in the same sense that dependabot does but that's not really possible because the Drone pipeline signature needs to be generated by a human when the pipelines are modified
2. The versions that we select is essentially: The last patch of every major since 7 (we want the provider to stay retrocompatible as much as possible). The last patch of the last x (5?) minor versions (ex: 9.0, 8.5, 8.4, 8.3, 8.2). We can't select too many versions because running too many tests at the same time is problematic in many ways

Now, I think 1 is just impossible, so I think your solution is as good as we're gonna get, so if you get the versions right, I'd be inclined to merge this and close the issue 😄

mdb:
> so I think your solution is as good as we're gonna get, so if you get the versions right, I'd be inclined to merge this and close the issue

@julienduchesne Thanks for the feedback! I believe I've cooked up a script that implements the logic you describe. Currently, it's a `bash` script that assumes the installation of `gh` and `jq`, though this could be containerized if folks would prefer it only assume the installation of `docker` (itself already assumed to be installed elsewhere in the `GNUMakefile`).

.drone/get-grafana-versions.sh
```diff
@@ -0,0 +1,49 @@
...
+#     build metadata suffixes.
+# 3. remove duplicates resulting from multiple major.minor.patch
+#     versions with unique '-X'-style prerelease build metadata
+# 4. split on line breaks and turn into an array
+# 5. map each element to a {"major": "x", "minor": "y", "patch": "z"}
+#     JSON object.
+# 6. subgroup each release into an array of its minor releases
+# 7. reverse the array to be in descending order
+# 8. map the result to an array of the latest 5 minor Grafana versions
+
```
mdb, line 18: I added this verbose explanation, as the command is somewhat complex and not-self-evident, plus I don't believe I could provide this context via in-line comments associated with each `jq` `|` without breaking the command.

.drone/get-grafana-versions.sh
```diff
@@ -0,0 +1,49 @@
...
+      'split("\n")[:-1]
+      | map({
+        major: (split(".")[0]),
+        minor: (split(".")[1]),
+        patch: (split(".")[2])
+      })
+      | group_by(.major, .minor)
+      | reverse
+      | map(.[-1] | join("."))[:5]
+    '
```
mdb, line 35: ☝️ Does someone else see a preferable way to do this?

mdb: ~Update: I'm not confident this is correct. I'm going to convert this PR to a draft~. I _believe_ this is now correct, but I'd certainly appreciate a review and sanity check :)

mdb:
> The idea was to have PRs to auto-update in the same sense that dependabot does but that's not really possible because the Drone pipeline signature needs to be generated by a human when the pipelines are modified

@julienduchesne Thanks for clarifying! I've removed the actual updates to `.drone/drone.yml` from this PR, such that this PR merely adds a `make drone` utility for authorized humans to update the `.drone/drone.yml` in an automated fashion. Is that a reasonably helpful baby step for now, or do you regard that as unnecessary overhead and complexity? (If the latter, perhaps both issue #411 and this PR should both be closed/abandoned?)

julienduchesne:
> > The idea was to have PRs to auto-update in the same sense that dependabot does but that's not really possible because the Drone pipeline signature needs to be generated by a human when the pipelines are modified
> 
> @julienduchesne Thanks for clarifying! I've removed the actual updates to `.drone/drone.yml` from this PR, such that this PR merely adds a `make drone` utility for authorized humans to update the `.drone/drone.yml` in an automated fashion. Is that a reasonably helpful baby step for now, or do you regard that as unnecessary overhead and complexity? (If the latter, perhaps both issue #411 and this PR should both be closed/abandoned?)

Yeah, I'm sorry that lead you to invest so much time in this. I can't really bring myself to merge this, the productivity gain is just not enough to counter the additional complexity, IMO.

I opened this issue pretty quick and hadn't investigated the details surrounding the task. Considering the limitations that Drone pipelines have to be generated by a human (for security reasons), and that we want things to be relatively static so that new failures aren't introduced randomly, I don't think we should pursue this any further and we should keep updating manually. I'll close the issue.

Really great work though!

PR #574, opened 2022-08-01
Longer Cloud Stack readiness timeout

julienduchesne:
Lots of tests have been failing recently. Bumping the readiness timeout from 2 minutes to 5 minutes should make sure the tests always pass. Adding some more info to the error messages also

### DIFF
--- a/grafana/resource_cloud_stack.go
+++ b/grafana/resource_cloud_stack.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	gapi "github.com/grafana/grafana-api-golang-client"
+	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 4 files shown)


grafana/resource_cloud_stack.go
```diff
@@ -349,7 +349,9 @@ func waitForStackReadiness(ctx context.Context, d *schema.ResourceData) diag.Dia
 		}
 		defer resp.Body.Close()
 		if resp.StatusCode != 200 {
-			return resource.RetryableError(errors.New("stack is not ready yet"))
+			buf := new(bytes.Buffer)
+			buf.ReadFrom(resp.Body)
```
inkel: This call can [potentially panic](https://pkg.go.dev/bytes#Buffer.ReadFrom), which would be less than ideal but it will be managed by `go test` (sort of) so there's that. I wonder it it would make sense to check the error for other than `io.EOF` 🤔

grafana/resource_cloud_stack.go
```diff
@@ -338,7 +338,7 @@ func waitForStackReadiness(ctx context.Context, d *schema.ResourceData) diag.Dia
 		return nil
 	}
 
-	err := resource.RetryContext(ctx, 2*time.Minute, func() *resource.RetryError {
+	err := resource.RetryContext(ctx, 5*time.Minute, func() *resource.RetryError {
```
inkel: Would it make sense to make this a configurable value via an environment variable?

inkel (approved):
Just left my usual questions, but this is ready to be merged!

grafana/resource_cloud_stack.go
julienduchesne: Made it configurable as an attribute

grafana/resource_cloud_stack.go
julienduchesne: Oops, I checked the error

PR #579, opened 2022-08-04
Alerting: Add support for unified alerting contact points and two receiver types

alexweav:
Adds baseline support for unified alerting contact points.

We support a large number of receivers, this PR consists only of the core infrastructure and first two receivers. This PR also implements an abstraction layer for receivers, making it easy to declare and extend receivers without touching the main logic.

We add two receivers in this PR, `email` and `discord`, in order to demonstrate how the abstraction works.

The rest of the receiver implementations have been separated into a following PR in order to keep review length reasonable. This PR does all of the heavy lifting for contact points, the remaining receivers are trivial to create.


alexweav:
Tests are failing on an unrelated 409 in from cloud on a resource not touched by this PR. Will give it some time and re-run, but this isn't a blocker for reviews.

grafana/resource_alerting_contact_point.go
```diff
@@ -0,0 +1,449 @@
...
+	pack(p gapi.ContactPoint) interface{}
+	unpack(raw interface{}, name string) gapi.ContactPoint
+}
+
+type notifierMeta struct {
+	typeStr string
+	desc    string
+}
+
+type emailNotifier struct{}
```
JohnnyQQQQ, line 261: We might want to move the notifier implementations to another file, as it will eventually become super long and packed.

JohnnyQQQQ (approved):
LGTM

PR #389, opened 2022-02-15
add data_source_folders

justinTM:

Issue #377, opened 2022-02-05
add data_source_folders


justinTM:
get all Grafana folders

julienduchesne:
Closed by https://github.com/grafana/terraform-provider-grafana/pull/583

PR #583, opened 2022-08-10
Add `grafana_folders` data source for fetching all folders

mdb:
This seeks to address issue #377 by adding a `grafana_folders` data source, which can be used to fetch all Grafana folders.

Note that this is an alternative implementation to that contained in PR #389, which I admittedly only noticed just now (I apologize) and which appears to have gone stale with git conflicts and an unsigned CLA. Would maintainers prefer I close this PR in favor of the PR #389 implementation?


examples/data-sources/grafana_folders/data-source.tf
```diff
@@ -0,0 +1,16 @@
+resource "grafana_folder" "test_a" {
+  title = "test-folder-a"
+  uid   = "test-ds-folder-uid-a"
+}
+
+resource "grafana_folder" "test_b" {
+  title = "test-folder-b"
+  uid   = "test-ds-folder-uid-b"
+}
```
mdb, line 9: Would it be preferable if the example used a single `grafana_folder.test` and a `for_each`, rather than declaring semi-redundant `grafana_folder` resources?

grafana/data_source_folders.go
```diff
@@ -0,0 +1,95 @@
...
+		return diag.FromErr(err)
+	}
+
+	uids := []string{}
+	for _, folder := range folders {
+		uids = append(uids, folder.UID)
+	}
+	md5 := md5.Sum([]byte(strings.Join(uids, "-")))
+
+	d.SetId(fmt.Sprintf("%x", md5))
```
mdb: I believe an ID is required. In this case, the ID is an MD5 hash of a `-`-delimited string of all the folder UIDs. Is that acceptable?

grafana/data_source_folders.go
```diff
@@ -0,0 +1,95 @@
...
+		return diag.FromErr(err)
+	}
+
+	uids := []string{}
+	for _, folder := range folders {
+		uids = append(uids, folder.UID)
+	}
+	sha := sha256.Sum256([]byte(strings.Join(uids, "-")))
+
+	d.SetId(fmt.Sprintf("%x", sha))
```
julienduchesne: The ID should be declarative because it needs to be unique. For a datasource like this, it's usually the combination of all entry parameters because you will typically not have the same datasource twice with the same parameters, it's useless

With no parameters, you can just set the ID as `grafana_folders`. No need to calculate a SHA

examples/data-sources/grafana_folders/data-source.tf
julienduchesne: All good like this

julienduchesne (commented):
Nice one. Just a small comment about the ID

mdb:
> Nice one. Just a small comment about the ID

Thanks @julienduchesne - I've implemented your suggestion, assuming I interpreted correctly :)

julienduchesne (approved):
LGTM! Thanks!

PR #587, opened 2022-08-15
Alerting: Add support for basic, root-level notification policies

alexweav:
Adds support for basic notification policies (ones that only implement simple grouping and not complex routing). Tree-based routing will be implemented in a follow-up PR, I split this to keep things tidy for reviewers.

Much of the logic in this PR is orthogonal to supporting the policy trees. This is done in such a way that there will not be a compatibility breakage between the two PRs.


grafana/resource_alerting_notification_policy_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"testing"
+
+	gapi "github.com/grafana/grafana-api-golang-client"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+func TestAccNotificationPolicy_basic(t *testing.T) {
+	CheckOSSTestsEnabled(t)
+	CheckOSSTestsSemver(t, ">=9.0.0")
```
alexweav: This depends on 9.1, will change once the container image is published.

Not changing it now so that the CI stays broken. That way nobody accidentally submits this too early.

JohnnyQQQQ (approved):
LGTM

PR #591, opened 2022-08-17
Alerting: Add support for `opsgenie`, `pagerduty`, `pushover`, and `sensugo` notifiers

alexweav:
Adds the next batch of strongly-typed notifiers to the provider.

Extended the notifier interface a little - the marshalling code can now propagate arbitrary errors, which appear nicely in the Terraform output.

### DIFF
--- a/grafana/resource_alerting_contact_point_notifiers.go
+++ b/grafana/resource_alerting_contact_point_notifiers.go
@@ -456,3 +456,125 @@ func (k kafkaNotifier) unpack(raw interface{}, name string) gapi.ContactPoint {
 		Settings:              settings,
 	}
 }
+
+type opsGenieNotifier struct{}
+
+var _ notifier = (*opsGenieNotifier)(nil)
+
+func (o opsGenieNotifier) meta() notifierMeta {
+	return notifierMeta{
+		field:   "opsgenie",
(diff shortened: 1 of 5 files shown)


grafana/resource_alerting_contact_point_notifiers.go
```diff
@@ -456,3 +457,524 @@ func (k kafkaNotifier) unpack(raw interface{}, name string) gapi.ContactPoint {
...
+		settings["message"] = v.(string)
+	}
+	return gapi.ContactPoint{
+		UID:                   uid,
+		Name:                  name,
+		Type:                  s.meta().typeStr,
+		DisableResolveMessage: disableResolve,
+		Settings:              settings,
+	}
+}
```
JohnnyQQQQ, line 980: I think we should start splitting this file moving forward as we do in Grafana.

JohnnyQQQQ (approved):
LGTM, one non blocking remark

grafana/resource_alerting_contact_point_notifiers.go
alexweav: I agree, but I did not want to deviate from the current package structure in this project. These all have quite a bit less code than the actual notifier implementations so I felt it was ok.

PR #594, opened 2022-08-18
Alerting: Add support for `slack`, `teams`, `telegram`, and `threema` notifiers

alexweav:
Adds the next batch of notifiers, plus a helper that makes serialization code simpler for string fields.


yuri-tceretian (approved):
LGTM

PR #595, opened 2022-08-19
Alerting: Add support for `victorops`, `webhook`, and `wecom` notifiers

alexweav:
Adds support for the last 3 notifiers.

### DIFF
--- a/grafana/resource_alerting_contact_point_notifiers.go
+++ b/grafana/resource_alerting_contact_point_notifiers.go
@@ -1297,3 +1297,56 @@ func (t threemaNotifier) unpack(raw interface{}, name string) gapi.ContactPoint 
 		Settings:              settings,
 	}
 }
+
+type victorOpsNotifier struct{}
+
+var _ notifier = (*victorOpsNotifier)(nil)
+
+func (v victorOpsNotifier) meta() notifierMeta {
+	return notifierMeta{
+		field:   "victorops",
+		typeStr: "victorops",
+		desc:    "A contact point that sends notifications to VictorOps (now known as Splunk OnCall).",
+	}
+}
+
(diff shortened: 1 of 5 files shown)


JohnnyQQQQ (approved):
LGTM

grafana/resource_alerting_contact_point_notifiers.go
```diff
@@ -1257,7 +1258,7 @@ func (t threemaNotifier) schema() *schema.Resource {
 	}
 	r.Schema["recipient_id"] = &schema.Schema{
 		Type:        schema.TypeString,
-		Optional:    true,
+		Required:    true,
```
inkel, line 1261: Is this change related to the PR? It doesn't look like. Can we extract it to a PR of its own?

inkel (commented):
Overall LGTM. I left a question, please let me know when addressed.

grafana/resource_alerting_contact_point_notifiers.go
alexweav: Sure, I can split it out.

Quick note - I know this change appears to be compat breaking, but that schema was submitted in a PR just a couple days prior. None of these types have been released or published to the provider registry yet. So, users will never see a compat difference, we should not need to document anything special for this change.

alexweav: Also, it's _somewhat_ related to this PR. I added ~15 of these notifiers recently - it was one very large change, so I split them up into batches to make reviews easier. This change was simply an adjustment to a prior batch. It's not an unrelated field/resource, just a touch-up to the batch I had just submitted.

Given this, and given that it's not compat breaking per above, I feel OK with keeping it in the same PR.

PR #599, opened 2022-08-22
Alerting: Add support for nested notification policies

alexweav:
https://github.com/grafana/terraform-provider-grafana/pull/587 added support for top-level notification policies only.

This PR extends this to fully support entire notification policy trees. Policies are a recursive self-referential resource, in addition to the top node. This does not change compatibility, as it's done through a new field `policy`.

Terraform does not support true recursive schemas. Instead, we only support policy trees up to a certain finite depth. We recursively build a finite schema based on the depth configured in the provider. Depth can be increased without breaking backward compatibility. For now, we stick with a depth of 2, knowing that this can easily be increased in the future.

### DIFF
--- a/grafana/resource_alerting_notification_policy.go
+++ b/grafana/resource_alerting_notification_policy.go
@@ -2,6 +2,7 @@ package grafana
 
 import (
 	"context"
+	"fmt"
 
 	gapi "github.com/grafana/grafana-api-golang-client"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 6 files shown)


grafana/resource_alerting_notification_policy.go
```diff
@@ -53,12 +54,112 @@ func ResourceNotificationPolicy() *schema.Resource {
...
+				Description: "Routing rules for specific label sets.",
+				Elem:        policySchema(supportedPolicyTreeDepth),
+			},
 		},
 	}
 }
 
+// The maximum depth of policy tree that the provider supports, as Terraform does not allow for infinitely recursive schemas.
+// This can be increased without breaking backwards compatibility.
+const supportedPolicyTreeDepth = 2
```
JohnnyQQQQ: I have a hunch (but not data) that 2 might be a bit low. I would start with something like 4.

JohnnyQQQQ (approved):
LGTM

PR #602, opened 2022-08-22
Alerting: Resource-level write locking for resources backed by UA config file

alexweav:
### Background
Some (not all) alerting resources are stored in json/yaml config file on the backend of the alerting system. The APIs for these resources use optimistic locking to perform write protection - the result is that all changes to alerting configs must be linearizable. Writing to these resources is safe from data loss, but as a tradeoff, requests may be rejected with a `409 Conflict`.

This may cause the `terraform apply` to intermittently fail on large plans, since TF naturally runs with a parallelism of 10, and these resources do not always have explicit dependencies between one another.

### Solution
This PR adds a mutex around write operations for affected resources.

### Alternatives Considered
- **Require the TF user to specify implicit dependencies** using `depends_on`. This is a really bad user experience.
- **Require applies with --parallelism=1** has the same behavior of this PR, but is much too far-reaching. We only need parallelism=1 for these four specific resources, not every single type of resource. This approach is too limiting.
- **Retries** are a good solution, but a mutex is more robust. Even with a high retry counter, there is always a chance an object may fail. With a mutex, there are fewer requests and fewer failures overall, the desired end result is usually achieved faster.
- **Rate-limiting** helps reduce the frequency of the problem (less contention) but is not sufficient to outright eliminate it. 
- **A semaphore** is already applied by TF with a value of 10. Even with parallelism of 2, we can intermittently see issues for very big plans.

### Reasoning
Read operations are atomic and not affected. This only applies to four specific alerting resources, namely **not rules.**

In practice, the vast majority of alerting resources are rules. Since rules are unaffected, the perf penalty here is kept relatively minimal.

We intend to improve this locking requirement at the alerting API layer in the future. When this is done, the mutex can be relaxed to a simple rate-limiter (or removed entirely) which will eliminate this penalty.

### DIFF
--- a/docs/resources/notification_policy.md
+++ b/docs/resources/notification_policy.md
@@ -29,10 +29,13 @@ resource "grafana_mute_timing" "a_mute_timing" {
     intervals {
         weekdays = ["monday"]
     }
+<<<<<<< HEAD
 
     depends_on = [
         grafana_contact_point.a_contact_point
     ]
+=======
+>>>>>>> 319ed2e (Regenerate docs)
 }
 
 
--- a/docs/resources/notification_policy.md
+++ b/docs/resources/notification_policy.md
@@ -29,13 +29,6 @@ resource "grafana_mute_timing" "a_mute_timing" {
     intervals {
         weekdays = ["monday"]
     }
-<<<<<<< HEAD
-
-    depends_on = [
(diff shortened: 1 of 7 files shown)


JohnnyQQQQ (approved):
LGTM

Issue #618, opened 2022-08-25
grafana_oncall_integration notification to slack is disabled when provisioned


alekhrycaiko:
### Terraform Version
* Terraform: 1.1.6
* Terraform Grafana Provider: 1.27
* Grafana: Cloud

### Affected Resource(s)
- grafana_oncall_integration
- grafana_oncall_route

### Expected Behavior
When building a new `grafana_oncall_integration` with a default block set to a Slack channel, the `post to channel` option should be enabled, as I've specified I want my integrations default route posting to Slack.

When building a new grafana on call route, and specifying a Slack channel, I expect the 'post to channel' option to be enabled. 

### Actual Behavior
New integration or route built with a default route to Slack, has `post to channel` disabled. A user needs to go in and manually enable messages to be sent to Slack in the integration to override this.

julienduchesne:
Implemented in https://github.com/grafana/terraform-provider-grafana/pull/621

PR #621, opened 2022-08-29
Oncall route messengers fields

Ferril:
Add support for telegram and ms teams settings for OnCall routes and default routes in OnCall integrations.
Add ability to disable notifications for every messenger in OnCall route.
Notifications will be enabled by default for existing messenger settings in routes.


julienduchesne:
This will fix https://github.com/grafana/terraform-provider-grafana/issues/618, right?

grafana/resource_oncall_integration.go
```diff
@@ -98,14 +98,59 @@ func ResourceOnCallIntegration() *schema.Resource {
 								Schema: map[string]*schema.Schema{
 									"channel_id": {
 										Type:        schema.TypeString,
-										Required:    true,
+										Optional:    true,
 										Description: "Slack channel id. Alerts will be directed to this channel in Slack.",
 									},
+									"enabled": {
+										Type:        schema.TypeBool,
+										Required:    true,
```
julienduchesne: Can we have it default to false. As it is, it's a breaking change because user configurations will now throw an error

Ferril: Makes sense. I set it to "true" by default because this is more expected behavior for users.

Ferril:
> This will fix #618, right?

Yes

julienduchesne:
LGTM! Just missing a `go generate`

julienduchesne (approved):

Issue #481, opened 2022-05-12
Add grafana_organization_preferences resource


adkinsjd:
This would allow the provisioning of a home dashboard, timezone and theme for a new organization.

mdb:
If I understand correctly, I believe https://github.com/grafana/grafana-api-golang-client/pull/102 is a pre-requisite to a `grafana_organaization_preferences` resource.

jesusheredia-consensys:
I am interested in that as well. 

Someone is working on this?

julienduchesne:
Implemented and released in v1.30.0

PR #622, opened 2022-08-29
add `organization_preferences` resource and data source

mdb:
This is ~a WIP PR to add address~ addresses issue #481 by introducing an `organization_preferences` data source and resource.


grafana/data_source_organization_preferences.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+func DatasourceOrganizationPreferences() *schema.Resource {
+	return &schema.Resource{
+		Description: `
+* [Official documentation](https://grafana.com/docs/grafana/latest/administration/manage-organizations/)
+* [HTTP API](https://grafana.com/docs/grafana/latest/developers/http_api/preferences/#get-current-org-prefs)
+`,
+		ReadContext: dataSourceOrganizationPreferencesRead,
+		Schema: map[string]*schema.Schema{
```
julienduchesne: You can reuse your resource's schema with this function: https://github.com/grafana/terraform-provider-grafana/blob/master/grafana/schema.go#L23-L45

grafana/data_source_organization_preferences.go
```diff
@@ -0,0 +1,82 @@
...
+	//
+	// TODO: is it problematic that every organization_preference data source will have the same ID?
+	//
+	// According to @julienduchesne (https://github.com/grafana/terraform-provider-grafana/pull/583/files/b261189cf70ae4c076d9319d83abda2a959e5112#r944357467) ...
+	// "The ID should be declarative because it needs to be unique.
+	// For a datasource like this, it's usually the combination of all entry parameters because you
+	// will typically not have the same datasource twice with the same parameters"
+	//
+	// However, in this instance, the data source does not accept any parameters; they are all computed.
+	// So, what would be a reasonable way to calculate its ID?
```
julienduchesne: In that case, if every instance of that datasource has the same value, the id can be static. Just `organization_preferences` or something

grafana/resource_organization_preferences.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	d.Set("theme", prefs.Theme)
+	d.Set("home_dashboard_id", prefs.HomeDashboardID)
+	d.Set("home_dashboard_uid", prefs.HomeDashboardUID)
+	d.Set("timezone", prefs.Timezone)
+	d.Set("week_start", prefs.WeekStart)
+	d.Set("locale", prefs.Locale)
+
+	d.SetId(generateOrgPrefsIDSha())
```
julienduchesne: Same as the datasource, the ID can be static, this resource is a singleton, we only expect to only ever be a single one

grafana/resource_organization_preferences.go
```diff
@@ -0,0 +1,151 @@
...
+		return diag.FromErr(err)
+	}
+
+	return diag.Diagnostics{}
+}
+
+func DeleteOrganizationPreferences(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*client).gapi
+	orgID, _ := strconv.ParseInt(d.Id(), 10, 64)
+	if err := client.DeleteOrg(orgID); err != nil {
```
julienduchesne: I don't think deleting an org preferences resource should delete the org. It should just be a no-op IMO

grafana/resource_organization_preferences.go
```diff
@@ -0,0 +1,151 @@
...
+	if d.HasChange("name") {
+		name := d.Get("name").(string)
+		err := client.UpdateOrg(orgID, name)
+		if err != nil {
+			return diag.FromErr(err)
+		}
+	}
+	if err := UpdateUsers(d, meta); err != nil {
+		return diag.FromErr(err)
+	}
```
julienduchesne: I would've expected this function to be the exact same as the create one. This seems like it's the update function for the organization resource

julienduchesne (commented):
This would need some tests. Other than that and my comments, this looks nice!

grafana/resource_organization_preferences.go
mdb: 👍 I suspect this was early-stage copy/pasta. I've updated `DeleteOrganizationPreferences` to use [gapi.UpdateAllOrgPreferences](https://github.com/grafana/grafana-api-golang-client/blob/master/org_preferences.go#L37-L38) to clear out the org preferences via an empty `gapi.Preferences{}`. Does that seem reasonable?

grafana/resource_organization_preferences.go
mdb: @julienduchesne I agree; I suspect this was early-stage copy pasta. I've updated `UpdateOrganizationPreferences` to simply return `CreateOrganizationPreferences`, given that both use [gapi.UpdateAllOrgPreferences](https://github.com/grafana/grafana-api-golang-client/blob/master/org_preferences.go#L38) under the hood. Is that reasonable?

grafana/resource_organization_preferences_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"testing"
+
+	gapi "github.com/grafana/grafana-api-golang-client"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+func TestAccResourceOrganizationPreferences(t *testing.T) {
+	CheckOSSTestsEnabled(t)
+	CheckOSSTestsSemver(t, ">=8.0.0")
```
mdb, line 16: Is this `CheckOSSTestsSemver` acceptable? Without it, it seems that `WeekStart` is always `""` in at least Grafana 7.5.15, which results in acceptance test failures like...

```
=== RUN   TestAccResourceOrganizationPreferences
    resource_organization_preferences_test.go:14: Step 1/1 error: Check failed: Check 6/6 error: grafana_organization_preferences.test: Attribute 'week_start' expected "Tuesday", got ""
--- FAIL: TestAccResourceOrganizationPreferences (0.47s)
```

...as seen here: https://drone.grafana.net/grafana/terraform-provider-grafana/1741/11/4

mdb:
> This would need some tests. Other than that and my comments, this looks nice!

@julienduchesne Would you mind offering another review, please? Are there any additional changes and/or tests you'd like to see?

grafana/resource_organization_preferences_test.go
julienduchesne: Yes, that works! I think v7 is not even supported anymore

julienduchesne (approved):
LGTM! Thanks for adding tests!

PR #644, opened 2022-09-08
Document that the notification policy is a singleton

### LINKED ISSUE #642: Adding new notification policy removes existing policies
### Terraform Version
* Terraform: 1.2.8
* Terraform Grafana Provider: 1.28.0
* Grafana: 9.1.3-5

### Affected Resource(s)
* grafana_notification_policy

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### Debug Output

### Panic Output

### Expected Behavior
Should have added a new notification policy on top of the existing ones.

### Actual Behavior
Removed all the policies set before manually and created the one added via terraform

julienduchesne:
This is a question that has come up a couple times since the alerting release Closes https://github.com/grafana/terraform-provider-grafana/issues/642

### DIFF
--- a/docs/resources/notification_policy.md
+++ b/docs/resources/notification_policy.md
@@ -3,11 +3,14 @@
 page_title: "grafana_notification_policy Resource - terraform-provider-grafana"
 subcategory: "Alerting"
 description: |-
-  Official documentation https://grafana.com/docs/grafana/latest/alerting/notifications/HTTP API https://grafana.com/docs/grafana/next/developers/http_api/alerting_provisioning/#notification-policies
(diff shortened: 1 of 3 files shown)


alexweav (approved):
Thanks! LGTM

NickLarsenNZ:
This is really strange behaviour. There's no warning, no safety :/
Documentation is one thing, but it should really be built safely to prevent multiple resources (terraform doesn't really have the concept of singletons).

I assume there are some reasons that it had to be one resource, instead of being two separate resources (a default policy, and then specific policies).

PR #647, opened 2022-09-09
RBAC: resource for RBAC role assignments

### LINKED ISSUE #380: Allow to map roles directly to users, teams and service accounts

IevaVasiljeva:
**What this PR does / why we need it**:
Adds RBAC role assignment resource that allows using Terraform to assign RBAC roles to users, service accounts and teams.

**Which issue(s) this PR fixes**:

Partially resolves https://github.com/grafana/terraform-provider-grafana/issues/380

**Special notes for your reviewer**:

Related Grafana's Golang API client PR: https://github.com/grafana/grafana-api-golang-client/pull/110
Related Grafana's backend PR: https://github.com/grafana/grafana-enterprise/pull/3823

Design doc for this work: https://docs.google.com/document/d/1HA2nhkEGjlYQrvofbX80RQqy9OquqByFDpminWafBXM/edit#heading=h.axfzr8w1lw7j


grafana/resource_role_assignment.go
```diff
@@ -0,0 +1,186 @@
...
+					},
+				},
+			},
+		},
+	}
+}
+
+func ReadRoleAssignments(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*client).gapi
+	uid := d.Get("role_uid").(string)
```
julienduchesne: A read function should always depend solely on the ID. If the role UID is the ID of the resource then: 
- That means that there can only be one role assignment per role. Is that OK?
- The Read function should start with `uid := d.GetID()` and all attributes should be set from the subsequent return of the API

This is needed for import to work. What `terraform import` does is it sets the ID and then triggers a read on the resource to fill in the details

grafana/resource_role_assignment.go
```diff
@@ -0,0 +1,186 @@
...
+	if err := setRoleAssignments(assignments, d); err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId(uid)
+	return nil
+}
+
+func UpdateRoleAssignments(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	if !d.HasChange("users") && !d.HasChange("teams") && !d.HasChange("service_accounts") {
```
julienduchesne: hmm, you may need to check `d.IsNewResource()` as well for creation to work reliably in all cases
```suggestion
	if !d.IsNewResource() && !d.HasChange("users") && !d.HasChange("teams") && !d.HasChange("service_accounts") {
```

julienduchesne (commented):
If possible, a test would be nice. I can't run them myself because I don't really know how :smile: but I'd like to run them in CI eventually

Jguer (approved):

grafana/resource_role_assignment.go
```diff
@@ -0,0 +1,147 @@
...
+	gapi "github.com/grafana/grafana-api-golang-client"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func ResourceRoleAssignment() *schema.Resource {
+	return &schema.Resource{
+		Description: `
+**Note:** This resource is available only with Grafana Enterprise 9.2+.
+* [Official documentatigrafana_role_assignmenton](https://grafana.com/docs/grafana/latest/enterprise/access-control/)
```
linoman:
```suggestion
* [Official documentation](https://grafana.com/docs/grafana/latest/enterprise/access-control/)
```

linoman: I've fixed this myself.

grafana/resource_role_assignment.go
```diff
@@ -0,0 +1,152 @@
...
+	gapi "github.com/grafana/grafana-api-golang-client"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func ResourceRoleAssignment() *schema.Resource {
+	return &schema.Resource{
+		Description: `
+**Note:** This resource is available only with Grafana Enterprise 9.2+.
+* [Official documentation](https://grafana.com/docs/grafana/latest/enterprise/access-control/)
```
vtorosyan:
```suggestion
* [Official documentation](https://grafana.com/docs/grafana/latest/administration/roles-and-permissions/access-control/)
```
The location has changed (we can update the rest of the resources in a separate PR).

grafana/resource_role_assignment.go
```diff
@@ -0,0 +1,152 @@
...
+* [HTTP API](https://grafana.com/docs/grafana/latest/http_api/access_control/)
```
vtorosyan:
```suggestion
* [HTTP API](https://grafana.com/docs/grafana/latest/developers/http_api/access_control/)
```

grafana/resource_role_assignment.go
```diff
@@ -0,0 +1,152 @@
...
+	if !d.HasChange("users") && !d.HasChange("teams") && !d.HasChange("service_accounts") {
+		return nil
+	}
+
+	client := meta.(*client).gapi
+
+	uid := d.Get("role_uid").(string)
+	users, err := collectRoleAssignmentsToFn(d.Get("users"))
+	if err != nil {
+		return diag.Errorf("invalid user IDs specifiedL %v", err)
```
vtorosyan:
```suggestion
		return diag.Errorf("invalid user IDs specified %v", err)
```

grafana/resource_role_assignment.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	teams, err := collectRoleAssignmentsToFn(d.Get("teams"))
+	if err != nil {
+		return diag.Errorf("invalid team IDs specifiedL %v", err)
```
vtorosyan:
```suggestion
		return diag.Errorf("invalid team IDs specified %v", err)
```

grafana/resource_role_assignment.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	serviceAccounts, err := collectRoleAssignmentsToFn(d.Get("service_accounts"))
+	if err != nil {
+		return diag.Errorf("invalid service account IDs specifiedL %v", err)
```
vtorosyan:
```suggestion
		return diag.Errorf("invalid service account IDs specified %v", err)
```

vtorosyan (approved):
Looks good to me - happy to merge this after the documentation is generated.

vtorosyan:
@IevaVasiljeva I think we also need to release the go-client if not already done, and bump the version here.

grafana/resource_role_assignment.go
```diff
@@ -0,0 +1,145 @@
...
+	assignments, err := client.UpdateRoleAssignments(ra)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	if err := setRoleAssignments(assignments, d); err != nil {
+		return diag.FromErr(err)
+	}
+
+	return nil
```
julienduchesne: Instead, we usually reuse the Read, so that we always have the same code to set attributes
```suggestion
	return ReadRoleAssignments(ctx, d meta)
```

julienduchesne (approved):
LGTM. One tiny comment

grafana/resource_role_assignment.go
vtorosyan: aeda30b61ef1b3ce244ecbfd4f5dd5c4df6749ff thnx

grafana/resource_annotation_test.go
```diff
@@ -164,31 +164,24 @@ resource "grafana_annotation" "test_with_dashboard_id" {
...
-  panel_id       = local.dashboard_json.panels[0].id
-}
-
 resource "grafana_annotation" "test_with_panel_id" {
     text     = "%s"
-		panel_id = local.panel_id
+		panel_id = %d
 }
-`, text, text, text)
+`, text, text, panelID, text, panelID)
```
IevaVasiljeva, line 186: Note: this change was added to make annotation resource tests pass. They were failing with the following error:
```
=== RUN   TestAccAnnotation_basic
    resource_annotation_test.go:24: Step 9/9 error running import: exit status 1
        Error writing state file: Failed to serialize resource instance in state: Instance data.grafana_dashboard.test_with_panel_id has status ObjectPlanned, which cannot be saved in state.
```

PR #651, opened 2022-09-13
Clean up Cloud API keys from tests

julienduchesne:
If a test fails, they aren't always deleted and they tend to add up This will clean them up before each test run. That way, we'll have at most one of each prefix


julienduchesne:
Dunno what's up with the cloud tests. Will try again tomorrow

inkel (approved):

PR #652, opened 2022-09-13
docs(rule_group): folder_uid wrong description

mentos1386:


inkel (approved):
Good catch! Thank you.

inkel:
Note that you need to run `go generate` from the repository's root to generate the docs; commit that and submit the changes again, please.

PR #656, opened 2022-09-14
Alerting: Workaround for bad handling of empty strings in contact points API

alexweav:
The Terraform default is to use empty-strings for blank or missing fields. This is consistent with a lot of Go-based clients, but Terraform is very aggressive with this. In fact, in many cases the provider API has no way of telling whether a field was missing from HCL entirely, or present with an empty string: https://github.com/hashicorp/terraform-plugin-sdk/issues/741

This is normally fine, assuming that the underlying API is well behaved. Most of Grafana's API handle this well, but the `settings` field in contact points do not handle this consistently. Empty strings circumvent the API validation entirely. See this issue in Grafana: https://github.com/grafana/grafana/issues/55139

This issue in Grafana will take quite some time to fix. This PR implements a much faster workaround in the provider. This allows us to get a workaround released to users faster, while we fix the main bug in the Grafana API. This also means the provider will continue to work with Grafana versions affected by the above issue, which is nice for compatibility.

We will now treat this settings blob as if it's a struct with all fields being `omitempty`. Empty settings values will simply not be sent to the server at all.

### DIFF
--- a/examples/resources/grafana_contact_point/_acc_default_settings.tf
+++ b/examples/resources/grafana_contact_point/_acc_default_settings.tf
@@ -1,0 +1,8 @@
+resource "grafana_contact_point" "default_settings" {
+    name = "Default Settings"
+
+    slack {
+        token = "xoxb-token"
+        recipient = "#channel"
+    }
+}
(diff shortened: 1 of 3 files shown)


JohnnyQQQQ (approved):
LGTM

marinnedea:
While it’s merged – it doesn’t look to be released yet and available in the main terraform registry.
Can you let us know when it would be released to the registry?

julienduchesne:
> While it’s merged – it doesn’t look to be released yet and available in the main terraform registry. Can you let us know when it would be released to the registry?

It's released in [v1.28.2](https://github.com/grafana/terraform-provider-grafana/releases/tag/v1.28.2)

chrisdetsicas:
@alexweav Thanks for this - I can confirm it works as intended for us now - notification policy provisioned by Terraform can trigger a pagerduty contact point provisioned by Terraform.

PR #684, opened 2022-10-14
Make alertmanager_uid configurable

### LINKED ISSUE #683: Configure alertmanager datasource for prometheus
Hi! I didn't find any mention of the possibility to configure alertmanager datasource for prometheus datasource via Terraform provider. 

This feature exists in the UI:
Screenshot 2022-10-14 at 12 35 09

I found that the option that controls this behaviour is `alertmanagerUid` and should be passed under `json_data`.

I don't mind to help and deliver this feature.

Jasstkn:
Signed-off-by: Mariia Kotliarevskaia <mariia.kotliarevskaia@gmail.com>

fixes #683

### DIFF
--- a/docs/resources/data_source.md
+++ b/docs/resources/data_source.md
@@ -124,6 +124,7 @@ resource "grafana_data_source" "stackdriver" {
 
 Optional:
 
+- `alertmanager_uid` (String) (Prometheus) The name of the Alertmanager datasource to manage alerts via UI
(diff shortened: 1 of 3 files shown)


julienduchesne:
Since it's already in the client, I don't mind merging this. Note that if you need new attributes that aren't in the client, you will need to convert your resources to use `json_data_encoded` instead, which supports an arbitrary map of json_data. We're deprecating the `json_data` attribute

julienduchesne (approved):

PR #692, opened 2022-10-25
Access control: Fix a bug with data source permission resource and allow assigning permissions for basic roles

IevaVasiljeva:
**What:**
* fix a bug with data source permission assignments when RBAC is enabled (Admin role assignment was always returned by Grafana but was not correctly parsed);
* allow assigning data source permissions to basic roles (Viewer, Editor) (will work from Grafana 9.2.3+);
* update and reenable data source permission tests.

**Dependencies:**
* https://github.com/grafana/grafana-api-golang-client/pull/115;
* basic role assignments require https://github.com/grafana/grafana-enterprise/pull/4041 to be released, it should be a part of `v9.2.3`.


grafana/resource_datasource_permission_test.go
```diff
@@ -72,10 +70,8 @@ func testAccDatasourcePermissionCheckDestroy(datasourceID *int64) resource.TestC
 		if err != nil {
 			return fmt.Errorf("error getting datasource permissions %d: %s", *datasourceID, err)
 		}
-		if response.Enabled {
-			return fmt.Errorf("datasource permissions %d still enabled", *datasourceID)
-		}
-		if len(response.Permissions) > 0 {
+		// Data source permissions for Admins are always set and can't be removed through provisioning
+		if len(response.Permissions) > 1 || (len(response.Permissions) == 1 && response.Permissions[0].BuiltInRole != "Admin") {
```
kalleep: With enterprise the admin role could be missing if the basic role admin is modified to not have required actions

grafana/resource_datasource_permission.go
```diff
@@ -118,16 +128,22 @@ func ReadDatasourcePermissions(ctx context.Context, d *schema.ResourceData, meta
 		return diag.FromErr(err)
 	}
 
-	permissionItems := make([]interface{}, len(response.Permissions))
-	for i, permission := range response.Permissions {
+	var permissionItems []interface{}
```
kalleep: NIT: You should use the make function to construct the slice because we know the required size. But instead of setting the size as it was before we can set the capacity and still use the append function `make([]interface{}, 0, len(response.Permissions))`

grafana/resource_datasource_permission_test.go
IevaVasiljeva: This particular change is only for tests, so shouldn't make a difference.

But more generally it is a problem, and I don't really know how to handle it. I think similar issues would arise if DS permissions were granted to viewers/editors/specific teams or users - if Terraform doesn't reflect this, it'll attempt to remove the changes but will fail at it. I'm not sure how such cases are usually handled through Terraform 🤔

IevaVasiljeva: One option would be to require that the users always specify the permissions that can't be changed in their Terraform configs - here it'd mean always including the following (note that Terraform doesn't yet support `Edit` permission):
`
  permissions {
    built_in_role = "Admin"
    permission    = "Query"
  }
`

grafana/resource_datasource_permission.go
```diff
@@ -119,15 +129,21 @@ func ReadDatasourcePermissions(ctx context.Context, d *schema.ResourceData, meta
 	}
 
 	permissionItems := make([]interface{}, len(response.Permissions))
-	for i, permission := range response.Permissions {
+	for _, permission := range response.Permissions {
```
inkel: Note that the line above creates a non-empty slice of this type, meaning that in line 146, when you append, you're appending to the slice and thus will end up with `2 * len(response.Permissions)`, see [this example](https://go.dev/play/p/6BN7o9WGyUm).

What you should do is keep `i` and assign to `permissionItems[i]` as it was done previously.

inkel (changes requested):

grafana/resource_datasource_permission.go
IevaVasiljeva: Ouph, that was a typo, I forgot to specify both length and capacity for the slice. Thanks for pointing it out! I've fixed it now, can you take another look at it please?

IevaVasiljeva: Update - we decided to take a different route with one of the issues that this PR was solving, so the change in the loop isn't necessary at all anymore. I'll remove it and restore this code to what it was before.

inkel: Cool. Let me know when ready to be reviewed again.

kalleep (approved):

grafana/resource_datasource_permission_test.go
```diff
@@ -72,9 +70,6 @@ func testAccDatasourcePermissionCheckDestroy(datasourceID *int64) resource.TestC
 		if err != nil {
 			return fmt.Errorf("error getting datasource permissions %d: %s", *datasourceID, err)
 		}
-		if response.Enabled {
-			return fmt.Errorf("datasource permissions %d still enabled", *datasourceID)
-		}
```
IevaVasiljeva, line 77: Removing this, as data source permissions are always enabled when RBAC is enabled (which is the default state now).

Jguer (approved):

inkel (approved):

PR #695, opened 2022-10-26
Access Control: Update example with permissions that are available for cloud users

IevaVasiljeva:
**What?**

Updating role resource example with permissions that are accessible by Cloud as well as Enterprise users. 

**Why?**

Permissions that were used in role resource example are not accessible by cloud users, which was causing confusion (related issue: https://github.com/grafana/support-escalations/issues/4275)


gamab (approved):
Thanks @IevaVasiljeva

PR #704, opened 2022-10-28
Add `terraform fmt` check

julienduchesne:
We currently have a bunch of examples in our docs that are formatted weirdly


inkel (approved):
**THANK YOU**

PR #707, opened 2022-11-01
Allow Machine Learning jobs to be specified using datasource UID

sd2k:
The Grafana Machine Learning API now supports jobs to specify their datasource
using UID; this PR updates the Terraform provider to also allow this.

### DIFF
--- a/docs/resources/machine_learning_job.md
+++ b/docs/resources/machine_learning_job.md
@@ -17,7 +17,6 @@ A job defines the queries and model parameters for a machine learning task.
 
 ### Required
 
-- `datasource_id` (Number) The id of the datasource to query.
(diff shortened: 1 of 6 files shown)


csmarchbanks (approved):

julienduchesne (commented):
You have a failing test following this change

sd2k:
> You have a failing test following this change

Yep, that was a bug in the actual API, fixed now! There's a new failing test but I don't think it's related to my change, seems like a transient thing.

julienduchesne:
> > You have a failing test following this change
> 
> Yep, that was a bug in the actual API, fixed now! There's a new failing test but I don't think it's related to my change, seems like a transient thing.

The account where we run our tests is broken in Grafana.com. I'll check tomorrow. In the meantime, I'll force merge your PR

PR #716, opened 2022-11-10
Tests: Get all configs from Vault

julienduchesne:
This allows us to switch to a different Cloud account 
There is currently a bug deleting Grafana Cloud instances in the current account so I'm switching over to a new one


inkel (approved):
Nice! Thank you.

PR #718, opened 2022-11-10
CI: Do not run Cloud tests automatically

julienduchesne:
This will lead to faster runtimes and less flakiness. We rarely have to run those tests, we have very few cloud resources


inkel (approved):
Good idea!

Issue #698, opened 2022-10-27
Incorrect reporting of users existency while creating an organization


wilfriedroset:
### Terraform Version
* Terraform: 1.2.6
* Terraform Grafana Provider: 1.30.0
* Grafana: 8.5.9

### Affected Resource(s)
* grafana_organization

### Terraform Configuration Files
```hcl
data "grafana_user" "from_login" {
  login = "wilfried.roset"
}

output "test" {
  value = data.grafana_user.from_login
}

resource "grafana_organization" "org" {
  create_users = false # users are automatically created upon connection via SSO
  name         = "test"
  admins       = [data.grafana_user.from_login.email]
}

terraform {
  required_providers {
    grafana = {
      source  = "grafana/grafana"
      version = "1.30.0"
    }
  }
  required_version = ">= 1.2.6"
}
provider "grafana" {}
```

### Debug Output
https://gist.github.com/wilfriedroset/26d50863225deea3eed0a36c2d101c2c

### Expected Behavior
Organization is successfully created and users are added into it with the correct privileges.

### Actual Behavior
The provider report that the user does not exists which is not true.

My understanding is that during `addIdsToChanges`, the provider first get the list of known users by calling 
https://github.com/grafana/terraform-provider-grafana/blob/a6cb3f36d6eef34c34bbb36357e2959be8228478/grafana/resource_organization.go#L284
Then a map is created to make check easier:
https://github.com/grafana/terraform-provider-grafana/blob/a6cb3f36d6eef34c34bbb36357e2959be8228478/grafana/resource_organization.go#L288-L290
And finally a check is perform to verify if the users we are trying to add actually exists in grafana:
https://github.com/grafana/terraform-provider-grafana/blob/a6cb3f36d6eef34c34bbb36357e2959be8228478/grafana/resource_organization.go#L294-L301

What is actually checked is that users are in the first 1000 users of Grafana. Indeed, there is pagination on  [/api/users](https://grafana.com/docs/grafana/latest/developers/http_api/user/#search-users) and by default only the first 1000 are return, which means if a user register after the 1000 (alphabetical order) it will be reported as not existing.

The underlying client does not follow the pagination
https://github.com/grafana/grafana-api-golang-client/blob/1bc12ec065ceeae1f9004dffbe2ea6bfcc112cbc/user.go#L48-L51

I reckon there is two solutions:
* set `perpage=<big_value>` when calling /api/users but we will be subject to this bug if Grafana has more users than `big_value`
[...]

PR #725, opened 2022-11-13
Allow to ignore missing users when creating team/organization

wilfriedroset:
In the context where Grafana relies on an external entity to create users automatically, this could be corporate SSO. Example:

```yaml
  auth:
    disable_login_form: true
    disable_signout_menu: true
  auth.anonymous:
    enabled: false
  auth.proxy:
    enabled: true
    header_name: "X-REMOTE-USER"
  users:
    allow_org_create: true
    allow_sign_up: false
    auto_assign_org: true
    auto_assign_org_role: "Viewer"
    viewers_can_edit: true
```

Users might exist in Grafana at the apply which could be acceptable for the administrator in charge of the instance. Without this setting the apply will fail and this is undetected during the plan. The setting allow to skip the missing users and allow the apply to complete with a warning message.

As an alternative to detect the missing users one can use the following snippet:

```tf
data "grafana_user" "existency_check" {
  count = length(var.members)
  email = var.members[count.index]
}
resource "grafana_team" "example" {
  name    = var.team_name
  email   = var.email
  members = concat(data.grafana_user.existency_check[*].email, [var.admin_email])
}
```

As data are fetch during plan this causes terraform to return an error for missing users. This is not optimal as this crowd the state and the output of terraform. Moreover if users are added in several teams/orgs their data are fetched multiple time.

Finally, depending on the size of the instance being managed with terraform this setting give more flexibility to admin.
Relates to: #698
Signed-off-by: Wilfried Roset <wilfriedroset@users.noreply.github.com>

### DIFF
--- a/grafana/resource_organization_test.go
+++ b/grafana/resource_organization_test.go
@@ -47,7 +47,7 @@ func TestAccOrganization_basic(t *testing.T) {
 				ResourceName:            "grafana_organization.test",
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"admins", "admin_user", "create_users"}, // Users are imported explicitly (with create_users == false)
+				ImportStateVerifyIgnore: []string{"admins", "admin_user", "create_users", "ignore_missing_users"}, // Users are imported explicitly (with create_users == false)
 				ImportStateCheck: func(s []*terraform.InstanceState) error {
 					if len(s) != 1 {
 						return fmt.Errorf("expected 1 state: %#v", s)
@@ -123,6 +123,18 @@ func TestAccOrganization_users(t *testing.T) {
 					),
 					resource.TestCheckResourceAttr(
 						"grafana_organization.test", "viewers.#", "0",
+					),
+				),
+			},
+			{
+				Config: testAccOrganizationConfig_ignoreMissingUsers,
+				Check: resource.ComposeTestCheckFunc(
+					testAccOrganizationCheckExists("grafana_organization.test", &org),
+					resource.TestCheckResourceAttr(
+						"grafana_organization.test", "name", "terraform-acc-test-missing-user",
+					),
+					resource.TestCheckResourceAttr(
+						"grafana_organization.test", "admins.#", "1",
 					),
 				),
 			},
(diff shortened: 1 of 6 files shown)


grafana/resource_organization_test.go
```diff
@@ -126,6 +126,18 @@ func TestAccOrganization_users(t *testing.T) {
...
 			},
+			{
+				Config: testAccOrganizationConfig_ignoreMissingUsers,
+				Check: resource.ComposeTestCheckFunc(
+					testAccOrganizationCheckExists("grafana_organization.test", &org),
+					resource.TestCheckResourceAttr(
+						"grafana_organization.test", "name", "terraform-acc-test-missing-user",
+					),
+					resource.TestCheckResourceAttr(
+						"grafana_organization.test", "admins.#", "1",
```
wilfriedroset, line 137: point to discuss: currently the state contains exactly what have been provided by the user but as we prune the list when `ignore_missing_users=true` should we adjust what is stored in the state?
note: The simpler is to not modify the content of the state as it adds more code logic and might not be counter intuitive.

The test is left failing purposely to avoid merging this without having the discussion.

The same logic should be applied for resource_team as well.

julienduchesne:
Not really sure about this one. It goes against the Terraform way of doing things. When this attribute is set, if there are any missing users, the state will be left inconsistent after apply. (Terraform will always propose to add the missing users on plan)

I think the good way of doing this would be to add `grafana_users` datasource which allows you to query the whole list of users. Then you can use it like this:
```
data "grafana_users" "all_users" {}

resource "grafana_team" "example" {
  name    = var.team_name
  email   = var.email
  members = setintersection(toset(["myuser@example.com", "myuser2@example.com]), [for user in data.grafana_users.all_users.users: user.email])
}
```

This wouldn't lead to any inconsistent apply. Only when a listed user is created, will the Terraform plan show that a new user is being added to the team/org

wilfriedroset:
> I think the good way of doing this would be to add grafana_users datasource which allows you to query the whole list of users

This idea solves the issue as well and it better fits terraform.

wilfriedroset:
Let's close this PR which is not the best way to address the issue and continue the discussion in https://github.com/grafana/terraform-provider-grafana/pull/734

PR #743, opened 2022-12-05
Allow configuring a report using a dashboard UID

julienduchesne:
The whole structure of the reports API has changed. It now allows multiple dashboards.
However, I'm not implementing this in this PR. I'm just changing the API to allow using UIDs because they are now the standard way to identify dashboards.

If we want to support multiple dashboards, it should be a completely different attribute anyways (like a `dashboard` block)

Needs: https://github.com/grafana/grafana-api-golang-client/pull/126


go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/grafana/terraform-provider-grafana
 
 go 1.18
 
+replace github.com/grafana/grafana-api-golang-client => github.com/grafana/grafana-api-golang-client v0.14.1-0.20221205173354-b400961adebd
```
inkel: Considering https://github.com/grafana/grafana-api-golang-client/pull/126 was approved, will you remove this once that one is merged?

inkel (approved):

PR #744, opened 2022-12-05
Dashboard Permissions: Support dashboard UIDs

julienduchesne:
Using a dashboard's ID as a reference now deprecated: https://grafana.com/docs/grafana/v9.3/developers/http_api/dashboard_permissions/

Also, added import support

Needs: https://github.com/grafana/grafana-api-golang-client/pull/127


go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/grafana/terraform-provider-grafana
 
 go 1.18
 
+replace github.com/grafana/grafana-api-golang-client => github.com/grafana/grafana-api-golang-client v0.14.1-0.20221205183704-2e3d18a0df20
```
IevaVasiljeva: nit: should probably just bump the version once the next API client is released?

IevaVasiljeva (approved):
Nice, thanks for making this change! Looks good :)

go.mod
julienduchesne: Yup, this was just a temporary replace to test out the client

PR #761, opened 2022-12-22
Document how to run a single cloud test

julienduchesne:
Cloud tests are run on Drone promotions. 
We can pass env variables in those 
Using that, we can run targeted tests, which is nice because our test base creates lots of stacks which can cause issues


inkel (approved):
**THANK YOU**

PR #762, opened 2022-12-22
Use the `ParallelTest` helper on all possible tests

julienduchesne:
Test time goes from ~4.5 mins to ~3 mins. Should be even better when all tests are parallel

Added TODO comments for failing tests


inkel (approved):
This is great, thanks for taking care!

PR #764, opened 2023-01-04
Add support for Machine Learning holidays

sd2k:
Note that holidays can be created and added to jobs, but not vice
versa, despite the API supporting it. This is because the API
would return an updated value for the holiday's `jobs` field if we
exposed it via Terraform, and we have no way of updating that from
the job resource's CreateContext (as far as I'm aware).


sd2k:
I don't think the failing test is related to this change.

grafana/resource_machine_learning_holiday.go
```diff
@@ -0,0 +1,202 @@
...
+
+To use a holiday in a job, use its id in the "holidays" attribute of a "grafana_machine_learning_job":
+
+		resource "grafana_machine_learning_job" "test_job" {
+			...
+			holidays = [
+				grafana_machine_learning_holiday.my_holiday.id
+			]
+		}
+`,
```
julienduchesne: This will probably look weird in the docs without code blocks

```suggestion
		Description: `
A holiday describes time periods where a time series is expected to behave differently to normal.

To use a holiday in a job, use its id in the "holidays" attribute of a "grafana_machine_learning_job":

` + 
"```terraform`" +
`
resource "grafana_machine_learning_job" "test_job" {
    ...
    holidays = [
        grafana_machine_learning_holiday.my_holiday.id
    ]
}
` + "```",
```

grafana/resource_machine_learning_holiday.go
```diff
@@ -0,0 +1,202 @@
...
+	d.SetId(holiday.ID)
+	return resourceMachineLearningHolidayRead(ctx, d, meta)
+}
+
+func resourceMachineLearningHolidayRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	c := meta.(*client).mlapi
+	holiday, err := c.Holiday(ctx, d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
```
julienduchesne: There should be a case where the holiday is missing (404). In that case, we remove the resource from state (`d.SetId("")`) and print out a warning

julienduchesne (approved):
Very nice! LGTM other than two minor comments

grafana/resource_machine_learning_holiday.go
sd2k: Done in [ab3bb4a](https://github.com/grafana/terraform-provider-grafana/pull/764/commits/ab3bb4a889204bf09edf367351f9aedffb00f76b) (for ML jobs as well).

grafana/resource_machine_learning_holiday.go
sd2k: Good spot! I've pushed a manual fix as there was an extra ` in the suggestion, thanks though 🙂

csmarchbanks (approved):
Looks good from the ML side!

PR #767, opened 2023-01-06
Add default value for RBAC role's scope

IevaVasiljeva:
**What**

Adding a default value (empty string) for RBAC role scope.

**Why**

To solve an issue with provider showing permission changes for all permissions with empty scope even if they haven't changed. This is the output before the change (the only thing that was actually changed in the config was removal of the `datasources:read` permission for scope `datasources:uid:other`):
```
  # grafana_role.basic_role_viewer will be updated in-place
  ~ resource "grafana_role" "basic_role_viewer" {
        id           = "basic_viewer"
        name         = "basic:viewer"
      ~ version      = 10 -> 11
        # (6 unchanged attributes hidden)

      - permissions {
          - action = "dashboards.insights:read" -> null
        }
      + permissions {
          + action = "dashboards.insights:read"
        }
      - permissions {
          - action = "datasources.insights:read" -> null
        }
      + permissions {
          + action = "datasources.insights:read"
        }
      - permissions {
          - action = "datasources:read" -> null
          - scope  = "datasources:uid:other" -> null
        }

        # (16 unchanged blocks hidden)
    }
```

And this is the output after the change:
```
  # grafana_role.basic_role_viewer will be updated in-place
  ~ resource "grafana_role" "basic_role_viewer" {
        id           = "basic_viewer"
        name         = "basic:viewer"
      ~ version      = 10 -> 11
        # (6 unchanged attributes hidden)

      - permissions {
          - action = "datasources:read" -> null
          - scope  = "datasources:uid:other" -> null
        }

        # (18 unchanged blocks hidden)
    }
```

**Note:** I'm still not quite sure why all the permissions with empty scopes were flagged up as changes, but adding a default value seems to resolve it. Let me know if you can think of a better way.


vtorosyan:
So API response does not contain the scope at all when not set? What is weird that it tries also thinks that action has changed or something? Or do I read the output wrong?

The change is looks good to me, I guess the only thing we need to double check to see if we can have default value the same as what API returns.

vtorosyan (approved):

IevaVasiljeva:
> So API response does not contain the scope at all when not set? What is weird that it tries also thinks that action has changed or something? Or do I read the output wrong?
> 
> The change is looks good to me, I guess the only thing we need to double check to see if we can have default value the same as what API returns.

API response has a default value for scope, and it is an empty string as expected. I honestly don't know why Terraform reports this as a change - I tried debugging it, and the values from the config and the state look the same in the debugger. But setting a default value seems to resolve the problem, so I decided to timebox it and stop investigating. Let me know if you have any theories of what might be causing the diff.

IevaVasiljeva:
I've looked into this some more, and I found the following warning message in the logs:
```
2023-01-10T12:19:51.873Z [WARN]  Provider "registry.terraform.io/grafana/grafana" produced an unexpected new value for grafana_role.test_roles during refresh.
      - .permissions: planned set element cty.ObjectVal(map[string]cty.Value{"action":cty.StringVal("alert.notifications:read"), "scope":cty.NullVal(cty.String)}) does not correlate with any element in actual
```

As far as I understand, Terraform interprets the default value of the optional attribute `scope` as `cty.NullVal(cty.String)`, which is `nil`. However, when we read the data from the state, we interpret an empty scope as an empty string `""`. Therefore Terraform thinks that the state differs from the config for permissions with empty scope, and wants to remove and recreate all such permissions.

By setting a default value of an empty string, we ensure that an empty scope in config has the same representation as empty scope read from the state.

julienduchesne (approved):

Issue #773, opened 2023-01-10
Deleting a folder with an editor API key is flaky


julienduchesne:
Recently, a test was added to verify that a Grafana bug is fixed: https://github.com/grafana/terraform-provider-grafana/pull/763

However, the test is flaky, it fails more often than not: https://drone.grafana.net/grafana/terraform-provider-grafana/2041 because Terraform is not able to destroy the folder

```
Error: status: 403, body: {"accessErrorId":"ACE3296391336","message":"You'll need additional permissions to perform this action. Permissions needed: apikeys:delete","title":"Access denied"}
```

It does pass sometime, as evidenced by the PR that added it

PR #774, opened 2023-01-10
Fix flaky tests

julienduchesne:
One doesn't behave well in parallel, the other is flaky for reasons unknown, opened an issue for it (https://github.com/grafana/terraform-provider-grafana/issues/773)!


inkel (approved):
Thank you

PR #779, opened 2023-01-11
Fix "create folder from editor key" test

### LINKED ISSUE #773: Deleting a folder with an editor API key is flaky

julienduchesne:
Closes https://github.com/grafana/terraform-provider-grafana/issues/773
The issue was that you can't create a new `provider` instance from tests. You need to use the factories functions in the test framework
The tests were failing in a random fashion because it seems like the test framework was sometimes using one provider to delete the api key, sometimes using the other

To fix the test, I instead create an API key and inject it in envvars at the test level (not in the Terraform config)
I also added a test for Viewers, it's a sanity check to make sure the auth is inject correctly. Using a Viewer api key causes an error

### DIFF
--- a/grafana/resource_folder_test.go
+++ b/grafana/resource_folder_test.go
@@ -118,6 +118,7 @@ func TestAccFolder_createFromDifferentRoles(t *testing.T) {
 			var folder gapi.Folder
 			var name = acctest.RandomWithPrefix(tc.role + "-key")
 
+			// Create an API key with the correct role and inject it in envvars. This auth will be used when the test runs
 			client := testAccProvider.Meta().(*client).gapi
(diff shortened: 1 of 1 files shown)


inkel (approved):
Oh this is smart, good workaround!

Issue #747, opened 2022-12-08
Support Setting of org_id outside of provider block


eviscares:
Hey,

with the current state of terraform I think that the abstraction of tying org_id to the provider is proving a lot more troublesome, as you would for example need to create a grafana org in one module, pass out its org_id to a provider, and then pass that provider into a new module that can then for example create data sources and dashboards for an organization.
At least Datasources and Users should support this out of the box as you could just set the org_id on the request to create them.

I think this would really simplify working with the provider.

julienduchesne:
This is going to happen. I'm planning to introduce an `org_id` attribute to resources gradually. Stay tuned!

julienduchesne:
Remaining resources to support:
- Datasource
- Team
- Folder
- Datasources (`data_xxx`)

Planning to remove the global `org_id` afterwards because it complicates things and is confusing. Some resources support `org_id` but not all (example: user is org-agnostic and alerting doesn't use the same header, so the global setting doesn't work)

adkinsjd:
We are really excited about the addition of org_id to these resources - it would make our configuration much, much cleaner!

Is there a target timeline the completion of this project?

pschulten:
Is there something, you need help with?

PR #782, opened 2023-01-12
Add `org_id` to dashboards

julienduchesne:
Issue: https://github.com/grafana/terraform-provider-grafana/issues/747 Follow-up to https://github.com/grafana/terraform-provider-grafana/pull/776

This will allow such a config:

```terraform
resource "grafana_organization" "test" {
	name = "test"
}

	org_id      = grafana_organization.test.id
	config_json = jsonencode({
	  title = "dashboard-test"
	})
}
```

The main change here is that the ID of resources becomes: `<org id>:<resource id>`. This can be changed transparently because if the first block isn't there, it means the currently configured org

Also added some helpers for org IDs, as these will be used in every resource soon

### DIFF
--- a/grafana/oss_org_id.go
+++ b/grafana/oss_org_id.go
@@ -1,0 +1,43 @@
+package grafana
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	gapi "github.com/grafana/grafana-api-golang-client"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func makeOSSOrgID(orgID int64, resourceID interface{}) string {
+	return fmt.Sprintf("%d:%s", orgID, fmt.Sprint(resourceID))
+}
+
+func splitOSSOrgID(id string) (int64, string) {
+	if strings.ContainsRune(id, ':') {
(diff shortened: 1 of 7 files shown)


grafana/oss_org_id.go
```diff
@@ -0,0 +1,43 @@
...
+
+// nolint: unparam
+func clientFromOSSOrgID(meta interface{}, id string) (*gapi.Client, int64, string) {
+	orgID, restOfID := splitOSSOrgID(id)
+	client := meta.(*client).gapi
+	if orgID > 0 {
+		client = client.WithOrgID(orgID)
+	}
+	return client, orgID, restOfID
+}
```
inkel, line 34: In the two places this function is invoked we're not using the second returned param (`orgID`), does still make sense to return it?

inkel (approved):
Left a question but approving nonetheless because the changes look good and it's not a blocking question.

grafana/oss_org_id.go
julienduchesne: I have cases that will use it in the future (this is one of multiple PRs)

inkel: Cool, looking forward to reviewing those!

PR #788, opened 2023-01-18
auth: Make access policy token's `expires_at` optional

cinaglia:
Relates to https://github.com/grafana/grafana-api-golang-client/pull/132.

In response to https://github.com/grafana/support-escalations/issues/4823, this PR updates the `resource_cloud_access_policy_token` resource to reflect the fact that the `expires_at` field is optional.


grafana/resource_cloud_access_policy_token_test.go
```diff
@@ -80,6 +84,18 @@ func TestResourceCloudAccessPolicyToken_Basic(t *testing.T) {
 					resource.TestCheckResourceAttr("grafana_cloud_access_policy_token.test", "expires_at", expiresAt),
 				),
 			},
+			{
+				Config: testAccCloudAccessPolicyTokenConfigBasic("initial-no-expiration", "", initialScopes, ""),
```
julienduchesne: Can you split this off into its own test?
`TestResourceCloudAccessPolicyToken_NoExpiration` or something

Each step of a test is incremental so I feel like this doesn't belong here. Also, if it were a different test, you could check that the import works also when expiration is not set (last step)

cinaglia: Great suggestion. Addressed in [af75129](https://github.com/grafana/terraform-provider-grafana/pull/788/commits/af751297262f65e5e0389173fdf4cbf5050ae600). PTAL.

julienduchesne (approved):
LGTM! Both tests passed in CI

PR #789, opened 2023-01-18
Update list of stack regions and avoid duplication

cinaglia:
This PR updates the list of stack regions available to the latest options. It also consolidates the two lists into a single one with helpers for generating the list of slugs and the friendly description.

Ideally, this would be fetched using the `GET /api/stack-regions` endpoint, but this seems like a reasonable compromise for the time being.


cinaglia:
Closing in favor of https://github.com/grafana/terraform-provider-grafana/pull/787.

PR #790, opened 2023-01-20
Configure User-Agent for OnCallClient

Konstantinov-Innokentii:
This PR adds terraform-provider  User-Agent header to the OnCallClient.


julienduchesne (approved):
LGTM!

PR #792, opened 2023-01-23
Grafana Cloud: Support HTTP headers

julienduchesne:
This will be used for Grafana employees to test against internal Grafana.com instances which are protected behind IAP


monodot (approved):
Tested this out this morning and it works great, headers are being passed.

PR #795, opened 2023-01-25
Fix removal of old team members when new are added

dlukman:
The grafana_team resource removes existing users from teams when new ones are added. This change should fix it.

How to reproduce the issue:
- Add new members to existing team members using the grafana_team resource and apply
- Run terraform plan and observe that all users previously there were removed and want to be re-added
- Run again, the old ones are re-added, the new ones are removed (repeats on each rerun)

TF version: 1.3.7
Grafana version: 9.1.5

### DIFF
--- a/grafana/resource_team.go
+++ b/grafana/resource_team.go
@@ -71,7 +71,7 @@ A set of email addresses corresponding to users who should be given membership
 to the team. Note: users specified here must already exist in Grafana.
 `,
 				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
-					if old == new || (new == "[]" && old == "") || (new == "" && old == "[]") {
+					if (new == "[]" && old == "") || (new == "" && old == "[]") {
 						return true
 					}
 					return false
--- a/grafana/resource_team_test.go
+++ b/grafana/resource_team_test.go
@@ -3,9 +3,11 @@ package grafana
 import (
 	"fmt"
 	"strconv"
+	"strings"
 	"testing"
 
 	gapi "github.com/grafana/grafana-api-golang-client"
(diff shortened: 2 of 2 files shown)


julienduchesne:
Did you test this fix? It shouldn't do anything, IIUC, because new == old is implicit, it's kind of a duplicate condition right now

dlukman:
Yes, I tested it. The removal of this condition made it behave as expected.

JuanJoanJon:
We need this ASAP :-)

julienduchesne:
> We need this ASAP :-)

I'd like to have a test on this before merging. From the description of it, it should be testable

pascal-hofmann:
[This](https://github.com/grafana/terraform-provider-grafana/pull/780/) introduced the issue and did not need tests. It would be great to have a consistent policy for PRs. 😬

Maybe just revert #780 for now?

Edit: I just checked. The issue is not present in version 1.33.0 of the provider.

dlukman:
@julienduchesne Unfortunately I don't have the time to follow the proper process. I understand it's not optimal. I hoped it would help resolve it faster. If you wish to add tests or improve upon the change, feel free to do so.

I currently have it locked to 1.33.0 which doesn't have the issue as mentioned earlier.

julienduchesne:
> @julienduchesne Unfortunately I don't have the time to follow the proper process. I understand it's not optimal. I hoped it would help resolve it faster. If you wish to add tests or improve upon the change, feel free to do so.
> 
> I currently have it locked to 1.33.0 which doesn't have the issue as mentioned earlier.

I got it. I'll try to reproduce this in tests

julienduchesne:
I added a test that reproduces the bug. Confirmed that the fix works! Thanks @dlukman

julienduchesne (approved):

PR #804, opened 2023-02-03
Dashboard with org id: Additional validation

julienduchesne:
In the tests deploying a dashboard in a specific org, additionally validate that the given dashboard is NOT in the default org 
This is a sanity check to make sure the OrgID logic does not get broken in the future 
If the dashboard exists when passing the orgID but does not when using the default org, it means that the org attribute works as expected


inkel (approved):
Great idea, thanks,

PR #810, opened 2023-02-06
Split resources and provider into different packages

julienduchesne:
Follow-up to https://github.com/grafana/terraform-provider-grafana/pull/807 

Following the `terraform-provider-aws` repository structure, the main provider instance is in its own package, and resources are separated from it 

For now, all resource are in a `grafana` package but I will split this off into multiple resource packages


inkel (approved):
Fantastic work!

PR #811, opened 2023-02-06
Split up resources into multiple folders

julienduchesne:
`cloud`, `grafana`, `machinelearning`, `oncall` and `syntheticmonitoring`

`grafana` may be split again in the future but it'll do for now


inkel (approved):
Thank you thank you thank you!

PR #823, opened 2023-02-20
`grafana_api_key`: Support `org_id`

julienduchesne:
Allows provisioning an API key in an org other than the one defined in the provider config

### DIFF
--- a/internal/resources/grafana/resource_api_key.go
+++ b/internal/resources/grafana/resource_api_key.go
@@ -25,6 +25,7 @@ Manages Grafana API Keys.
 		DeleteContext: resourceAPIKeyDelete,
 
 		Schema: map[string]*schema.Schema{
+			"org_id": orgIDAttribute(),
 			"name": {
 				Type:     schema.TypeString,
 				Required: true,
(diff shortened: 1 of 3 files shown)


internal/resources/grafana/resource_api_key.go
```diff
@@ -146,13 +148,19 @@ func resourceAPIKeyDelete(ctx context.Context, d *schema.ResourceData, m interfa
 	return nil
 }
 
-func getClientForAPIKeyManagement(d *schema.ResourceData, m interface{}) (c *gapi.Client, cleanup func() error, err error) {
+func getClientForAPIKeyManagement(d *schema.ResourceData, m interface{}) (c *gapi.Client, cleanup func() error, err error, orgID int64) {
```
inkel: I know this is a change with a massive blast radius, but can we keep the Goism of returning errors as the last returned value? 🙏🏽

inkel (changes requested):
Other than my comment of returning errors as the last returned value, this LGTM.

internal/resources/grafana/resource_api_key.go
julienduchesne: Yeah, of course. Dunno how I let that slip, that's a very weird func sig. Fixed!

inkel (approved):

PR #829, opened 2023-02-27
add 'handleGrafanaManagedAlerts' argument for Alertmanager to docs

github-frank-fidanza:
Hello,

I'm willing to add this argument for the *grafana_data_source* [resource documentation](https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/data_source) that allows to check the "Receive Grafana Alerts" checkbox. It shows in the console browser network tab and works in Terraform code but it seems it hasn't been added to the docs just yet.
Also the `implementation` argument does not mention 'mimir' as an option, which appears in the GUI dropdown.

Screenshot :
image


julienduchesne:
We're not accepting any changes to the `json_data` attribute, it is deprecated. You can use the `json_data_encoded` attribute instead: https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/data_source#json_data_encoded. It supports an arbitrary json payload, and therefore, all possible attributes

PR #836, opened 2023-03-07
Support anonymous login

### LINKED ISSUE #10: Allow anonymous access to Grafana
In Grafana it is possible to open access to everyone. By doing so, you don't need to provide an `Authorization` header with your `POST` call.

The `terraform-provider-grafana` does not support this concept yet. The `auth` field must be provided and even when keeping it empty, the provider always sends an `Authorization` header with every request.

Can we make it optional to send the `auth` token or at least give us the option to disable auth completely within the provider?

julienduchesne:
Just closing https://github.com/grafana/terraform-provider-grafana/issues/10, the oldest issue still open. It was literally a 5 minute fix

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -204,7 +204,7 @@ resource "grafana_oncall_escalation" "example_notify_step" {
 
 ### Optional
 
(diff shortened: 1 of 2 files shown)


inkel (approved):
Thank you 🙏🏽

PR #838, opened 2023-03-07
Deprecate `grafana_team_preferences`

julienduchesne:
This resource doesn't make much sense:
- It has a one-to-one relationship with a team resource
- Teams are always user-created so users can definitely manage the preferences in the same resource
- It doesn't have a true delete, it clears out the preferences on delete
- It still uses the dashboard ID. This PR converts to UID (Which is needed in Grafana 9.4+ and is why tests don't run, currently: https://github.com/grafana/terraform-provider-grafana/pull/837)

Also, I've removed deprecated resources from docs


internal/resources/grafana/resource_team.go
```diff
@@ -126,10 +161,23 @@ func ReadTeam(ctx context.Context, d *schema.ResourceData, meta interface{}) dia
...
 	}
-	if err := ReadMembers(d, meta); err != nil {
+
+	preferences, err := client.TeamPreferences(teamID)
+	if err != nil {
 		return diag.FromErr(err)
 	}
-	return nil
+
+	if preferences.Theme+preferences.Timezone+preferences.HomeDashboardUID != "" {
```
inkel, line 170: That's smart 😸

tools/generate-docs.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # Generate docs!
-go run github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs
+go run github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs -ignore-deprecated
```
inkel, line 4: This probably should have been a different PR 🤔

inkel (approved):

tools/generate-docs.sh
julienduchesne: In retrospect, yeah. Sorry about that

inkel: No worries 😉

Issue #827, opened 2023-02-22
Terraform: Creating multiple resources (investigate for SAs) makes APIs non-responsive


eleijonmarck:
**what/why**
We have users provision service accounts via terraform when creating multiple resources of service accounts ~20 the APIs get non-responsive, and terraform creates a loop of trying to create the SAs.

This results in a invalid state between grafana and terraform. Which can only be solved by `terraform import` each resource. This is a time consuming and yields creation of multiple resources hard to use in terraform.

**DoD**
- [x] investigate why API calls for creating ~20 service accounts through terraform is not successful
- [ ] investigate if this is consistent with creating any resource or only service accounts

**investigation**
- locally I can create 20 service accounts without problem
- cloud is consistent that the API deems unresponsive when creating multiple SAs at the same time
- 
**note**
This might not be specific to service accounts, should be tested also with creating users

the specific terraform that can be used to try this out:
```tf
terraform {
  required_providers {
    grafana = {
      source  = "grafana/grafana"
    }
  }
}

# Declaring the first provider to be only used for creating the cloud-stack
provider "grafana" {
  alias = "first"

  cloud_api_key = "<API-Key>" 
}

resource "grafana_cloud_stack" "eleijonmarck" {
  provider = grafana.first

  name        = "eleijonmarck.grafana.net"
  slug        = "eleijonmarck"
  region_slug = "eu" # Example “us”,”eu” etc
}

# Creating an API key in Grafana instance to be used for creating resources in Grafana instance
resource "grafana_api_key" "cloud_api_key" {
  provider = grafana.first

  cloud_stack_slug = grafana_cloud_stack.eleijonmarck.slug
  name             = "cloud_api_key"
  role             = "Admin" 
}

# Declaring the second provider to be used for creating resources in Grafana        
provider "grafana" {
  alias = "second"

  url  = grafana_cloud_stack.eleijonmarck.url
  auth = grafana_api_key.cloud_api_key.key
}

# Creating a service account in Grafana instance to be used as auth
resource "grafana_service_account" "eleijonmarck_service_account" {
  provider = grafana.second

  name             = "eleijonmarck-sa"
  role             = "Admin"
  is_disabled = false
}

# Creating a service account token in Grafana instance to be used for creating resources in Grafana instance
resource "grafana_service_account_token" "eleijonmarck_service-account_token" {
  provider = grafana.second

  name               = "eleijonmarck-sa-token"
[...]

PR #842, opened 2023-03-08
Provider: MaxConnsPerHost tests with creating 60 service accounts

eleijonmarck:
**why/what**
We have seen that too many simultaneous calls to grafana API can cause the DB to be congested.

**Features**
- tests for creating 60 service accounts
- limit the number of HTTP connections that each provider can have open

**downside**
might slow down terraform providers for our users

**fixes**
https://github.com/grafana/terraform-provider-grafana/issues/827

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -340,6 +340,7 @@ func createGrafanaClient(d *schema.ResourceData) (string, *gapi.Config, *gapi.Cl
 	auth := strings.SplitN(d.Get("auth").(string), ":", 2)
 	cli := cleanhttp.DefaultClient()
 	transport := cleanhttp.DefaultTransport()
+	transport.MaxConnsPerHost = 2
 	transport.TLSClientConfig = &tls.Config{}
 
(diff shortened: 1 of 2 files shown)


internal/provider/provider.go
```diff
@@ -340,6 +340,7 @@ func createGrafanaClient(d *schema.ResourceData) (string, *gapi.Config, *gapi.Cl
 	auth := strings.SplitN(d.Get("auth").(string), ":", 2)
 	cli := cleanhttp.DefaultClient()
 	transport := cleanhttp.DefaultTransport()
+	transport.MaxConnsPerHost = 2
```
eleijonmarck: limiting the amount of concurrent HTTP connections from the provider makes it not overload the API and DB

julienduchesne: Can you make that a comment in the code?

julienduchesne (approved):
LGTM, thanks!

IevaVasiljeva (approved):

PR #858, opened 2023-03-23
Service accounts: Create Grafana service accounts through cloud provider

IevaVasiljeva:
New resources for creating Grafana service accounts and service account tokens through the cloud provider.

Changes:
* add two new resources (`grafana_cloud_stack_service_account` and `grafana_cloud_stack_service_account_token`);
* tiny fix to allow changing service account name.

Related platform changes: https://github.com/grafana/grafana-com/pull/6136


examples/resources/grafana_cloud_stack_service_account_token/resource.tf
```diff
@@ -0,0 +1,17 @@
+resource "grafana_cloud_stack_service_account" "cloud_sa" {
+  stack_slug = "<your stack slug>"
+
+  name        = "cloud service account"
+  role        = "Admin"
+  is_disabled = false
+}
+
+resource "grafana_service_account_token" "foo" {
```
julienduchesne:
```suggestion
resource "grafana_cloud_stack_service_account_token" "foo" {
```

julienduchesne (commented):
Just a quick initial comment, we can discuss this monday. What do you think about have a single resource for the SA AND the token? My point of view is that the "raison d'être" of this resource is to create a management entrypoint into the stack.

After that management service account is created, users can use the grafana provider to create more service accounts as they wish

We could call it `grafana_cloud_stack_management_token` or something

julienduchesne:
Also, other upper-level comment

> Related platform changes: [grafana/grafana-com#6136](https://github.com/grafana/grafana-com/pull/6136)

This is not being used AFAICT, right? To do so, we'd have to modify https://github.com/grafana/grafana-api-golang-client/blob/master/cloud_grafana_api_key.go#L28-L57

But if we think about this, it doesn't make sense to create a temp client anymore. The temporary client logic was built because the `api_key` resource was used for both cloud and grafana itself. Now that it's a separate resource for cloud, and if we have an entrypoint into instances to directly manage service accounts, we should use that instead of creating a temporary client

What do you think?

docs/resources/cloud_stack_service_account_token.md
```diff
@@ -0,0 +1,64 @@
...
+```terraform
+resource "grafana_cloud_stack_service_account" "cloud_sa" {
+  stack_slug = "<your stack slug>"
+
+  name        = "cloud service account"
+  role        = "Admin"
+  is_disabled = false
+}
+
+resource "grafana_service_account_token" "foo" {
```
eleijonmarck:
```suggestion
resource "grafana_cloud_stack_service_account_token" "foo" {
```

same on line 38 https://github.com/grafana/terraform-provider-grafana/pull/858/files#diff-5d331c8a790d51c4a831414f1b3c1f9c167047869816b1a42098f64be942a015R38

and line 14 https://github.com/grafana/terraform-provider-grafana/pull/858/files#diff-83c405ec104113af8b9a9cdd117769d3e7dedb8ea7373a83c86d4ca96d1816c2R14

internal/resources/cloud/resource_cloud_stack_service_account_token.go
```diff
@@ -0,0 +1,177 @@
...
+
+	name := d.Get("name").(string)
+	ttl := d.Get("seconds_to_live").(int)
+
+	request := gapi.CreateServiceAccountTokenRequest{
+		Name:             name,
+		ServiceAccountID: serviceAccountID,
+		SecondsToLive:    int64(ttl),
+	}
+	response, err := c.CreateServiceAccountToken(request)
```
eleijonmarck, line 88:
```suggestion
	response, err := c.CreateStackServiceAccountToken(request)
```

this imply that we need the client to be changed as well? 🤔

internal/resources/cloud/resource_cloud_stack_service_account_token.go
```diff
@@ -0,0 +1,177 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	id, err := strconv.ParseInt(d.Id(), 10, 32)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	_, err = c.DeleteServiceAccountToken(serviceAccountID, id)
```
eleijonmarck, line 166:
```suggestion
	_, err = c.DeleteStackServiceAccountToken(serviceAccountID, id)
```
same here

eleijonmarck (commented):
made some comments on making use of the name `stack_service_account_token` instead of `service_account_token`. It makes sense if they are different to name the resources accordingly.

We could also change it to `grafana_cloud_service_account_token`

internal/resources/cloud/resource_cloud_stack_service_account_token.go
IevaVasiljeva: **Edit**: sorry, I think I completely misunderstood your point in my first response. If you're saying the same that Julien is saying [here](https://github.com/grafana/terraform-provider-grafana/pull/858#issuecomment-1482107129), then it totally makes sense and yes, we should do it :)

examples/resources/grafana_cloud_stack_service_account_token/resource.tf
IevaVasiljeva: Thanks, fixed it.

IevaVasiljeva:
> Just a quick initial comment, we can discuss this monday. What do you think about have a single resource for the SA AND the token? My point of view is that the "raison d'être" of this resource is to create a management entrypoint into the stack.
> 
> After that management service account is created, users can use the grafana provider to create more service accounts as they wish
> 
> We could call it `grafana_cloud_stack_management_token` or something

That's an interesting thought. The main reason why I went for this implementation is that it's the same as for `grafana_service_account` resource, so should be familiar to the user. But I think your point on this service account being used for stack maintenance and hopefully nothing else makes a lot of sense. What do you think @eleijonmarck and @vtorosyan?

IevaVasiljeva:
> Also, other upper-level comment
> 
> > Related platform changes: [grafana/grafana-com#6136](https://github.com/grafana/grafana-com/pull/6136)
> 
> This is not being used AFAICT, right? To do so, we'd have to modify https://github.com/grafana/grafana-api-golang-client/blob/master/cloud_grafana_api_key.go#L28-L57
> 
> But if we think about this, it doesn't make sense to create a temp client anymore. The temporary client logic was built because the `api_key` resource was used for both cloud and grafana itself. Now that it's a separate resource for cloud, and if we have an entrypoint into instances to directly manage service accounts, we should use that instead of creating a temporary client
> 
> What do you think?

Oh yes, that totally makes sense! Thanks for pointing it out 😄 I'll extend the client to expose the new endpoints.

eleijonmarck:
> That's an interesting thought. The main reason why I went for this implementation is that it's the same as for grafana_service_account resource, so should be familiar to the user. But I think your point on this service account being used for stack maintenance and hopefully nothing else makes a lot of sense.

@IevaVasiljeva i also thought about this initially actually. 

In one way it is the current way to manage the cloud stacks and therefore we should keep it that way. It also makes the experience a bit better, to only create one resource to manage your stack. Personally, I would like to keep it more **consistent** in terms of how we initiate permissions for an instance. If we go down the route of having a SA and SAT created within one resource, I would like to discuss having that also part of the grafana provider.

The other aspect of the implementation being separate is that cloud could use this as a way to initiate None service accounts with specific permissions. So we miss the potential upside SAs could provide.

---
The resource flow could be discussed to make it easier, potentially we could add a token creation in the service account resource as they are very much linked. @julienduchesne we discussed this briefly about the team and group resources creation some time ago as well.

IevaVasiljeva:
@julienduchesne @eleijonmarck I've changed the implementation to use service accounts for the temp token instead of using the API keys: https://github.com/grafana/grafana-api-golang-client/pull/140

Unfortunately, I think we still need temp tokens, as we only have methods for creating service accounts and service account tokens through the cloud API, but not for listing them or doing anything else with them.

Secondly, about having one resource for managing service accounts and SA tokens - sounds like we all agree that it makes sense? I would be reluctant to change the `grafana_service_account` resource though, I feel like it's outside the scope of grafana_service_account` is implemented. So if you feel strongly about having consistency between `grafana_service_account` and the new resource, I'd prefer to stick with the way in which the existing resource is implemented.

I don't think I quite understand your point about SA permissions @eleijonmarck. If we had one resource, we could still assign the empty role to it? And we still need to implement a way to assign RBAC roles to `grafana_cloud_stack_service_account`, that is not covered by this work.

eleijonmarck:
> I don't think I quite understand your point about SA permissions @eleijonmarck. If we had one resource, we could still assign the empty role to it? And we still need to implement a way to assign RBAC roles to grafana_cloud_stack_service_account, that is not covered by this work.

We could still assign the empty role for sure. The point is the abstraction of a `machine user` and accomponing tokens to it. If we would create one resource, we would want a 0..n tokens being provided by it potentially. If we are are fine with a lot of these resources being created for the same purpose then I am okay with having one resource for one token.

The example I am thinking about is the use case of having one SA that has a set of specific permissions where one would want multiple tokens for it. In the one resource implementation we would create multiple of these same resources, while having a service account and tokens separate one only needs one and many tokens.

This is something that I am thinking about, but not really a concern.

I am actually more concerned with consistency than anything.

IevaVasiljeva:
I tried combining Service Accounts and Service Account tokens in one resource, here's the PoC: https://github.com/grafana/terraform-provider-grafana/compare/create-sas-through-cloud-provider-2?expand=1

I ran into some difficulties - the main problem with the above implementation is that it doesn't pick up changes to the token associated with the service account. For instance, if the token has been deleted, no changes will be registered. I'm not sure how to work around it nicely. So unless you have some suggestions, I think it might be easier to stick with having two separate resources.

eleijonmarck:
noting this down as we need to update the docs appropriately in the terraform provider and some of the places in our grafana cloud stack provisioning docs in grafana docs.

why?
**Two Resource Creation for Service Account and Token in Grafana and Grafana Cloud**

We want to have consistent behavior similar way of creating service accounts in both Grafana and Grafana Cloud, as demonstrated by the following resources:

Grafana:
grafana_service_account
grafana_service_account_token

Grafana Cloud:
grafana_cloud_stack_service_account
grafana_cloud_stack_service_account_token

via a cloud provider you can also create cloud accounts via:
grafana_service_account
grafana_service_account_token


Therefore it makes it even more consistent if we use two resources defined.

```terraform
terraform {
  required_providers {
    grafana = {
      source  = "grafana/grafana"
    }
  }
}

# Declaring the first provider to be only used for creating the cloud-stack
provider "grafana" {
  alias = "first"

  # cloud_api_key = "<API-Key>" 
}

resource "grafana_cloud_stack" "eleijonmarck" {
  provider = grafana.first

  name        = "eleijonmarck"
  slug        = "eleijonmarck"
  region_slug = "eu" # Example “us”,”eu” etc
}


##########
# First way of creating service account and token to the cloud stack
##########
resource "grafana_cloud_stack_service_account" "cloud_sa" {
  stack_slug = "<your stack slug>"
  name        = "cloud service account"
  role        = "Admin"
  is_disabled = false
}
resource "grafana_cloud_stack_service_account_token" "foo" {
  name               = "key_foo"
  service_account_id = grafana_cloud_stack_service_account.cloud_sa.id
}
output "service_account_token_foo_key" {
  value     = grafana_cloud_stack_service_account_token.foo.key
  sensitive = true
}

##########
# Second way of creating service account and token to the cloud stack
# this uses the provider and the stack managed service account token for creating service accounts
##########
# Declaring the second provider to be used for creating resources in Grafana        
provider "grafana" {
  alias = "second"

  url  = grafana_cloud_stack.eleijonmarck.url
  auth = grafana_cloud_stack_service_account_token.key
}

# Creating a service account in Grafana instance to be used as auth
resource "grafana_service_account" "eleijonmarck_service_account" {
  provider = grafana.second

  name             = "eleijonmarck-sa"
  role             = "Admin"
}

# Creating a service account token in Grafana instance to be used for creating resources in Grafana instance
resource "grafana_service_account_token" "eleijonmarck_service-account_token" {
  provider = grafana.second

  name               = "eleijonmarck-sa-token"
  role               = "Admin"
  service_account_id = grafana_service_account.eleijonmarck_service_account.id
}
```


I like this as it makes it consistent with how you might want to create service accounts in the cloud. either via already defined service account creation in some `service_account.tf` file but changing the provider.

@julienduchesne me and ieva talked that going forward it will be best to have two resources. My reasoning for consistency and documentation is outlined in this comment

eleijonmarck (approved):
thx for all the discussion! LGTM! 🥇

PR #860, opened 2023-03-24
Switch report and datasource permissions tests to Enterprise

### LINKED ISSUE #755: Fix the `TestAccDatasourcePermission_basic` test
Broken since 9.3+ in Grafana Cloud. Was disabled here: https://github.com/grafana/terraform-provider-grafana/pull/754

julienduchesne:
Closes https://github.com/grafana/terraform-provider-grafana/issues/755

They were using the cloud instance rather than the enterprise instance which we have setup in CI now 
This will lead to more predictable failures (cloud can be auto-updated)


inkel (approved):
❤️ thanks

PR #880, opened 2023-04-13
Alerting: Fix grafana_contact_point ignores changes in secure fields

### LINKED ISSUE #713: grafana_contact_point ignores changes in pagerduty integration_key
### Terraform Version
* Terraform: 1.3.4
* Terraform Grafana Provider: 1.30.0
* Grafana: 9.2.0

### Affected Resource(s)
* grafana_contact_point

```hcl
# Apply this....
resource "grafana_contact_point" "contact_point" {
  name     = "Test contact point changes"

  pagerduty {
    integration_key = "test_key_1"
  }
}

# And then change pagerduty.integration_key to any other string and run terraform apply or plan
```

### Debug Output
https://gist.github.com/selverob/fef13f6b03539bb23b22480070371354

### Panic Output
N/A

### Expected Behavior
After changing the integration key, PagerDuty receiver in Grafana Alerting should be updated with the new key.

### Actual Behavior
No changes to infrastructure planned. Nothing is changed.

### Steps to Reproduce
1. Create a `grafana_contact_point` resource with a `pagerduty` block with an `integration_key` set to any value
2. Change the value of `integration_key` in the block

JacobsonMT:
Previously we suppressed the comparison of secure fields and stored "[REDACTED]" in state. This change will instead store the unredacted values in state on create/update and patch them during read.

This way, Terraform can identify and update changed secure field values.

Fixes: #713

### DIFF
--- a/internal/resources/grafana/resource_alerting_contact_point.go
+++ b/internal/resources/grafana/resource_alerting_contact_point.go
@@ -219,6 +219,22 @@ func unpackContactPoints(data *schema.ResourceData) []gapi.ContactPoint {
 	return result
 }
 
+func unpackContactPoint(data *schema.ResourceData, uid string) gapi.ContactPoint {
+	name := data.Get("name").(string)
+	for _, n := range notifiers {
+		if points, ok := data.GetOk(n.meta().field); ok {
+			for _, p := range points.([]interface{}) {
+				pt := unpackPointConfig(n, p, name)
+				if pt.UID == uid {
+					return pt
+				}
+			}
+		}
+	}
+
+	return gapi.ContactPoint{}
+}
+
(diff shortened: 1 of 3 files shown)


julienduchesne:
Weird, I see tests failing with REDACTED values being seen, but I don't see anymore references about that string anymore in the code 🤔 

Does the API return `[REDACTED]`?

JacobsonMT:
Yes, the API returns `[REDACTED]`. There seems to be an edge case or issue that I didn't catch in local testing when I copied the code over from slack to all the other notifiers. I'm looking into it and will update.

julienduchesne (approved):
Good job. Pretty tricky considering you have to re-send the whole config every time. The tests still look good so 👍

PR #889, opened 2023-04-18
Cloud Stack: Fix 404 check

julienduchesne:
This is something that probably occurs in Crossplane. 
On the first read of the cloud stack resource, it doesn't exist so it 404s, but I broke the 404 mechanism in https://github.com/grafana/terraform-provider-grafana/pull/884


inkel (approved):
I wonder if there's a way to return a signal error instead 🤔

julienduchesne:
> I wonder if there's a way to return a signal error instead 🤔

It's not an error though, Terraform resources should detect that the underlying resource doesn't exist. It's a normal-ish part of the TF workflow

PR #897, opened 2023-04-25
Folder: Add option to prevent deletion if not empty

### LINKED ISSUE #855: folder cannot be deleted: folder contains alert rules
### Terraform Version
* Terraform: 1.3.4
* Terraform Grafana Provider: 1.34.0
* Grafana: 9.3

### Affected Resource(s)
* grafana_folder

```hcl
resource "grafana_folder" "hbg" {
  provider = grafana.org

  title = var.folder_title
  uid   = var.folder_uid
}
```

### Debug Output
`Error: status: 400, body: {"message":"folder cannot be deleted: folder contains alert rules"}`

### Expected Behavior
A Folder is removed, including it's associated Alert Rules. As explained in the [Official Documentation](https://grafana.com/docs/grafana/v9.3/developers/http_api/folder/#delete-folder):

> If [Grafana Alerting](https://grafana.com/docs/grafana/v9.3/alerting/) is enabled, you can set an optional query parameter `forceDeleteRules=false` so that requests will fail with 400 (Bad Request) error if the folder contains any Grafana alerts. However, if this parameter is set to true then it will delete any Grafana alerts under this folder.

### Actual Behavior
The Folder is not removed because it contains Alert Rules. My guess is that the [DeleteFolder function](https://github.com/grafana/terraform-provider-grafana/blob/39a329009f77839ee284a5b740d281451685e2e2/internal/resources/grafana/resource_folder.go#L141) calls the [GrafanaAPI client DeleteFolder function](https://github.com/grafana/grafana-api-golang-client/blob/618bf9712c47a57802dd5c1610f23c26da58f675/folder.go#L100), but this one is not up-to-date to what the documentation says.

However, this works, though
`curl --location --request DELETE 'http://grafana.service/api/folders/FOLDER_UID?forceDeleteRules=true'`

### Steps to Reproduce
1. `terraform apply`
2. Create a dashboard in the UI, a Panel and an Alert
3. `terraform destroy`

julienduchesne:
Closes https://github.com/grafana/terraform-provider-grafana/issues/855 
This will allow Terraform to take on folders that potentially have manually-created dashboards in them, without the risk of destroying them without warning

Also, in this condition we can handle the case of alert rules


internal/resources/grafana/resource_folder.go
```diff
@@ -138,7 +145,29 @@ func ReadFolder(ctx context.Context, d *schema.ResourceData, meta interface{}) d
...
+			var dashboardNames []string
+			for _, dashboard := range dashboards {
+				dashboardNames = append(dashboardNames, dashboard.Title)
+			}
+			return diag.Errorf("folder %s is not empty and prevent_destroy_if_not_empty is set. It contains the following dashboards: %v", d.Get("uid").(string), dashboardNames)
+		}
+	} else {
+		// If we're not preventing destroys, then we can force delete folders that have alert rules
+		deleteParams = append(deleteParams, gapi.ForceDeleteFolderRules())
+	}
```
inkel, line 168: Can we omit the `else`?
```suggestion
	}

	// If we're not preventing destroys, then we can force delete folders that have alert rules
	deleteParams = append(deleteParams, gapi.ForceDeleteFolderRules())
```

inkel (approved):
Thank you!

internal/resources/grafana/resource_folder.go
julienduchesne: We cannot. The param should only be there if `prevent_destroy_if_not_empty` is false

inkel: Oohh right, I missed the lack of `return` in the block above 🤦🏽

PR #907, opened 2023-05-02
Folder: Allow read operations from either UID or ID

julienduchesne:
Same as https://github.com/grafana/terraform-provider-grafana/pull/884 but for folders

Currently, the logic is in the importer but that doesn't work for Crossplane that doesn't use import but instead builds a custom state 
This will be functionally the same but the slug is now usable at all points in the process, instead of being translated in the importer

### DIFF
--- a/internal/resources/grafana/resource_folder.go
+++ b/internal/resources/grafana/resource_folder.go
@@ -200,6 +200,7 @@ func GetFolderByIDorUID(client *gapi.Client, id string) (*gapi.Folder, error) {
 		for _, folder := range folders {
 			if folder.ID == numericalID {
 				uid = folder.UID
+				break
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


internal/resources/grafana/resource_folder.go
```diff
@@ -208,22 +187,22 @@ func NormalizeFolderConfigJSON(configI interface{}) string {
...
+	// We need to use two API calls in the numerical ID case, because the "list" call doesn't have all the info
+	uid := id
+	if numericalID, err := strconv.ParseInt(id, 10, 64); err == nil {
+		folders, err := client.Folders()
+		if err != nil {
+			return nil, err
+		}
+		for _, folder := range folders {
+			if folder.ID == numericalID {
+				uid = folder.UID
```
inkel, line 202: Shouldn't we `break` here?

inkel (approved):

internal/resources/grafana/resource_folder.go
julienduchesne: Definitely should

PR #913, opened 2023-05-08
Adds a Global Header to all Grafana Terraform Provider requests

elainevuong:
Proposal to add a Global HTTP Custom Header to all requests coming from the Grafana Terraform Provider.

This PR would add the custom HTTP header:
`"Grafana-Terraform-Provider" : "true"`

to all requests originating from the Terraform Provider, in addition to any other `http_headers` or custom `GRAFANA_HTTP_HEADERS` environment variables that are set by the user.

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -439,7 +439,10 @@ func createOnCallClient(d *schema.ResourceData) (*onCallAPI.Client, error) {
 	return onCallAPI.New(baseURL, aToken)
 }
 
+// Sets a custom HTTP Header on all requests coming from the Grafana Terraform Provider to Grafana-Terraform-Provider: true
+// in addition to any headers set within the `http_headers` field or the `GRAFANA_HTTP_HEADERS` environment variable
 func getHTTPHeadersMap(d *schema.ResourceData) (map[string]string, error) {
+	headers := map[string]string{"Grafana-Terraform-Provider": "true"}
 	headersMap := d.Get("http_headers").(map[string]interface{})
 	if len(headersMap) == 0 {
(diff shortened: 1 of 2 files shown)


julienduchesne (approved):

internal/provider/provider.go
```diff
@@ -439,7 +439,10 @@ func createOnCallClient(d *schema.ResourceData) (*onCallAPI.Client, error) {
 	return onCallAPI.New(baseURL, aToken)
 }
 
+// Sets a custom HTTP Header on all requests coming from the Grafana Terraform Provider to Grafana-Terraform-Provider: true
+// in addition to any headers set within the `http_headers` field or the `GRAFANA_HTTP_HEADERS` environment variable
 func getHTTPHeadersMap(d *schema.ResourceData) (map[string]string, error) {
+	headers := map[string]string{"Grafana-Terraform-Provider": "true"}
```
joeblubaugh, line 445: Sorry to ask after this is already merged, but:
What do you think about making this header more flexible? Maybe something like `"Grafana-Request-Origion": "terraform"`? That lets us provide options for other tooling in the future - we could propose adding headers to grabana and also to tools Grafana provides.

I've raised an issue on grafana/grafana

joeblubaugh:
https://github.com/grafana/grafana/discussions/68049

internal/provider/provider.go
elainevuong: Open to renaming this key-value pair to something more suitable - let's see if anybody has more thoughts on the [discussion](https://github.com/grafana/grafana/discussions/68049) within grafana/grafana that you've linked out below, and I'm open to modifying this.

PR #918, opened 2023-05-23
Fix removing role from state if it does not exist

skatsaounis:
Fix the current provider state which fails to perform plan and also remove from state roles that do not exist anymore.


skatsaounis:
**Current state**

```
Error: status: 404, body: {"message":"Role not found","traceID":"123456"}
```

**Expected behavior**

Plan succeeds and also removes the deleted role from state.

**Reason**

The if condition checks for `role` rather than `Role`.

julienduchesne (approved):
Thanks

PR #923, opened 2023-06-02
Report: Add `org_id` attribute

julienduchesne:
Issue: https://github.com/grafana/terraform-provider-grafana/issues/747

Allows creating and reading reports from orgs other than default (without creating multiple `provider` instances)


inkel (approved):
Love it, thanks!

PR #928, opened 2023-06-02
`grafana_contact_point`: Force the use of at least one notifier

### LINKED ISSUE #797: Creating grafana_contact_point with no destination results in `Root resource was present, but now absent.`
### Terraform Version
* Terraform: 1.2.9
* Terraform Grafana Provider: 1.34.0
* Grafana: 9.3.2

### Affected Resource(s)
* grafana_contact_point

```hcl
resource "grafana_contact_point" "dev_null" {
  name = "empty-test"
}
```

### Debug Output

https://gist.github.com/vCra/59b7cc52d8cf278f921a45457962363c#file-grafana-contact-point-error

### Panic Output

### Expected Behavior
What should have happened?

A contact point with no destination should be created

### Actual Behavior
What actually happened?

│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to grafana_contact_point.empty_test, provider "provider[\"registry.terraform.io/grafana/grafana\"]" produced an unexpected new value: Root resource was present, but now absent.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.

### Steps to Reproduce

1. `terraform apply`

julienduchesne:
Closes https://github.com/grafana/terraform-provider-grafana/issues/797 
Build a list of all notifier names and set `AtLeastOneOf` to make sure an empty contact point isn't created


internal/resources/grafana/resource_alerting_contact_point.go
```diff
@@ -61,12 +61,19 @@ This resource requires Grafana 9.1.0 or later.
 		},
 	}
 
+	// Build list of available notifier fields, at least one has to be specified
+	notifierFields := []string{}
+	for _, n := range notifiers {
+		notifierFields = append(notifierFields, n.meta().field)
+	}
```
inkel: **Feel free to ignore this**

Low hanging perf optimization:
```suggestion
	notifierFields := make([]string{}, len(notifiers))
	for i, n := range notifiers {
		notifierFields[i] = n.meta().field
	}
```

I suppose we could even move this out into an `init` function so if you use this resource more than once we don't need to recompute this slice over and over.

inkel (approved):

PR #930, opened 2023-06-05
docs: replace cloud_api_key in favor of service accounts in examples

theSuess:
As the cloud_stack_api_key has been deprecated, it should also be replaced in favor of service accounts


julienduchesne (approved):
Thanks!

Issue #891, opened 2023-04-19
`traces_url` in grafana_cloud_stack doesn't include /tempo endpoint


mvadu:
### Terraform Version
* Terraform: 1.4.2
* Terraform Grafana Provider: 1.37.2
* Grafana: NA

The `trace_url` in `grafana_cloud_stack` resource doesn't include the `/tempo` endpoint, thus if this is used to create a datasource it fails with http 404. `prometheus_remote_endpoint` for example does contain the correct read endpoint. 

Usecase: 
create stacks on the fly and create the datasources in another grafana instance pointing back to newly created stack. Below is minimal cleaned up code showing the use case. 

``` terraform
# Declaring the first provider to be only used for creating the cloud-stack
provider "grafana" {
  alias         = "cloud"
  cloud_api_key = var.cloud_api_key
}


resource "grafana_cloud_stack" "new_stack" {
  provider    = grafana.cloud
  name        = var.stack_name
  slug        = var.stack_name
  region_slug = var.region
}


#reference to existing frontend stack
data "grafana_cloud_stack" "frontend_stack" {
  provider = grafana.cloud
  slug     = "my-frontend"
}


# Declaring the second provider to be used for creating resources in Grafana
provider "grafana" {
  alias = "frontend"
  url   = var.frontend_stack_url
  auth  = var.frontend_stack_api_key
}

# Create/Get an API key in frontend Grafana instance to be used for creating resources inside grafana ui
resource "grafana_cloud_stack_api_key" "tf_automation_frontend_dev" {
  provider   = grafana.cloud
  stack_slug = data.grafana_cloud_stack.dev_frontend_stack.slug
  name       = "automation_frontend"
  role       = "Admin"
}



resource "grafana_data_source" "metric_datasource" {
  provider = grafana.frontend
  type     = "prometheus"
  name     = "grafanacloud-${var.stack_name}-prom"
  url      = grafana_cloud_stack.new_stack.prometheus_remote_endpoint

  basic_auth_enabled  = true
  basic_auth_username = grafana_cloud_stack.new_stack.prometheus_user_id
  secure_json_data_encoded = jsonencode({
    basicAuthPassword = var.viewer_api_key
  })
}

resource "grafana_data_source" "logs_datasource" {
  provider = grafana.frontend
  type     = "loki"
  name     = "grafanacloud-${var.stack_name}-logs"
[...]

PR #932, opened 2023-06-05
Document that a stack's trace URL is the base URL

julienduchesne:
Issue: https://github.com/grafana/terraform-provider-grafana/issues/891 

I will recommend that this user open an issue to `grafana/grafana` instead
1. If the `loki` datasource supports a bare URL and not the tempo datasource, then it seems like a feature gap
2. We can't change the traces URL now, it will break existing setups
3. I think it's better to reflect what the API sends us


inkel (approved):
I agree with you, we shouldn't change this, and in any case, the issue is in `grafana/grafana` and not the provider.

PR #933, opened 2023-06-05
Alerting: Update Discord contact point to treat URL as a secure field

yuri-tceretian:
The field `url` in the Discord contact point declaration is declared as sensitive but in reality, it is treated as a regular setting. This PR updates the Discord contact point to treat `url` field as a secure field. 

This is part of the fix for https://github.com/grafana/grafana/issues/40148. 

### Testing
I tested it against the current version of Grafana which treats the URL field as a regular setting, and it does not change the behavior. Then I upgraded to https://github.com/grafana/grafana/pull/69588 and included https://github.com/grafana/alerting/pull/90 and checked that the `terraform plan` still has 0 changes. Then I changed one field (can be URL or any other), applied the plan and verified that change went through and the follow-up plan has 0 changes.

### DIFF
--- a/internal/resources/grafana/resource_alerting_contact_point_notifiers.go
+++ b/internal/resources/grafana/resource_alerting_contact_point_notifiers.go
@@ -174,9 +174,10 @@ var _ notifier = (*discordNotifier)(nil)
 
 func (d discordNotifier) meta() notifierMeta {
 	return notifierMeta{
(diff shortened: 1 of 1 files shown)


julienduchesne (approved):
LGTM!

alexweav (approved):
LGTM

PR #939, opened 2023-06-10
Ensure Alerting Rules Not Generated for Empty Alerting Fields on TF State File

### LINKED ISSUE #938: SLO: Empty Alerting Block on Terraform State File Generating Alerting Rules
Empty Alerting Blocks on the Terraform State File should NOT generate alerting rules.

elainevuong:
Updates Terraform Provider to ensure Alerting rules are not generated for instances where the Alerting field on the Terraform state file is not specified by the user

This PR addresses the three scenarios on the state of the Alerting Field within a Terraform State file:

- if the Alerting Block within the Terraform State file is populated, then the appropriate recording/alerting/labels are generated
- if the Alerting Block within the Terraform State file is non-existent, then no alerting rules are generated
- if the Alerting Block within the Terraform State file is an empty block (`{}`) then the default alerting rules are generated


internal/resources/slo/resource_slo.go
```diff
@@ -298,19 +298,42 @@ func packSloResource(d *schema.ResourceData) (gapi.Slo, error) {
 		tflabels = packLabels(labels)
 	}
 
-	alerting := d.Get("alerting").([]interface{})
-	if len(alerting) > 0 {
-		alert := alerting[0].(map[string]interface{})
-		tfalerting = packAlerting(alert)
+	alerting, alertingOK := d.GetOk("alerting")
```
joeblubaugh: It's more idiomatic to collapse this into the conditional - you don't need to worry about alerting, and alertingOK "contaminating" the scope outside of the block in that case:
```go
if alerting, ok := d.GetOk("alerting"); ok {
  ...
}

// alerting is no longer declared here, compile-time error:
alerting = nil
```

internal/resources/slo/resource_slo.go
```diff
@@ -298,19 +298,42 @@ func packSloResource(d *schema.ResourceData) (gapi.Slo, error) {
...
+		alertData := alerting.([]interface{})
+
+		// if the Alerting field is an empty block, alertData[0] has a value of nil
+		if alertData[0] != nil {
+			// only pack the Alerting TF fields if the user populates the Alerting field with blocks
+			alert := alertData[0].(map[string]interface{})
+			tfalerting = packAlerting(alert)
+		}
+
+		slo := gapi.Slo{
```
joeblubaugh: Instead of an early return here, consider just modifying the Slo object and having a single return for a no- error execution. I think it reduces the mental burden of modifying it later:
```go
slo := gapi.Slo{
	UUID:        d.Id(),
	Name:        tfname,
	Objectives:  tfobjective,
	Query:       tfquery,
	Alerting:    nil,
	Labels:      tflabels,
}

if alerting, ok := d.GetOk("alerting"); ok {
  ...
  tfAlerting := packAlerting(alert)
  slo.Alerting = &tfAlerting
}

return slo, nil
```

joeblubaugh (commented):
I "approve" this - I'll ask one of the provider maintainers to also give this a look today; since it's my first review of this codebase I want to have an expert review my review.

>  *    if the Alerting Block within the Terraform State file is populated, then the appropriate recording/alerting/labels are generated
> *   if the Alerting Block within the Terraform State file is non-existent, then no alerting rules are generated
> * if the Alerting Block within the Terraform State file is an empty block ({}) then the default alerting rules are generated

👍

julienduchesne (approved):
LGTM. I echo @joeblubaugh's comments

internal/resources/slo/resource_slo.go
inkel: Agreed with ☝🏽

internal/resources/slo/resource_slo.go
inkel: Same here.

inkel (approved):
LGTM but if you can address @joeblubaugh comments it will look much better 😬

PR #966, opened 2023-07-05
Remove unused fields in `synthetic_monitoring_installation`

julienduchesne:


internal/resources/cloud/resource_synthetic_monitoring_installation.go
```diff
@@ -44,7 +44,6 @@ This resource cannot be imported but it can be used on an existing Synthetic Mon
 `,
 		CreateContext: ResourceInstallationCreate,
 		ReadContext:   ResourceInstallationRead,
-		UpdateContext: ResourceInstallationRead,
```
julienduchesne, line 47: Since all fields are `ForceNew`, the TF framework told me to drop the update func

inkel: Makes sense.

inkel (approved):

PR #976, opened 2023-07-12
Datasource: Add `org_id` attribute

julienduchesne:
Issue: #747
Allows a datasource to be created and managed in a separate organization without needing two provider blocks 
Also adds tests that a datasource can be created and managed in an org

### DIFF
--- a/internal/resources/grafana/resource_data_source_permission.go
+++ b/internal/resources/grafana/resource_data_source_permission.go
@@ -12,7 +12,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 
 	gapi "github.com/grafana/grafana-api-golang-client"
-	"github.com/grafana/terraform-provider-grafana/internal/common"
 )
 
 func ResourceDatasourcePermission() *schema.Resource {
@@ -28,8 +27,9 @@ func ResourceDatasourcePermission() *schema.Resource {
 		DeleteContext: DeleteDatasourcePermissions,
 
 		Schema: map[string]*schema.Schema{
+			"org_id": orgIDAttribute(),
 			"datasource_id": {
-				Type:        schema.TypeInt,
+				Type:        schema.TypeString,
 				Required:    true,
 				ForceNew:    true,
 				Description: "ID of the datasource to apply permissions to.",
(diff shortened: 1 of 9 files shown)


internal/resources/grafana/resource_data_source_test.go
```diff
@@ -277,12 +312,22 @@ func testAccDataSourceCheckExists(rn string, dataSource *gapi.DataSource) resour
...
+		id, err := strconv.ParseInt(idStr, 10, 64)
 		if err != nil {
 			return fmt.Errorf("resource id is malformed")
 		}
 
 		client := testutils.Provider.Meta().(*common.Client).GrafanaAPI
+		// If the org ID is set, check that the datasource doesn't exist in the default org
+		if orgID > 1 {
+			datasource, err := client.DataSource(id)
+			if err == nil || datasource != nil {
```
inkel, line 325: Could it ever happen that both `err` and `datasource` are nil?

internal/resources/grafana/resource_data_source_permission.go
```diff
@@ -28,8 +27,9 @@ func ResourceDatasourcePermission() *schema.Resource {
 		DeleteContext: DeleteDatasourcePermissions,
 
 		Schema: map[string]*schema.Schema{
+			"org_id": orgIDAttribute(),
 			"datasource_id": {
-				Type:        schema.TypeInt,
+				Type:        schema.TypeString,
```
inkel, line 32: Why changing this to a string, when it seems all of the uses below treat it as an int?

inkel (commented):
In general looks good, but I've left some questions for things that I'm not understanding :yay:

internal/resources/grafana/resource_data_source_permission.go
julienduchesne: Since the datasource now has an id in the `org_id:id` format, it's a string. So, even if we're only using the second part, the ref has to be a string. 

It's not an issue to switch types because terraform does seamless conversions between int and string

internal/resources/grafana/resource_data_source_test.go
julienduchesne: I don't think so but it'll throw an error so we'll definitely see something is wrong

internal/resources/grafana/resource_data_source_permission.go
julienduchesne: One point I might add is that I'm aiming to also deprecate all of the `permissions` resources. They don't really make sense as standalones, the biggest sign that something is wrong is that the delete function doesn't delete anything, it clears out the permissions in the datasource

inkel (approved):
:shipit:

PR #977, opened 2023-07-13
Refactor: Common read error check function

julienduchesne:
In each OSS and enterprise resources, we have the same pattern to check for 404 
We can have a common function for that
Also, added that check to resources that were missing it

### DIFF
--- a/internal/common/errcheck.go
+++ b/internal/common/errcheck.go
@@ -16,19 +16,21 @@ const NotFoundError = "status: 404"
 // - If there is an error, return the error.
 // - Otherwise, do not return to continue processing the read.
 func CheckReadError(resourceType string, d *schema.ResourceData, err error) (returnValue diag.Diagnostics, shouldReturn bool) {
-	if err != nil {
-		if strings.Contains(err.Error(), NotFoundError) {
-			log.Printf("[WARN] removing %s with ID %q from state because it no longer exists in grafana", resourceType, d.Id())
-			var diags diag.Diagnostics
-			diags = append(diags, diag.Diagnostic{
-				Severity: diag.Warning,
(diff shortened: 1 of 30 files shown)


internal/common/errcheck.go
```diff
@@ -0,0 +1,34 @@
...
+				Severity: diag.Warning,
+				Summary:  fmt.Sprintf("%s with ID %q is in Terraform state, but no longer exists in Grafana", resourceType, d.Id()),
+				Detail:   fmt.Sprintf("%q will be recreated when you apply", d.Id()),
+			})
+			d.SetId("")
+			return diags, true
+		}
+		return diag.Errorf("error reading %s with ID`%s`: %v", resourceType, d.Id(), err), true
+	}
+	return nil, false
```
inkel: How about inverting the first `if` to reduce indentation?
```suggestion
	if err == nil {
		return nil, false
	}
	if strings.Contains(err.Error(), NotFoundError) {
		log.Printf("[WARN] removing %s with ID %q from state because it no longer exists in grafana", resourceType, d.Id())
		var diags diag.Diagnostics
		diags = append(diags, diag.Diagnostic{
			Severity: diag.Warning,
			Summary:  fmt.Sprintf("%s with ID %q is in Terraform state, but no longer exists in Grafana", resourceType, d.Id()),
			Detail:   fmt.Sprintf("%q will be recreated when you apply", d.Id()),
		})
		d.SetId("")
		return diags, true
	}
	
	return diag.Errorf("error reading %s with ID`%s`: %v", resourceType, d.Id(), err), true
	}
```

We could probably do something similar for the second `if strings.Contains` and reduce it even further.

inkel (approved):
This is great, thanks! Just left a small suggestion, but definitely not blocking from getting this merged.

internal/common/errcheck.go
julienduchesne: Good idea. Done!

PR #979, opened 2023-07-13
Team: Support `org_id` attribute

### LINKED ISSUE #747: Support Setting of org_id outside of provider block

julienduchesne:
Closes https://github.com/grafana/terraform-provider-grafana/issues/747
Final resource to receive the `org_id`
Allows a team to be created and managed in a separate organization without needing two provider blocks
Also adds tests that a team can be created and managed in an org

Note: `*_permissions` are a bit messy. I'm planning on refactoring them afterwards, they mostly follow the same pattern so it should be possible to make them much nicer

### DIFF
--- a/internal/resources/grafana/resource_dashboard_permission.go
+++ b/internal/resources/grafana/resource_dashboard_permission.go
@@ -29,6 +29,7 @@ func ResourceDashboardPermission() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
+			"org_id": orgIDAttribute(),
 			"dashboard_id": {
 				Type:         schema.TypeInt,
 				ForceNew:     true,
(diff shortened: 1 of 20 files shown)


internal/resources/grafana/resource_dashboard_permission.go
```diff
@@ -50,6 +51,12 @@ func ResourceDashboardPermission() *schema.Resource {
 				Type:        schema.TypeSet,
 				Required:    true,
 				Description: "The permission items to add/update. Items that are omitted from the list will be removed.",
+				// Ignore the org ID of the team when hashing. It works with or without it.
```
inkel, line 54: If it works with or without it, why ignoring it then? Not sure I follow 🤔

inkel (commented):
It looks good to me, but I have one question that I'd like to understand before moving on.

internal/resources/grafana/resource_dashboard_permission.go
julienduchesne: This could be explained better in the comment. The reason why the org ID can be ignored is that resources that refer to each other have to be in the same org anyways (except global resources like a user). So if you specify a team that is not within the same org as the permissions, the API will fail.

And the reason why it HAS to be ignored with the set function is that `1:1` (AKA: team 1 in org 1) is functionally equivalent to team `1` (that is `team 1` in an unspecified org). Lots of this is due to migration, and can definitely be improved once we deprecate the global org setting. At that point, all of the team IDs should always have the org ID in them and we can stop doing weird exceptions like this

inkel (approved):

PR #984, opened 2023-07-14
Silence `org_id` deprecation warning when not using it

julienduchesne:
The default value is making Terraform complain even if `org_id` isn't explicitly set


inkel (approved):
Good catch

PR #990, opened 2023-07-24
SLO Ratio Query 

### LINKED ISSUE #983: SLO: Add Ratio Query Type
- Add the Ratio Query type to the Grafana Terraform Provider, requires modifying the HCL Schema for both the resource and data source
- Update documentation and examples to use Ratio Query Type

elainevuong:
- adds the SLO Ratio Query type to the Grafana Terraform Provider modifying the HCL Schema for both the resource and data source
- updates documentation and examples to use Ratio Query Type

Closes #983


elainevuong:
Quick video for demo purposes attached. 

https://github.com/grafana/terraform-provider-grafana/assets/73451363/a60acde9-e1a4-444f-88b8-eed6ec7f70b9

docs/data-sources/slos.md
```diff
@@ -199,3 +200,13 @@ Read-Only:
 Read-Only:
 
 - `query` (String)
+
+
+<a id="nestedobjatt--slos--query--ratio"></a>
+### Nested Schema for `slos.query.ratio`
+
+Read-Only:
```
ellisda, line 208: nit - Why does this say read-only? These are fields that user can edit in their terraform right?

I'm probably missing something, because a bunch of these sections include "read only" - forgive me if I'm unclear whether we expect others to consume this or the schema as documentation

docs/data-sources/slos.md
```diff
@@ -199,3 +200,13 @@ Read-Only:
...
+
+- `group_by_labels` (List of String)
```
ellisda, line 210: I this a good place to mention it's optional?

docs/resources/slo.md
```diff
@@ -34,29 +34,29 @@ resource "grafana_slo" "test" {
...
-    value = "value"
+    key   = "slo"
+    value = "terraform"
   }
   alerting {
     fastburn {
       annotation {
         key   = "name"
-        value = "Critical - SLO Burn Rate Alert"
+        value = "SLO Burn Rate Very High"
```
ellisda, line 44: nit - why change?  - does this match changes we made in alert rule generation or something?

docs/resources/slo.md
```diff
@@ -159,17 +156,34 @@ Required:
 
 Required:
 
-- `freeform` (Block List, Min: 1, Max: 1) (see [below for nested schema](#nestedblock--query--freeform))
 - `type` (String) Query type must be one of: "freeform", "query", "ratio", or "threshold"
 
+Optional:
```
ellisda, line 161: Instead of "Optional" you could say "must include exactly one of the following (based on type above)" ??

docs/resources/slo.md
```diff
@@ -159,17 +156,34 @@ Required:
...
+### Nested Schema for `query.ratio`
+
+Required:
+
+- `success_metric` (String) Metric for success events (numerator)
+- `total_metric` (String) Metric for total events (denominator)
+
+Optional:
+
+- `group_by_labels` (List of String) Defines Group By Labels used for per-label alerting. Also appear as variables on SLO dashboards to enable filtering and aggregation
```
ellisda: you could mention that it must adhere to prometheus label name schema (i.e. we'll reject creation if they try to use fancy chars)

see comment on https://github.com/grafana/slo/blob/558e3c81683d907b5f36accc81eaed086442d3d2/schemas/slo_api.cue#L136 for regex and possible wording

examples/resources/grafana_slo/resource_ratio.tf
```diff
@@ -0,0 +1,43 @@
...
+    value = "terraform"
+  }
+  alerting {
+    fastburn {
+      annotation {
+        key   = "name"
+        value = "SLO Burn Rate Very High"
+      }
+      annotation {
+        key   = "name"
```
ellisda: shouldn't this be `key = "description", so that the generated alert has both a name, and a description annotation?

examples/resources/grafana_slo/resource_ratio.tf
```diff
@@ -0,0 +1,43 @@
...
+      }
+    }
+
+    slowburn {
+      annotation {
+        key   = "name"
+        value = "SLO Burn Rate High"
+      }
+      annotation {
+        key   = "name"
```
ellisda: `s/name/description/g` here too?

examples/resources/grafana_slo/resource_update.tf
```diff
@@ -12,29 +12,28 @@ resource "grafana_slo" "update" {
...
     fastburn {
       annotation {
         key   = "name"
-        value = "Critical - SLO Burn Rate Alert"
+        value = "SLO Burn Rate Very High"
       }
-      label {
-        key   = "type"
-        value = "slo"
+      annotation {
```
ellisda, line 24: I see that you're adding a second annotation (you want `description` here I think, not `name` again), but why did you remove the label as well?

Most alerts will have name and description annotations (helps with notifications), but we also allow labels. Why not show both?

examples/resources/grafana_slo/resource_update.tf
```diff
@@ -12,29 +12,28 @@ resource "grafana_slo" "update" {
...
       annotation {
         key   = "name"
-        value = "Warning - SLO Burn Rate Alert"
+        value = "SLO Burn Rate High"
       }
-      label {
-        key   = "type"
-        value = "slo"
+      annotation {
+        key   = "name"
```
ellisda: `s/name/description/g` here

internal/resources/slo/data_source_slo.go
```diff
@@ -99,6 +100,22 @@ func unpackQuery(apiquery gapi.Query) []map[string]interface{} {
...
+			"total_metric":    apiquery.Ratio.TotalMetric.PrometheusMetric,
+			"group_by_labels": apiquery.Ratio.GroupByLabels,
+		}
+
+		ratio = append(ratio, body)
+		query["ratio"] = ratio
+
+		retQuery = append(retQuery, query)
+	}
+
```
ellisda, line 118: do you need error handling for case where `apiQuery.Type` didn't match either of the if blocks?

internal/resources/slo/data_source_slo.go
```diff
@@ -99,6 +100,22 @@ func unpackQuery(apiquery gapi.Query) []map[string]interface{} {
 		retQuery = append(retQuery, query)
 	}
 
+	if apiquery.Type == gapi.QueryTypeRatio {
```
ellisda, line 103: nit - it'd be more idomatic in golang to use a `switch apiQuery.Type {` with two case statements (and a default if we know how to handle the case where it didn't match either). You could use `if {} else if {}` as well, but the switch is even clearer to reader that only one of the blocks is going to run

internal/resources/slo/resource_slo.go
```diff
@@ -57,17 +57,44 @@ Resource manages Grafana SLOs.
...
 										Type:        schema.TypeString,
-										Optional:    true,
+										Required:    true,
 										Description: "Freeform Query Field",
 									},
 								},
 							},
 						},
+						"ratio": &schema.Schema{
+							Type:     schema.TypeList,
```
ellisda, line 72: is this a list just because ratio contains more than one field? I'm a HCL noob, just curious. I almost thought it meant a list of ratio definitions.

internal/resources/slo/resource_slo.go
```diff
@@ -338,6 +365,40 @@ func packQuery(query map[string]interface{}) (gapi.Query, error) {
 		return sloQuery, nil
 	}
 
+	if query["type"] == "ratio" {
+		ratioquery := query["ratio"].([]interface{})[0].(map[string]interface{})
```
ellisda, line 369: is this type assertion safe? It would panic if `query["ratio"]` held something besides a list...

internal/resources/slo/resource_slo.go
```diff
@@ -338,6 +365,40 @@ func packQuery(query map[string]interface{}) (gapi.Query, error) {
...
+		successMetric := ratioquery["success_metric"].(string)
+		totalMetric := ratioquery["total_metric"].(string)
+		groupByLabels := ratioquery["group_by_labels"].([]interface{})
+
+		var labels []string
+
+		for ind := range groupByLabels {
+			if groupByLabels[ind] == nil {
```
ellisda, line 377: you might as well drop any empty-string labels (i.e. length < 1). The backend will do the same, but only cause the FE had a similar bug. We might in the future,  start rejecting SLOs because empty-string isn't a valid prometheus label name

ellisda: not a big deal, but would be better not to send emptystrings in the list

internal/resources/slo/resource_slo.go
```diff
@@ -57,17 +57,44 @@ Resource manages Grafana SLOs.
...
+										Required:    true,
+									},
+									"total_metric": &schema.Schema{
+										Type:        schema.TypeString,
+										Description: `Metric for total events (denominator)`,
+										Required:    true,
+									},
+									"group_by_labels": &schema.Schema{
+										Type:        schema.TypeList,
+										Description: `Defines Group By Labels used for per-label alerting. Also appear as variables on SLO dashboards to enable filtering and aggregation `,
```
ellisda: if this is the place end users will find documenation, then we should mention here that GroupByLabels are optional, but need to confirm to prometheus label name syntax 

https://github.com/grafana/slo/blob/558e3c81683d907b5f36accc81eaed086442d3d2/schemas/slo_api.cue#L135

internal/resources/slo/resource_slo.go
```diff
@@ -57,17 +57,44 @@ Resource manages Grafana SLOs.
...
 						},
+						"ratio": &schema.Schema{
+							Type:     schema.TypeList,
+							MaxItems: 1,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"success_metric": &schema.Schema{
+										Type:        schema.TypeString,
+										Description: `Metric for success events (numerator)`,
```
ellisda: nit - you might say "counter metric" to help clarify that we don't want them to add "rate", etc. They'll figure it out from example, but...

ellisda (approved):
LGTM. A made a couple of nit comments for you to consider, but no changes really needed.

I was hoping that schema (and/or the .md?? -- which will end users see as docs??) could follow the comments we've made in https://github.com/grafana/slo/blob/558e3c81683d907b5f36accc81eaed086442d3d2/schemas/slo_api.cue#L41... but :cry:  we don't have many comments there.

and while it should match https://github.com/grafana/slo/blob/4a46b5ccc4a7ced0894e2f1d43f7a7d4100d62f9/openapi/slo.yaml#L105 you're ahead of that, too :- )

docs/data-sources/slos.md
elainevuong: So this is the README for the SLOs Datasource - basically Terraform has two types of resources:
resources and datasources.
Datasources allow Terraform to read objects in - so that's why all the fields within this datasource markdown are read-only fields.

docs/resources/slo.md
elainevuong: Yep it does - I modified the key-value pairs within our annotations to match the default values that come out when the user clicks "Add Alert Rules" from the UI, just to make it crystal clear. Also if users decide to start with our documentation as a place to make SLOs with Terraform, it should create a SLO in a very similar format as if they had used the UI.

internal/resources/slo/resource_slo.go
elainevuong: Yeah it's a TypeList because I wanted the nesting, but the list has a restriction of being only a length of one (hence the line `MaxItems: 1`. And then that singular item is a struct consisting of `success_metric`, `total_metric`, and `group_by_labels`

internal/resources/slo/resource_slo.go
elainevuong: We are expecting value within the `query["ratio"]` field to be in a list format, I think this is fine.

internal/resources/slo/data_source_slo.go
elainevuong: This is on the READ Path from our SLO API back to Terraform (not from the User Input into Terraform) so we should always have the apiQuery type being either Ratio or Freeform. 

The user also gets a validation error if they input anything other than freeform or ratio within the type on the terraform state file as well.

PR #991, opened 2023-07-27
Delete gracefully if user no longer exist in grafana

pascal-hofmann:
When applying a terraform plan that was created with `-refresh=false` the detection of resources that no longer exist in grafana does not work, as the resources are not read at all.

Thus on apply, a deletion of a user that no longer exists in grafana fails with:
```
Error: status: 404, body: {"message":"user not found"}
```

This change modifies `DeleteUser` so that it catches this case and does not return an error.

### DIFF
--- a/internal/resources/grafana/resource_user.go
+++ b/internal/resources/grafana/resource_user.go
@@ -3,6 +3,7 @@ package grafana
 import (
 	"context"
 	"strconv"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


julienduchesne (approved):

pascal-hofmann:
Thank you for the quick review and merge! 🚀👍

PR #992, opened 2023-07-28
add multihttp check type

rdubrock:
Synthetics has a new check type that needs support in the provider


internal/resources/syntheticmonitoring/resource_check.go
```diff
@@ -752,6 +924,14 @@ func ResourceCheckRead(ctx context.Context, d *schema.ResourceData, meta interfa
 		settings.Add(map[string]interface{}{
 			"traceroute": traceroute,
 		})
+	case chk.Settings.Multihttp != nil:
+		multiHttp := schema.NewSet(
```
rdubrock: I'm pretty sure something more than this has to happen here, but I'm not sure what

internal/resources/syntheticmonitoring/resource_check.go
```diff
@@ -961,6 +1141,95 @@ func makeCheckSettings(settings map[string]interface{}) sm.CheckSettings {
 		}
 	}
 
+	multihttp := settings["multihttp"].(*schema.Set).List()
+	if len(traceroute) > 0 {
+		m := multihttp[0].(map[string]interface{})
+
+		cs.Multihttp = &sm.MultiHttpSettings{}
+
+		if m["entries"].(*schema.Set).Len() > 0 {
```
rdubrock: This whole section might be overkill? I'm not really grokking how this exactly works

docs/resources/synthetic_monitoring_check.md
```diff
@@ -559,6 +561,96 @@ Optional:
...
+
+Required:
+
+- `type` (Number) The type of assertion to make. TEXT = 0, JSON_PATH_VALUE = 1, JSON_PATH_ASSERTION = 2, REGEX_ASSERTION = 3
+
+Optional:
+
+- `condition` (Number) The condition of the assertion. NOT_CONTAINS = 1, EQUALS = 2, STARTS_WITH = 3, ENDS_WITH = 4, TYPE_OF = 5, CONTAINS = 6
+- `expression` (String) The expression of the assertion. Should start with $.
+- `subject` (Number) The subject of the assertion. RESPONSE_HEADERS = 1, HTTP_STATUS_CODE = 2, RESPONSE_BODY = 3
```
mem: I think we can do better here. Let me take a look.

docs/resources/synthetic_monitoring_check.md
```diff
@@ -559,6 +561,96 @@ Optional:
...
+### Nested Schema for `settings.multihttp.entries.checks`
+
+Required:
+
+- `type` (Number) The type of assertion to make. TEXT = 0, JSON_PATH_VALUE = 1, JSON_PATH_ASSERTION = 2, REGEX_ASSERTION = 3
+
+Optional:
+
+- `condition` (Number) The condition of the assertion. NOT_CONTAINS = 1, EQUALS = 2, STARTS_WITH = 3, ENDS_WITH = 4, TYPE_OF = 5, CONTAINS = 6
+- `expression` (String) The expression of the assertion. Should start with $.
```
rdubrock, line 590: Does the expression actually have to start with $? I'm not sure it does

mem: I'm not sure why it says it should start with `$`...

docs/resources/synthetic_monitoring_check.md
mem: I think it's good now.

_I_ was very confused with the numbers, and I am supposed to know what's going on.

docs/resources/synthetic_monitoring_check.md
mem: Ah, I see what you mean now.

For JSON_PATH_ASSERTION and JSON_PATH_VALUE, the JSONPath starts with `$.` because that's the root of the JSON object.

rdubrock: The thing is, it works without `$.` , and I think maybe it won't work with `$.`

Also fun fact \$ does really interesting things to github markdown

mem (approved):

Issue #1016, opened 2023-08-30
Configure retryTimeout through the provider configuration


stefanoboriero:
Currently it's only possible to configure the maximum number of retries for the underlying client before giving up. It would be great to be able to configure the wait time between retries, which is already supported as a fixed wait time by the underlying client https://github.com/grafana/grafana-api-golang-client/blob/master/client.go#L44. This would be useful in cases where the default fallback to 5s in the client is not enough for the API to accept throttled requests

PR #1017, opened 2023-08-30
Configure retry_wait parameter through the provider configuration

stefanoboriero:
Allow configuring wait time between retries for the underlying client by creating a new optional configuration option for the provider.

Address issue #1016


julienduchesne (approved):
LGTM!

PR #1022, opened 2023-09-05
Add retry_status_codes to Cloud client

monodot:
The cloud client object doesn't have the config property `retry_status_codes` attached to it when it's initialised, so the client seems to be retrying only on the default error codes, which are 429 and 5xx.

This PR adds the retry_status_codes value to the cloud client object.


julienduchesne:
Discussed internally. Won't do this one

Issue #1023, opened 2023-09-07
grafana_data_source import documentation error


jtama:
### Terraform Version
N/A

### Affected Resource(s)
* grafana_datasource

### Expected Behavior
It would be nice to document how to path organization identifier

### Actual documentation
The documentation says the following import syntax are available : 
```
terraform import grafana_data_source.by_integer_id {{datasource id}}
terraform import grafana_data_source.by_uid {{datasource uid}}
```

klemen-df:
Could you be even more clear, what am I missing?

`terraform import module.grafana_datasources.grafana_data_source.azure_monitor.by_uid {{1:N0KbQRynk}}`

julienduchesne:
No brackets. Should be `terraform import module.grafana_datasources.grafana_data_source.azure_monitor.by_uid 1:N0KbQRynk`

jtama:
In fact almost all import should specify this format

jtama:
@julienduchesne there was alrady a MR for this bug https://github.com/grafana/terraform-provider-grafana/pull/1024. 

Aren't contributions welcome ?

klemen-df:
> No brackets. Should be `terraform import module.grafana_datasources.grafana_data_source.azure_monitor.by_uid 1:N0KbQRynk`

Using terraform 1.2.9, grafana provider 2.3.0

It seems that I don't understand the logic, running this above I get:

```
╷
│ Error: Invalid address
│
│   on <import-address> line 1:
│    1: module.grafana_datasources.grafana_data_source.azure_monitor.by_uid
│
│ Resource instance key must be given in square brackets.
╵

For information on valid syntax, see:
https://www.terraform.io/docs/cli/state/resource-addressing.html
```

julienduchesne:
> It seems that I don't understand the logic, running this above I get:

It tells you that the error is at `module.grafana_datasources.grafana_data_source.azure_monitor.by_uid`. This address is invalid. I don't really understand why. The ID at the end `1:N0KbQRynk` never even comes into play because Terraform fails before going into the provider

klemen-df:
> > It seems that I don't understand the logic, running this above I get:
> 
> It tells you that the error is at `module.grafana_datasources.grafana_data_source.azure_monitor.by_uid`. This address is invalid. I don't really understand why. The ID at the end `1:N0KbQRynk` never even comes into play because Terraform fails before going into the provider

But, I can confirm that this works:
[...]

PR #1024, opened 2023-09-07
Update data_source.md

jtama:
Specify id format for datasource import

https://github.com/grafana/terraform-provider-grafana/issues/1023

### DIFF
--- a/examples/resources/grafana_data_source/import.sh
+++ b/examples/resources/grafana_data_source/import.sh
@@ -1,2 +1,5 @@
-terraform import grafana_data_source.by_integer_id {{datasource id}}
-terraform import grafana_data_source.by_uid {{datasource uid}}
+terraform import grafana_data_source.by_integer_id {{datasource_id}} # To use the default provider org
+terraform import grafana_data_source.by_uid {{datasource_uid}} # To use the default provider org
+
+terraform import grafana_data_source.by_integer_id {{org_id}}:{{datasource_id}} # When "org_id" is set on the resource
+terraform import grafana_data_source.by_uid {{org_id}}:{{datasource_uid}} # When "org_id" is set on the resource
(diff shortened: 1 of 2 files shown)


klemen-df:
Could you be even more clear, what am I missing?

`terraform import module.grafana_datasources.grafana_data_source.azure_monitor.by_uid {{1:N0KbQRynk}}`

examples/resources/grafana_data_source/import.sh
```diff
@@ -1,2 +1,2 @@
-terraform import grafana_data_source.by_integer_id {{datasource id}}
-terraform import grafana_data_source.by_uid {{datasource uid}}
+terraform import grafana_data_source.by_integer_id {{datasource org_id:id}}
```
julienduchesne: Can you mention both cases (with org id and without it). The same way we do here: https://github.com/grafana/terraform-provider-grafana/blob/master/docs/resources/dashboard.md#import

examples/resources/grafana_data_source/import.sh
```diff
@@ -1,2 +1,2 @@
...
+terraform import grafana_data_source.by_uid {{datasource org_id:uid}}
```
julienduchesne:
```suggestion
terraform import grafana_data_source.by_integer_id "{{org_id}}:{{datasource_id}}"
terraform import grafana_data_source.by_uid "{{org_id}}:{{datasource_uid}}"
```

It should look somewhat like templating IMO, otherwise the brackets are confusing

examples/resources/grafana_data_source/import.sh
jtama: Like that ?

examples/resources/grafana_data_source/import.sh
jtama: Done

julienduchesne (approved):
LGTM, thanks!

PR #1050, opened 2023-09-21
feat: add oncall contact point type

theSuess:
See the example for a complete configuration in combination with the oncall resources


julienduchesne (commented):
Tests fail with:
```
Error: status: 400, body: {"message":"invalid object specification: failed to validate integration \"Receiver Types\" (UID ) of type \"oncall\": notifier oncall is not supported","traceID":""}
```

theSuess:
Hm, do the tests run against an instance with the OnCall plugin installed?

I've tested against grafana cloud and it works there

theSuess:
implemented in #1087

Issue #512, opened 2022-06-03
HTTP/2 not working for `grafana_synthetic_monitoring_check`


justinTM:
### Terraform Version
* Terraform: 1.0.0
* Terraform Grafana Provider: v1.16.0
* Grafana: v8.5.4 (78ee95814)

### Affected Resource(s)
Please list the resources as a list, for example:
* `grafana_synthetic_monitoring_check`

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```yaml
# config.yml
# default values to apply for each check
check_defaults:
  probes:
    - Atlanta
    - Dallas
    - NorthVirginia
    - Oregon
    # - Amsterdam
    # - Atlanta
    # - Bangalore
    # - CapeTown
    # - Chicago
    # - Frankfurt
    # - London
    # - LosAngeles
    # - Miami
    # - Mumbai
    # - NewYork
    # - Newark
    # - NorthCalifornia
    # - NorthVirginia
    # - Ohio
    # - Paris
    # - SanFrancisco
    # - SanJose
    # - SaoPaulo
    # - Seattle
    # - Seol
    # - Seoul
    # - Singapore
    # - Sydney
    # - Tokyo
    # - Toronto
  enabled: true
  allowed_types:
    - http
    - ping
    - dns
    - tcp
  # Settings for HTTP check. The target must be a URL (http or https).
  type: http
  http:
    # Token for use with bearer authorization header.
    bearer_token: ""
    # The body of the HTTP request used in probe.
    body: ""
    # The name of the query parameter used to prevent the server from using a cached response. Each probe will assign a random value to this parameter each time a request is made.
    cache_busting_query_param_name: ""
    # List of regexes. If any match the response body, the check will fail.
    fail_if_body_matches_regexp: []
    # List of regexes. If any do not match the response body, the check will fail.
    fail_if_body_not_matches_regexp: []
    # Fail if SSL is present. Defaults to false.
    fail_if_ssl: false
    # Fail if SSL is not present. Defaults to false.
    fail_if_not_ssl: false
    # The HTTP headers set for the probe
    headers: []
    # Options are V4, V6, Any. Specifies whether the corresponding check will be performed using IPv4 or IPv6. The Any value indicates that IPv6 should be used, falling back to IPv4 if that's not available. Defaults to V4.
    ip_version: V4
[...]

PR #1059, opened 2023-09-28
Fix Synthetic monitoring http versions documentation to HTTP/2.0

mbarrien:
For the `valid_http_versions` field in  `grafana_synthetic_monitoring_checks`, passing a value of `HTTP/2` breaks any checks that actually return HTTP/2, since the Grafana check actually checks for the value of `HTTP/2.0` instead. Grafana Synthetics API blindly accepts whatever value is passed in and stores it, but the backend expects the value of `HTTP/2.0` (with the point-0 in the text), as seen by the errors in the following anonymized error after the change:

```
level=error target=https://foobar.example.com/ probe=SanFrancisco region=AMER instance=https://foobar/example.com job=check check_name=http source=synthetic-monitoring-agent label_cluster=prod msg="Invalid HTTP version number" version=HTTP/2.0
```

My organization found this out the hard way, since we imported existing checks into Terraform, ran the apply on what should've been a no-op, and found all our checks break simultaneously.

Even though the Grafana Synthetics UI shows `HTTP/2` as an option, checking the payload sent to the server shows a value of `HTTP/2.0` being sent. Also, when querying the API, or pulling the synthetics using a tool like Grizzly, the valid http versions showed the value of HTTP/2 before readding the value in the UI and `HTTP/2.0` afterwards.

This PR changes the documentation and the tests to use the correct value "HTTP/2.0". 
This is a followup to https://github.com/grafana/terraform-provider-grafana/issues/512 which tested that the provider was sending and receiving the value `HTTP/2` but never checked whether it was the expected value on the server side.

### DIFF
--- a/docs/resources/synthetic_monitoring_check.md
+++ b/docs/resources/synthetic_monitoring_check.md
@@ -199,7 +199,7 @@ EOS
       valid_http_versions = [
         "HTTP/1.0",
         "HTTP/1.1",
-        "HTTP/2",
+        "HTTP/2.0",
       ]
 
       fail_if_body_matches_regexp = [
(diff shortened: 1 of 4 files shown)


julienduchesne (approved):
I queried the API and it really is `HTTP/2.0`, thanks!

PR #1072, opened 2023-10-12
Alerting: update route schema to require group_by only for the root policy

yuri-tceretian:
In Alertmanager only the root policy requires the field `group_by` to be specified. In the nested policies it's optional and inherited from the closest parent that has this field specified.


julienduchesne (approved):
LGTM!

julienduchesne:
(Other than the failing test)

alexweav (approved):

PR #1080, opened 2023-10-17
Migrate smaller CI pipelines to Github Actions

julienduchesne:
We are looking at getting off Drone and these are the easiest 
The rest will be done at a later time


GNUmakefile
```diff
@@ -56,7 +56,7 @@ golangci-lint:
 		--rm \
 		--volume "$(shell pwd):/src" \
 		--workdir "/src" \
-		golangci/golangci-lint:v1.52 golangci-lint run ./...
+		golangci/golangci-lint:v1.52 golangci-lint run ./... -v
 
 linkcheck:
-	docker run -it --entrypoint sh -v "$$PWD:$$PWD" -w "$$PWD" python:3.11-alpine -c "pip3 install linkchecker && linkchecker --config .linkcheckerrc docs"
\ No newline at end of file
+	docker run --entrypoint sh -v "$$PWD:$$PWD" -w "$$PWD" python:3.11-alpine -c "pip3 install linkchecker && linkchecker --config .linkcheckerrc docs"
```
inkel: How about passing `--rm` so it deletes the container once it ends? 🤔

inkel (approved):
This is awesome! 👏🏽 thank you so much.

PR #1091, opened 2023-10-20
Service Account: Allow changing name (without recreating)

### LINKED ISSUE #1070: bug: grafana_service_account should not be recreated when name attribute changes
### Terraform Version
* Terraform: 1.5.7
* Terraform Grafana Provider: 2.3.3
* Grafana: 10.1.2

### Affected Resource(s)
* grafana_service_account

### Expected Behavior
Changing the `name` attribute should not force a resource replacement.

### Actual Behavior
When changing the name attribute the resource is replaced.

### Important Factoids
Changing the name of an existing service account is allowed thru the Grafana UI, but not with terraform.

julienduchesne:
Closes https://github.com/grafana/terraform-provider-grafana/issues/1070 
The name, while required, can be changed because the service account is queried and updated by its ID

### DIFF
--- a/internal/resources/grafana/resource_service_account.go
+++ b/internal/resources/grafana/resource_service_account.go
@@ -32,7 +32,6 @@ func ResourceServiceAccount() *schema.Resource {
 			"name": {
 				Type:        schema.TypeString,
 				Required:    true,
-				ForceNew:    true,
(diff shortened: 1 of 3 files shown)


inkel (approved):
Good call!

PR #1013, opened 2023-08-28
Use the swagger-generated OpenAPI client for the `Folder` service

nikimanoledaki:
## Use generated client for `Folder` service
- Add func `OAPIClientFromExistingOrgResource` to get client from existing or `OAPIClientFromNewOrgResource ` for new org, similarly to manual client
- Swap model used for folders in integration tests

## Testing

1. At the top of your directory, create a terraform file called `.terraformrc`. Use the template that can be found [here](https://github.com/grafana/terraform-provider-grafana#local-development-with-grafana), as below. Make sure to change the `grafana/grafana` field to point to the directory where the binary is built.
```
provider_installation {
   dev_overrides {
      "grafana/grafana" = "/Users/nikimanoledaki/grafanalabs/terraform-provider-grafana" # this path is the diretory where the binary is built
  }
  # For all other providers, install them directly from their origin provider
  # registries as normal. If you omit this, Terraform will _only_ use
  # the dev_overrides block, and so no other providers will be available.
  direct {}
}
```
2. Also create a `test.sh` script with the following:
```
#!/usr/bin/env bash
export GRAFANA_AUTH=admin:admin 
export GRAFANA_URL=http://localhost:3000 
export GRAFANA_VERSION=10.0.1 
docker kill grafanatests || echo "No running grafana tests container" 
docker run --rm -d --name grafanatests -p 3000:3000 grafana/grafana:$GRAFANA_VERSION
sleep 5
make testacc-oss # this tests OSS features - change this to any of the other make commands
```
3. Make sure to make this script executable with `chmod +x test.sh`
4. Run `./test.sh` to run the integration tests.

PR #1096, opened 2023-10-24
Update OAPI client dependency to use retries

nikimanoledaki:
This PR updates the OAPI client dep to use retries.


julienduchesne:
Already merged through https://github.com/grafana/terraform-provider-grafana/pull/1013

PR #1097, opened 2023-10-24
name is read-only, slug isn't

hertzsprung:
`slug` is an output, not an input


julienduchesne:
Absolutely right. You'll need to change it here: https://github.com/grafana/terraform-provider-grafana/blob/master/examples/data-sources/grafana_cloud_organization/data-source.tf#L2 and run `go generate` for CI to pass though

PR #1102, opened 2023-10-25
Testing: Common `Exists` and `Destroy` helpers

julienduchesne:
**!!11!!11! GENERICS !1!!!!!1111!!!**

These functions are repeated in all tests and are always the same pattern:
- Exists: Check that the resource can be queried in the correct org (and if not in the default org, that it can't be found in the default org)
- Destroy: Check that the resource can't be queried anywhere

To enable this, we only need two things:
- A short function that gets us the TF ID from an API model
- A short function that gets a resource from the API from a given ID

Benefits of this approach:
- Easier to add new helpers for resources
- Since `exists` and `destroyed` are opposites, if an error is made, it will usually fail in one of the two. This makes mistakes harder to make than with two manually composed functions
- It will allow us to migrate to the new generated client more easily
- Removes a bunch of code (once we have more than one resource using this pattern)

### DIFF
--- a/internal/common/errcheck.go
+++ b/internal/common/errcheck.go
@@ -20,7 +20,7 @@ func CheckReadError(resourceType string, d *schema.ResourceData, err error) (ret
 		return nil, false
 	}
 
-	if !strings.Contains(err.Error(), NotFoundError) {
+	if !IsNotFoundError(err) {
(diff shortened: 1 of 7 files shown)


inkel (approved):
Love it. Great usage of generics.

PR #1107, opened 2023-10-27
Merge `CheckOSSTestsEnabled` and `CheckOSSTestsSemver` helpers

julienduchesne:
I saw the semver function being used in a cloud test, which is invalid 
This should prevent that happening

### DIFF
--- a/internal/testutils/provider.go
+++ b/internal/testutils/provider.go
@@ -113,7 +113,7 @@ func IsUnitTest(t *testing.T) {
 }
 
 // CheckOSSTestsEnabled checks if the OSS acceptance tests are enabled. This should be the first line of any test that uses Grafana OSS features only
-func CheckOSSTestsEnabled(t *testing.T) {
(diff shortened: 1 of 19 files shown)


julienduchesne:
The cloud test in question is failing but I'll fix it in a future PR

internal/testutils/provider.go
```diff
@@ -183,4 +167,29 @@ func CheckEnterpriseTestsEnabled(t *testing.T) {
...
 		"GRAFANA_AUTH",
 	)
+	checkSemverConstraint(t, semverConstraintOptional...)
+}
+
+func checkSemverConstraint(t *testing.T, semverConstraintOptional ...string) {
+	t.Helper()
+
+	if len(semverConstraintOptional) > 1 {
+		panic("checkSemverConstraint accepts at most one argument")
```
inkel, line 177: Good call using a `panic` here! I was going to suggest `t.Fatal` instead, but then I realized it is not the test that fails but actually the test's definition, so this is definitely one of the right usages of `panic` 👏🏽

inkel (approved):

PR #1111, opened 2023-10-30
SM: rename checks to assertions

mem:
Synthetic Monitoring has a resource called `checks`, and for the specific case of MultiHTTP checks, each request can have zero or more "assertions", which in the code are called "checks" because they check that something is true. Let's not expose that to the user, as it will create confunsion. The UI calls these things "assertions", so we can use that term consistenly.


examples/resources/grafana_synthetic_monitoring_check/multihttp_complex.tf
```diff
@@ -49,31 +49,31 @@ resource "grafana_synthetic_monitoring_check" "multihttp" {
...
           type      = "TEXT"
           subject   = "RESPONSE_BODY"
           condition = "NOT_CONTAINS"
           value     = "xyyz"
         }
-        checks {
+        assertions {
           type       = "JSON_PATH_VALUE"
           condition  = "EQUALS"
           expression = "$.slideshow.author"
```
rdubrock, line 67: Off topic: I did some looking, and the underlying library used for json paths is [this one](https://github.com/tidwall/gjson#path-syntax), which doesn't use `$.` in the string, so I think we should ditch our references to that prefix

rdubrock (approved):

examples/resources/grafana_synthetic_monitoring_check/multihttp_complex.tf
mem: Ah, I see what you mean.

I _looks to me_ that the thing with `$` is partly confusion in JSONPath.

In principle, JSONPath _requires_ that expressions start with `$` because that references the root of the document (the outermost object or array).

The obvious question is what happens if you don't include it, and the general answer seems to be 🤷🏽 

Let's say `$.result` references an object inside the root object, if you have something like `{"result":"OK"}`.

What would `result` reference? "Some" object with "result" as key. If you have this document `{"a":{"result":"OK"},"b":{"result":42}}`, what would that be? It seems many (all?) implementations out there answer the question with "nothing, duh!", and no implementations try to find "the first thing that matches" or "everything that matches".

So, you are right, "$" is not doing anything there other than being pedantic about the fact that it _should_ be there.

rdubrock: OK I did some testing, and it works with `$` and without `$`, so it's all good. I was worried it would actually break with the `$`

PR #1124, opened 2023-11-03
Alerting Rule Groups: Fix provisioning in non-default org

### LINKED ISSUE #1120: grafana_rule_group specifying org_id at resource level has no effect
When creating an alert rule group, specifying org_id does not have the desired effect at the resource level. The alert rules are instead created in the default org (org_id = 1), rather than the specified org. Note that it works as expected when specifying org_id at the provider level.

### Terraform Version
* Terraform: v1.6.2
* Terraform Grafana Provider: v2.4.0
* Grafana: v10.2.0

### Affected Resource(s)
* grafana_rule_group

```hcl
provider "grafana" {}

resource "grafana_rule_group" "my_alert_rule" {
  org_id = 17

  ...
}
```

### Expected Behavior

Grafana alert group / rules are created in the org_id specified at the resource level.

### Actual Behavior

Grafana alert group / rules are created in the main org (org_id = 1), not the specified org_id at the resource level. Note that instead specifying the org_id at the provider level does work.

### Steps to Reproduce

1. Write code when sets org_id at the resource level for grafana_rule_group, and does not set org_id on the provider
2. `terraform apply`
3. `terraform plan` again. Observe that a drift is reported due to the alert rule existing in org_id = "1" instead of the specified org id.

### Important Factoids

The resource (https://github.com/grafana/terraform-provider-grafana/blob/master/internal/resources/grafana/resource_alerting_rule_group.go) doesn't appear to use the new org_id client infrastructure established in https://github.com/grafana/terraform-provider-grafana/issues/747.

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:

- GH-747

julienduchesne:
Closes https://github.com/grafana/terraform-provider-grafana/issues/1120 
Use the same pattern as other org-scoped resources to create the client

### DIFF
--- a/docs/resources/rule_group.md
+++ b/docs/resources/rule_group.md
@@ -116,8 +116,11 @@ EOT
 - `folder_uid` (String) The UID of the folder that the group belongs to.
 - `interval_seconds` (Number) The interval, in seconds, at which all rules in the group are evaluated. If a group contains many rules, the rules are evaluated sequentially.
 - `name` (String) The name of the rule group.
-- `org_id` (String) The ID of the org to which the group belongs.
 - `rule` (Block List, Min: 1) The rules within the group. (see [below for nested schema](#nestedblock--rule))
+
+### Optional
+
+- `org_id` (String) The Organization ID. If not set, the Org ID defined in the provider block will be used.
 
 ### Read-Only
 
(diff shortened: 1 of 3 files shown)


julienduchesne:
Enterprise tests failing is expected. I messed up the license here: https://github.com/grafana/terraform-provider-grafana/pull/1121

PR #1130, opened 2023-11-06
Retry by default on 401s

julienduchesne:
These errors come up a lot during testing
It seems like Grafana may fail to authenticate users when lots of requests are being processed


mhulscher:
@julienduchesne  might I suggest you add a retry on `400`  as well? See: https://github.com/grafana/grafana/issues/77462 . After a retry (another terraform apply), the apply succeeds.

inkel (approved):

malcolmholmes (approved):

PR #1136, opened 2023-11-08
User Tests: Use OpenAPI client to check if they exist

julienduchesne:
Gets rid of the manually written checkers
Also, added two common code functions to shorten each helper. We can now define a "checker" in 7 lines. This is definitely an improvement from the 30-50 lines of manually written code :)


inkel (approved):
Nice!

PR #1152, opened 2023-11-12
add jira integration to integrationTypes in grafna on call

romga:


julienduchesne:
An additional config value like this will need a test. Otherwise, it's hard to tell if it's the correct string

romga:
hi @julienduchesne !

When i was trying to import a grafana_oncall_integration with jira configuration that i have made manually, this is the configuration i got.

if you want, i can share here the resource :)

julienduchesne (approved):

PR #1161, opened 2023-11-14
Service Accounts: Use OpenAPI client

julienduchesne:
Use https://github.com/grafana/grafana-openapi-client-go instead of the manually generated client

Also:
- Uses a get-by-id call to read service accounts instead of a list
- This change exacerbated a long-standing issue with service accounts. They sometimes fail when creating in parallel. The list call gave it just enough slowness to not fail very often. By making it more efficient, it now reliably fails. To fix that, I added a mutex on service account creation

### DIFF
--- a/internal/resources/grafana/resource_service_account.go
+++ b/internal/resources/grafana/resource_service_account.go
@@ -2,14 +2,19 @@ package grafana
 
 import (
 	"context"
-	"log"
 	"strconv"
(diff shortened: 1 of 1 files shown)


inkel:
> By making it more efficient, it now reliably fails.

Loved this comment 👏🏽

inkel (approved):

PR #1176, opened 2023-11-21
Org Resource: Use OpenAPI client

julienduchesne:


internal/resources/grafana/data_source_users.go
```diff
@@ -101,3 +89,22 @@ func flattenUsers(items []*models.UserSearchHitDTO) []interface{} {
 
 	return userItems
 }
+
+func getAllUsers(client *goapi.GrafanaHTTPAPI) ([]*models.UserSearchHitDTO, error) {
+	allUsers := []*models.UserSearchHitDTO{}
+	var page int64 = 1
+	for {
+		params := users.NewSearchUsersParams().WithDefaults()
```
inkel: This one looks like it could be moved outside the `for` loop, am I correct?

inkel (approved):
LGTM!

internal/resources/grafana/data_source_users.go
julienduchesne: Indeed

PR #1185, opened 2023-11-28
Nil check on oncall integration templates

julienduchesne:
If a user defines an empty templates block, it currently panics 
An empty block should be functionally equivalent to not defining the block at all


inkel (approved):
LGTM but there are some comments from a CI check 🤔

julienduchesne:
Asking for a new review. I added more test cases and a DiffSupressFunc (if the templates is empty or missing should be the same outcome)

inkel (approved):
LGTM!!

quatauta:
@julienduchesne Thank you so much for this PR! This improves our life with the Grafana Terraform Provider! 🎉

PR #1192, opened 2023-11-30
Un-deprecate team external group

julienduchesne:
It's not hurting anything to leave it. AWS is actually going the opposite direction and splitting up their resources into multiple (ex: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket) 

Instead, we can give the choice to use the `team_sync` attribute on the team or the separate resource, one option or the other may be better depending on use case


inkel (approved):
Makes sense.

PR #1200, opened 2023-12-05
Alerting Rule Group: Use OpenAPI client

julienduchesne:
First of the alerting resources!


inkel (approved):

mpursley:
Hey @julienduchesne after applying this update, we noticed that we had a few alerts in our TF repo with `for: 0s`...
Not intentionally I think.  Your update here:
https://github.com/grafana/terraform-provider-grafana/pull/1200/files#diff-3ba89f7e59aad9a808dc7cb5f3cfe3b36171ba77fb28ea494339a24cc201fa39R349
- Was:`For:          json["for"].(string),` 
- Now:`For:          common.Ref(strfmt.Duration(forDuration)),`

...allows `for: 0s` to pass the `tf plan`, but fails on `tf apply` with this error... 
```Error: unable to parse  as duration```

I changed all our mentions of `for: 0s` to `for: 5m` as that's better for us.  But, seems like `for: 0s` should also be valid?  (e.g. fire immediately, as in https://community.grafana.com/t/grafana-alerting-evaluation/85064/5)

I'm not sure if this should be a feature or a bug? Seems like `for: 0s` _should be_ ok? If not, that it should fail the `tf plan`? I started a new issue for it here... https://github.com/grafana/terraform-provider-grafana/issues/1273

PR #1206, opened 2023-12-07
Set-up muxing with new provider framework

julienduchesne:
This PR sets up the migration from the old provider SDK (what we're using) to the new plugin framework 
The migration process is outlined here: https://developer.hashicorp.com/terraform/plugin/framework/migrating

Essentially, it consists of running two providers side by side with the exact same config and migrate resources one by one 
The client configuration is now done using the new plugin framework config (go struct) object, and the legacy provider config (d.Get(...)) is translated into the plugin framework config 
This PR is a no-op in the sense that no current resources are modified

This PR is also the first step to use TF code generation (https://developer.hashicorp.com/terraform/plugin/code-generation) since it generates TF plugin framework code (not old SDK code)

**Note: this is a pretty big PR and the Github diff is not doing us any favors. Happy to review this synchronously with someone**


internal/provider/configure_clients.go
```diff
@@ -0,0 +1,310 @@
...
+	u, err := url.Parse(providerConfig.URL.ValueString())
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse API url: %v", err.Error())
+	}
+	apiPath, err := url.JoinPath(u.Path, "api")
+	if err != nil {
+		return nil, fmt.Errorf("failed to join API path: %v", err.Error())
+	}
+
+	userInfo, orgID, APIKey, err := parseAuth(providerConfig)
```
spinillos:
```suggestion
	userInfo, orgID, apiKey, err := parseAuth(providerConfig)
```

internal/provider/configure_clients.go
```diff
@@ -0,0 +1,310 @@
...
+		Host:             u.Host,
+		BasePath:         apiPath,
+		Schemes:          []string{u.Scheme},
+		NumRetries:       int(providerConfig.Retries.ValueInt64()),
+		RetryTimeout:     time.Second * time.Duration(providerConfig.RetryWait.ValueInt64()),
+		RetryStatusCodes: setToStringArray(providerConfig.RetryStatusCodes.Elements()),
+		TLSConfig:        tlsClientConfig,
+		BasicAuth:        userInfo,
+		OrgID:            orgID,
+		APIKey:           APIKey,
```
spinillos:
```suggestion
		APIKey:           apiKey,
```

internal/provider/configure_clients.go
```diff
@@ -0,0 +1,310 @@
...
+			headers[k] = vString.ValueString()
+		} else {
+			return nil, fmt.Errorf("invalid header value for %s: %v", k, v)
+		}
+	}
+
+	return headers, nil
+}
+
+func createTempFileIfLiteral(value string) (path string, tempFile bool, err error) {
```
spinillos, line 214: You can remove `tempFile bool` and check if the path is empty. In that case, as the way that you used this function, you should return an empty string instead of `value`.

spinillos (approved):

PR #1210, opened 2023-12-07
Ensure all documentation examples are tested and valid

julienduchesne:
In some other PRs, I've been relying less and less on the examples files for tests (example: https://github.com/grafana/terraform-provider-grafana/pull/1209)
The reason to not use them is that these configs can't be modified easily when testing updates and they don't lend themselves well to parallel testing

However, this can lead to examples staleness.
In order to ensure that the examples always stay up-to-date, this adds a new test pipeline that applies all example files 
This pipeline runs with a parallelism of 1 so we don't really care about conflicts between examples 
The pipeline also checks if resources are missing an example file and will fail if any are missing


K-Phoen (approved):
Very nice!

Issue #697, opened 2022-10-27
Allow alerting resources to be optionally editable from UI


teokeecheng:
Currently, for any rule group created from terraform, it will show a status 'Provisioned', and not allow any amendment through console. However, when we view it from console, not all alert details are displayed (e.g. queries and criteria). Would be good if there is an option in terraform resource to indicate if the alert can be editable from UI for testing and flexibility.

greatvovan:
I do not think this issue is relevant to TF provider project. This is how provisioned alerts work in Grafana. If you provision an alert through [provisioning API](https://grafana.com/docs/grafana/latest/developers/http_api/alerting_provisioning/), you'll see the same.

There are also open issues for this feature request in Grafana repo. See: https://github.com/grafana/grafana/issues/57793

HartmannVolker:
Related issue in the grafana repo: https://github.com/grafana/grafana/issues/51211

tomashejatko:
Hello, this is now resolved upstream in Grafana 9.3.0 and should be implemented: https://github.com/grafana/grafana/pull/58410

greatvovan:
Hello
Reading the comments to the pull request, I understand the option was added to API only. The motivation was to relief the pain for all kinds of provisioned alerts. I am wondering, how it will help me with Terraformed alerts?

tomashejatko:
I see you have posted a question to the original issue but with a somewhat negative answer. So maybe dumb question from my side: 

I thought that TF provider is using Grafana API to create dashboard alerts and stuff (My Grafana access logs and https://github.com/grafana/terraform-provider-grafana/blob/master/grafana/resource_alerting_rule_group.go#L11 kinda approves this), so I thought that you can add an optional parameter to https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/rule_group which will send `x-disable-provenance` header when set to true.

But maybe the problem is that workaround from https://github.com/grafana/grafana/pull/58410 is using provisioning API and TF provider is using "another API"?

Just to clarify what I want from this feature (using similar procedure from the dashboard and for alert rules creation):

1. I have `grafana_rule_group` resource defined with a full "data model" in my .tf file
2. I want to go to UI to make changes to the alert, even changing the threshold is much more feasible if you see graphical representation
3. Save alert group in UI
4. Make `terraform plan` which tells me what I have to change in .tf files

Maybe it sound bit complicated but I am doing this with my terraformed dashboards for about 3 years :)

[...]

PR #1214, opened 2023-12-08
Alerting Rule Groups: Support editing from UI

julienduchesne:
Issue: https://github.com/grafana/terraform-provider-grafana/issues/697 
Requires https://github.com/grafana/grafana-openapi-client-go/pull/45 

There has been informal support through setting the `X-Disable-Provenance` header manually but it doesn't work that well and the current way it's done is really hard to test. Notably, it currently fails on updates for contact points (see issue) and it isn't easily discoverable

This adds a field on the rule group to allow editing from UI

Note: The provenance is actually a rule attribute (not group). Because we manage the rule group as a whole and the setting has to set through a header, I chose to add the setting as a group-scoped attribute because it's easier to manage 
Otherwise, we'd have to do POSTs, PUTs and DELETEs for each rule within the group

Also, I added tests to make sure that the provenance is correctly synced from Terraform to the alerting service and that it can be read back

### DIFF
--- a/docs/resources/rule_group.md
+++ b/docs/resources/rule_group.md
@@ -120,7 +120,7 @@ EOT
 
 ### Optional
 
-- `allow_editing_from_ui` (Boolean) Whether to allow editing of the rule group from the Grafana UI. Defaults to `false`.
+- `disable_provenance` (Boolean) Allow modifying the rule group from other sources than Terraform or the Grafana API. Defaults to `false`.
 - `org_id` (String) The Organization ID. If not set, the Org ID defined in the provider block will be used.
 
 ### Read-Only
--- a/docs/resources/rule_group.md
+++ b/docs/resources/rule_group.md
@@ -120,6 +120,7 @@ EOT
 
 ### Optional
 
+- `allow_editing_from_ui` (Boolean) Whether to allow editing of the rule group from the Grafana UI. Defaults to `false`.
 - `org_id` (String) The Organization ID. If not set, the Org ID defined in the provider block will be used.
 
 ### Read-Only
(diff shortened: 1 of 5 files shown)


internal/resources/grafana/resource_alerting_rule_group.go
```diff
@@ -198,9 +204,30 @@ func readAlertRuleGroup(ctx context.Context, data *schema.ResourceData, meta int
...
+			return diag.FromErr(err)
+		}
+		r := ruleResp.Payload
+		data.Set("org_id", strconv.FormatInt(*r.OrgID, 10))
+		packed, err := packAlertRule(r)
+		if err != nil {
+			return diag.FromErr(err)
+		}
+		if r.Provenance != "" {
+			allowEditingFromUI = false
```
spinillos: If at least one rule doesn't have Provenance set, do we completely disallow ui edition?

julienduchesne: Yeah, it'll mean that Terraform sees the whole rule group as not editable, and there will be a diff. On apply, all rules in the group will be updated to be the same (as configured)

spinillos (approved):

docs/resources/rule_group.md
```diff
@@ -120,6 +120,7 @@ EOT
 
 ### Optional
 
+- `allow_editing_from_ui` (Boolean) Whether to allow editing of the rule group from the Grafana UI. Defaults to `false`.
```
yuri-tceretian: I am not sure I like this name because it allows editing in general (via file-provisioning, unstable API, which is used by UI, but can be accessed directly as well).

julienduchesne: I think it could just be disable_provenance and be described further in description. I was trying to be clever but it's not necessary 😃

JohnnyQQQQ: I think `disable_provenance` is better, as it would align more with the naming in the header and the references in the docs.

julienduchesne: Done!

PR #1231, opened 2023-12-14
Fix `TestAccGrafanaServiceAccountFromCloud`

julienduchesne:
This is a weird one. This test was failing but returning: `a stack with the name 'tfserviceaccounttestqnkmr08pvm' already exists` 
See https://github.com/grafana/terraform-provider-grafana/actions/runs/7211112859/job/19645774937

The reason for that error was that the stack name in the test was too long (over 29 chars) and the API returns a 409 Conflict in that case (weird). 
I changed the stack name + I fixed the error management and now the test passes


julienduchesne:
New Cloud tests run: https://github.com/grafana/terraform-provider-grafana/actions/runs/7211697756/job/19647722284

spinillos (approved):

PR #1233, opened 2023-12-15
SM Installation: Determine API URL from stack's region slug

julienduchesne:
Most APIs (other than some exceptions) have determinate API URLs based on the stack's region slug 
For the exceptions which are the older regions, we can keep a map


julienduchesne:
Test run here: https://github.com/grafana/terraform-provider-grafana/actions/runs/7217459338/job/19665313417

spinillos (approved):

PR #1247, opened 2023-12-19
Alerting Contact Points: Refer by name

julienduchesne:
Background:
Currently, the contact point's ID is a collection of notifier UIDs. However, in Grafana (and in the resource's import ref), it's referred to by name

This PR:
Change the Terraform ID from semi-colon separated UIDs to the contact point name. This will be easier to test, and this will allow simplifying the code in the next major version (removing the code fetching notifiers by UID) Since the UID reading logic is moved to the read function, it also allows importing contact points in Crossplane where the import function is not used


spinillos (approved):

duncan485:
Shouldn't this have been announced as a breaking change? As we created the resources with a version prior to this change, our TF state broke after upgrading the provider. As it tried to fetch  the contact points based on the ID which was already in state, rather than the name.

julienduchesne:
> Shouldn't this have been announced as a breaking change? As we created the resources with a version prior to this change, our TF state broke after upgrading the provider. As it tried to fetch the contact points based on the ID which was already in state, rather than the name.

This was tested with IDs in the state. The read by name or ID should still work in the latest version also. As tested here: https://github.com/grafana/terraform-provider-grafana/blob/main/internal/resources/grafana/resource_alerting_contact_point_test.go#L93-L108. (An import test will set only the ID and try to read the resource from that)

duncan485:
Not sure what happend then, because in our case (grafana cloud), we got a `context deadline exceeded (HTTP 500)` on 

`Get https://company.grafana.net/api/v1/provisioning/contact-points?name=00000000-00000-00000-0000-0000000`

fyi, we go this on updating to grafana provider from 2.7.0 --> 2.10.0

julienduchesne:
> Not sure what happend then, because in our case (grafana cloud), we got a `context deadline exceeded (HTTP 500)` on
> 
> `Get https://company.grafana.net/api/v1/provisioning/contact-points?name=00000000-00000-00000-0000-0000000`
> 
> fyi, we go this on updating to grafana provider from 2.7.0 --> 2.10.0

huh, interesting. `00000000-00000-00000-0000-0000000` is not a valid ID either. Not sure what's up with that error

duncan485:
I redacted the actual uid before posting it here. just like company.grafana.net was not the actual endpoint.

julienduchesne:
Right 😄. The logic goes that if a contact point isn't found by name (404, or empty list), then it tries by ID. In your case, it's a 500, so that's going to crash the process right away. As if you had gotten a 500 while trying to fetch the contact points by ID. 

Not sure if the 500 is linked to the API call in particular, the `?name=` query is not a new thing, it's at least as old as the resource itself

duncan485:
I just checked, if I query any name, other than an existing one, I get a 500 error instead of 404 or []. For example, 
https://company.grafana.net/api/v1/provisioning/contact-points?name=doesnotexist returns 500. 

Maybe it has to do with how it is implemented in grafana-cloud.

julienduchesne:
Interesting. I'll test it out and put out a fix

julienduchesne:
Hey, the 500 error should be fixed soon in Grafana Cloud. It was a Grafana bug, and the fix has been merged to the main branch

duncan485:
@julienduchesne awesome, thanks !

PR #1249, opened 2023-12-20
fix(contact points): panic when post contact points request fails

akselleirv:
When I tried to create a contact point the provider paniced when I get a 401 back from Grafana. It was due to not checking the error before accessing the response struct.


julienduchesne:
Will be fixed by https://github.com/grafana/terraform-provider-grafana/pull/1247

akselleirv:
> Will be fixed by #1247

Okay! Then I'll close this PR.

PR #1251, opened 2023-12-21
Alerting Contact Points: Support org-specific provisioning

julienduchesne:
Adds the `org_id` attribute to be consistent with other org-specific resources


Cleymax (approved):
LGTM

PR #1252, opened 2023-12-21
Alerting Message Templates: Add `disable_provenance`

julienduchesne:
Issue: https://github.com/grafana/terraform-provider-grafana/issues/697

Unfortunately, the API doesn't return the provenance so we have to rely on the last applied value in the state

### DIFF
--- a/docs/resources/message_template.md
+++ b/docs/resources/message_template.md
@@ -36,6 +36,7 @@ resource "grafana_message_template" "my_template" {
 
 ### Optional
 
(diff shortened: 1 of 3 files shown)


Cleymax (commented):
No tests ?

Cleymax (commented):
LGTM

julienduchesne:
> No tests ?

No use. The API doesn't return it. It's set to whatever value the user provides

PR #1259, opened 2024-01-02
Reports: Add missing time range in old reports and update test client

spinillos:
It sets the missing time range in old reports and it updates the test client.


internal/resources/grafana/resource_report_test.go
```diff
@@ -153,7 +152,7 @@ func TestAccResourceReport_InOrg(t *testing.T) {
 	})
 }
 
-func testAccReportCheckExists(rn string, report *gapi.Report) resource.TestCheckFunc {
+func testAccReportCheckExists(rn string, report *models.ConfigDTO) resource.TestCheckFunc {
```
julienduchesne: You can remove this function and create a standard helper instead: https://github.com/grafana/terraform-provider-grafana/pull/1213

It properly checks that the report exists in the expected org (and not in the default org)

julienduchesne (approved):

julienduchesne:
LGTM!

PR #1263, opened 2024-01-05
Datasource Permissions: Use OpenAPI client

julienduchesne:


internal/resources/grafana/resource_data_source_permission.go
```diff
@@ -82,25 +86,24 @@ Manages the entire set of permissions for a datasource. Permissions that aren't
...
 	if v, ok := d.GetOk("permissions"); ok {
 		list = v.(*schema.Set).List()
 	}
 
-	_, datasourceIDStr := SplitOrgResourceID(d.Get("datasource_id").(string))
-	datasourceID, _ := strconv.ParseInt(datasourceIDStr, 10, 64)
-
-	dataSource, err := client.DataSource(datasourceID)
+	_, datasourceID := SplitOrgResourceID(d.Get("datasource_id").(string))
+	resp, err := client.Datasources.GetDataSourceByID(datasourceID)
```
IevaVasiljeva, line 97: We should probably extend the data source resource to include UID, so that we can operate with UIDs directly.

IevaVasiljeva (approved):

internal/resources/grafana/resource_data_source_permission.go
julienduchesne: Indeed. We can definitely support both in the read function: https://github.com/grafana/terraform-provider-grafana/blob/master/internal/resources/grafana/resource_folder.go#L122

PR #1282, opened 2024-01-12
Add the OpenAPI-generated SLO client 

julienduchesne:
Client here: https://github.com/grafana/slo-openapi-client
This replaces the old https://github.com/grafana/grafana-api-golang-client

I was also able to remove the `DeprecatedGrafanaClient` attribute because it was only used by SLO at this point

### DIFF
--- a/internal/resources/slo/resource_slo.go
+++ b/internal/resources/slo/resource_slo.go
@@ -5,12 +5,19 @@ import (
 	"fmt"
 	"regexp"
 
-	gapi "github.com/grafana/grafana-api-golang-client"
+	slo "github.com/grafana/slo-openapi-client/go"
 
 	"github.com/grafana/terraform-provider-grafana/internal/common"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+const (
+	QueryTypeFreeform  string = "freeform"
+	QueryTypeHistogram string = "histogram"
+	QueryTypeRatio     string = "ratio"
+	QueryTypeThreshold string = "threshold"
 )
 
 func ResourceSlo() *schema.Resource {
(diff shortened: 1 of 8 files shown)


internal/resources/slo/resource_slo_test.go
```diff
@@ -138,12 +145,19 @@ func testAlertingExists(expectation bool, rn string, slo *gapi.Slo) resource.Tes
...
 }
 
-func testAccSloCheckDestroy(slo *gapi.Slo) resource.TestCheckFunc {
+func testAccSloCheckDestroy(slo *slo.Slo) resource.TestCheckFunc {
 	return func(s *terraform.State) error {
-		client := testutils.Provider.Meta().(*common.Client).DeprecatedGrafanaAPI
-		client.DeleteSlo(slo.UUID)
+		client := testutils.Provider.Meta().(*common.Client).SLOClient
+		req := client.DefaultAPI.V1SloIdGet(context.Background(), slo.Uuid)
+		gotSlo, resp, _ := req.Execute()
```
spinillos, line 152: Don't you check the error here? If you have any error doing the request maybe you cannot retrieve status code or status type.

julienduchesne: The resp is always not-nil. And so, to check that a SLO was destroyed, checking that either we get a 404, or an SLO with a "deleting" status is sufficient. All other cases cause a test failure

internal/resources/slo/resource_slo.go
```diff
@@ -5,14 +5,21 @@ import (
...
+	slo "github.com/grafana/slo-openapi-client/go"
 
 	"github.com/grafana/terraform-provider-grafana/internal/common"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
 
+const (
+	QueryTypeFreeform  string = "freeform"
```
joeblubaugh, line 17: question: No way to extract this from the generated API client?

joeblubaugh (approved):

internal/resources/slo/resource_slo.go
julienduchesne: hmm, maybe if this is turned into an enum in the spec. I'll take a look

julienduchesne: Will get it through a follow-up

PR #1312, opened 2024-01-29
Update oncall shift interval param docs

matiasb:
Related to https://github.com/grafana/support-escalations/issues/9142

Double-checked about `time_zone` parameter, couldn't reproduce the issue (it worked as optional in my tests, and reviewing the logic, it seems that should be always the case).


internal/resources/oncall/resource_shift.go
```diff
@@ -126,7 +126,7 @@ func ResourceOnCallShift() *schema.Resource {
 				Type:         schema.TypeInt,
 				Optional:     true,
 				ValidateFunc: validation.IntAtLeast(1),
-				Description:  "The positive integer representing at which intervals the recurrence rule repeats.",
+				Description:  "The positive integer representing at which intervals the recurrence rule repeats. Required if a frequency is set (fallback to 1 if omitted but it will later be detected as a change).",
```
julienduchesne: Rather than document this in the description, I would use a `RequiredWith`

julienduchesne (approved):
LGTM!

Issue #402, opened 2022-02-24
Support timezones for reports


ekpdt:
Fix https://github.com/grafana/terraform-provider-grafana/blob/master/grafana/resource_report.go#L311 to support accepting a timezone (e.g. `America/Los_Angeles`) via the resource block.

julienduchesne:
Both the start_time and the end_time of the schedule support the `RFC3339` format so you can set dates like this: `start_time    = "2020-01-01T00:00:00-07:00"`. However, it's true that it does not support daylight saving. Perhaps another attribute for timezone could be useful

ekpdt:
Yes, that's precisely the issue. The docs rightly indicate that the time is converted to UTC and stored that way in the backend. In terms of local time, scheduled reports drift back and forth by an hour for those in DST-prone areas.

inkel:
The "problem" with using named timezones is that, depending on the OS you're running, the named zone might not exist, as stated in [`time.LoadLocation` documentation](https://pkg.go.dev/time#LoadLocation):

> The time zone database needed by LoadLocation may not be present on all systems, especially non-Unix systems. LoadLocation looks in the directory or uncompressed zip file named by the ZONEINFO environment variable, if any, then looks in known installation locations on Unix systems, and finally looks in $GOROOT/lib/time/zoneinfo.zip.

Using something like [`time.FixedZone`](https://pkg.go.dev/time#FixedZone) to define the timezone manually won't work either, as in this case the offset would need to change every time DST changes.

This means that the resulting resource definition might not be 100% portable, which might not be an issue for particular use cases but defeats the purpose of the provider as a tool to guarantee reproducibility.

julienduchesne:
I do remember facing issue loading timezones which is why I went with fixed time zones for the the initial resource

inkel:
Yup. This is an issue that I've found in many different languages I used at work in the last 20+ years: PHP, C, Java, Python, Go, Ruby… all of them have the same issue with named timezones when you try to port them to other OSes.

inkel:
For some reason my mind decided to forget about Perl and C# 😬 (confession time: I actually like C#)

ekpdt:
>This means that the resulting resource definition might not be 100% portable, which might not be an issue for particular use cases but defeats the purpose of the provider as a tool to guarantee reproducibility.

[...]

PR #1320, opened 2024-01-30
Reports: Add schedule timezone for reports

spinillos:
Contributes to: https://github.com/grafana/terraform-provider-grafana/issues/402

It add the possibility to customize the report's timezone when we schedule them.

I added a Timezone validation and imported `tzdata` (it increases the client ~800kb). `tzdata` is imported because some environments don't include some timezones and it could fail. I can remove it if it doesn't worth it.


internal/resources/grafana/resource_report.go
```diff
@@ -571,3 +583,9 @@ func parseCustomReportInterval(i interface{}) (int, string, error) {
 
 	return number, unit, nil
 }
+
+func validateTimezone(i interface{}) error {
+	timezone := i.(string)
+	_, err := time.LoadLocation(timezone)
+	return err
+}
```
julienduchesne:
```suggestion
func validateTimezone(i interface{}, path cty.Path) diag.Diagnostics {
	timezone := i.(string)
	_, err := time.LoadLocation(timezone)
	return diag.FromErr(err)
}
```

This will allow you to use it directly: `ValidateDiagFunc: validateTimezone`

internal/resources/grafana/resource_report_test.go
```diff
@@ -114,6 +116,7 @@ func TestAccResourceReport_basic(t *testing.T) {
 					resource.TestCheckResourceAttr("grafana_report.test", "schedule.0.workdays_only", "true"),
 					resource.TestCheckResourceAttr("grafana_report.test", "schedule.0.start_time", "2020-01-01T07:00:00Z"), // Date transformed to UTC
 					resource.TestCheckResourceAttr("grafana_report.test", "schedule.0.end_time", "2020-01-15T08:30:00Z"),   // Date transformed to UTC
```
julienduchesne, line 120: Do the given times require timezone information? If that's the case, it's not possible to fully use the timezone, right?

spinillos: I'm seeing that the frontend is sending startDate and endDate with the timezone selected.
In the backend, it uses the timezone in the scheduler [here](https://github.com/grafana/grafana-enterprise/blob/main/src/pkg/extensions/report/scheduler/utils.go#L81) copying the location information into the start date, but it isn't applied to the end date. Its working because the dates have the timezone applied in the create/update report calls.

When we retrieve the dashboards we applied the timezone in both values, but internally isn't in both... 

I'll talk with sharing team tomorrow about that

spinillos: I re-checked it again and I'm seeing that currently we need to set the start/end dates with the timezone, but we are forcing `GMT` timezone. The result is that in the backend is using GMT to send the report ignoring the timezone set in the dates (except for the end time like I said before). 

You can see the test code [here](https://github.com/grafana/terraform-provider-grafana/blob/master/examples/resources/grafana_report/all-options.tf#L19-L20). It sets the timezones but in the test result, these values aren't appear because report's backend is applying the timezone for the response. So this test is fine.

I think that because this inconsistency in the end date, some reports aren't finishing in the correct date and it was wrong all the time 😅.

We can do 2 things:
* Add the timezone field like I did and apply it in the star/end dates ignoring the timezone from these fields.
*  Don't add the timezone and extract the timezone from the dates.

WDYT that it's the best way to do it?

julienduchesne: > Don't add the timezone and extract the timezone from the dates.

I think this one doesn't work because timezone indicators don't include daylight savings, it's just UTC or `[-+]\d`

> Add the timezone field like I did and apply it in the star/end dates ignoring the timezone from these fields.

What about, we combine both of those options:

- If the user has timezone indicators in their dates AND the `timezone` field set, throw an error because they shouldn't be used together
- If the `timezone` is set but the dates don't have timezones to them, use the `timezone` field
- If the `timezone` field isn't set, but the dates have timezones, use those

What do you think?

spinillos: We have two things:
* Timezone is mandatory to send into the reports and we can't leave it empty. (Now its hardcoded to GMT)
* We need to change the validation function from start/end time since we are validating against [RFC3339](https://golang.cafe/blog/golang-time-format-example.html). `DateTime` could be close to what we need.

julienduchesne: Yeah, we shouldn't require RFC3339 in the dates themselves if we're setting the timezone field. Perhaps a RFC3339 but without the TZ info (just `2006-01-02T15:04:05`)

spinillos: I just updated everything.

I needed to create a custom time format and I set it in the description of the field. If not isn't going to be intuitive enough.

internal/resources/grafana/resource_report_test.go
```diff
@@ -38,9 +38,10 @@ func TestAccResourceReport_Multiple_Dashboards(t *testing.T) {
 					resource.TestCheckResourceAttr("grafana_report.test", "recipients.0", "some@email.com"),
 					resource.TestCheckNoResourceAttr("grafana_report.test", "recipients.1"),
 					resource.TestCheckResourceAttr("grafana_report.test", "schedule.0.frequency", "monthly"),
-					resource.TestCheckResourceAttrSet("grafana_report.test", "schedule.0.start_time"), // Date set to current time
-					resource.TestCheckResourceAttr("grafana_report.test", "schedule.0.end_time", ""),  // No end time
+					resource.TestCheckResourceAttr("grafana_report.test", "schedule.0.start_time", "2024-02-10T15:00:00-05:00"),
```
julienduchesne, line 41: Can't wrap my head around how this works on Grafana's side. We're sending a data with an offset to the API (-5) and also a timezone. What happens when daylight savings kicks in? A TF drift, a report sent at a different hour, or does Grafana send a correct date at read-time and all is good?

spinillos: Reports apply the timezone offset [here](https://github.com/grafana/grafana-enterprise/blob/main/src/pkg/extensions/report/scheduler/utils.go#L55).

julienduchesne: heh, interesting

julienduchesne (approved):
LGTM. Thanks!

PR #1321, opened 2024-01-30
SLO: Add destination datasource type

julienduchesne:
Seems like it's required to apply SLOs now
Also:
- Fix errors so that the actual error message is shown
- Remove redundant `&schema.Schema`


elainevuong (approved):
LGTM, thank you for doing this 🙏 !

PR #1322, opened 2024-01-30
Cloud Resources: Use OpenAPI client for tests

julienduchesne:
Leaves all the actual code the same, but all he functions in *_test.go files now use the OpenAPI client


Duologic (approved):
+1 for TDD

PR #1323, opened 2024-01-31
Reports: Add report_variables to dashboards

### LINKED ISSUE #1008: Add support for dashboard variables in grafana_report
Currently the UI can handle setting variables on dashboards for reports but the provider can't. A simple use case from me is I want to automate sending reports to teams every Monday. This is done with a single dashboard with a `team` variable that I'm not able to control from Terraform with the `grafana_report` resource.

spinillos:
Closes https://github.com/grafana/terraform-provider-grafana/issues/1008

It adds `report_variables` under `dashboards` using the new way to do it.

Reports expect a map of a list of string for this field, but for a SDK limitation/bug it doesn't work even if we specify the type of the map. 

The workaround of this was to expect a map of strings only and parse the values send the correct values.

So:
```tf
report_values = {
  query0 = "val1,val2"
  query1 = "val3,val4"
}
```
will send:
```json
"report_values" {
    "query0": ["val1", "val2"]
    "query1": ["val3", "val4"]
}
```

I was tested it locally and I checked that deprecated message isn't prompted.

### DIFF
--- a/internal/resources/grafana/resource_report.go
+++ b/internal/resources/grafana/resource_report.go
@@ -273,6 +273,13 @@ func ResourceReport() *schema.Resource {
 								return oldValue == "1" && newValue == "0"
 							},
 						},
+						"report_variables": {
+							Type:             schema.TypeMap,
+							Description:      "Add report variables to the dashboard. Values should be separated by commas.",
+							Optional:         true,
+							Elem:             schema.TypeString,
+							ValidateDiagFunc: validateReportVariables,
+						},
 					},
 				},
 				DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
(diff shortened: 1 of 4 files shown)


internal/resources/grafana/resource_report.go
```diff
@@ -571,3 +580,40 @@ func parseCustomReportInterval(i interface{}) (int, string, error) {
...
+				values[i] = v.(string)
+			}
+			newMap[k] = strings.Join(values, ",")
+		case string:
+			newMap[k] = strings.Split(rvType, ",")
+		}
+	}
+
+	return newMap
+}
```
julienduchesne: Took me a while to understand that this function goes both ways.

This is a bit of a nit, but even if it's more code, I'd still rather have two functions (pack/unpack) with type validation in each (and errors) than a single function that converts depending on the input

julienduchesne (approved):
Nice! Just one comment

internal/resources/grafana/resource_report.go
spinillos: I split it into two functions. One for the request and the other for the response. 

In for request `interface{}` is the strings separated by coma and for the response, the interface{} is returned as an array of interfaces (a list of string).

julienduchesne: Thank you!

Issue #1324, opened 2024-01-31
Grafana Cloud Stack - Description not set on create


fishgrimsby:
### Terraform Version
* Terraform: v1.3.9
* Terraform Grafana Provider: 2.10.0
* Grafana: Cloud

### Affected Resource(s)
Please list the resources as a list, for example:
* grafana_cloud_stack

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
resource "grafana_cloud_stack" "this" {
  provider = grafana.cloud

  name        = var.hosted_zone_name != null ? aws_route53_record.this[0].fqdn : var.stack_name
  slug        = var.slug
  region_slug = var.region_slug
  description = var.stack_description
  url         = var.hosted_zone_name != null ? "https://${aws_route53_record.this[0].fqdn}" : var.url
}
```

### Expected Behavior
Description should have been set on the newly created Grafana stack in Grafana Cloud

### Actual Behavior
No description is set on the newly created stack in Grafana Cloud

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. Create any new grafana_cloud_stack resource with terraform
2. Observe no description present on the newly created Grafana stack

### Important Factoids
When applying the terraform again, the update operation successfully sets the description. It is just the initial create of the stack resource where it does not

PR #1325, opened 2024-01-31
Cloud Stack: Fix description on stack create

fishgrimsby:
This pull request should fix the issue with grafana_cloud_stack resources being created with an empty description. See issue #1324

### DIFF
--- a/internal/resources/cloud/resource_cloud_stack_test.go
+++ b/internal/resources/cloud/resource_cloud_stack_test.go
@@ -31,6 +31,7 @@ func TestResourceStack_Basic(t *testing.T) {
 		resource.TestMatchResourceAttr("grafana_cloud_stack.test", "id", common.IDRegexp),
 		resource.TestCheckResourceAttr("grafana_cloud_stack.test", "name", resourceName),
 		resource.TestCheckResourceAttr("grafana_cloud_stack.test", "slug", resourceName),
+		resource.TestCheckResourceAttr("grafana_cloud_stack.test", "description", stackDescription),
 		resource.TestCheckResourceAttr("grafana_cloud_stack.test", "status", "active"),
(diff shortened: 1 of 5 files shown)


internal/resources/cloud/resource_cloud_stack_test.go
```diff
@@ -31,6 +31,7 @@ func TestResourceStack_Basic(t *testing.T) {
 		resource.TestMatchResourceAttr("grafana_cloud_stack.test", "id", common.IDRegexp),
 		resource.TestCheckResourceAttr("grafana_cloud_stack.test", "name", resourceName),
 		resource.TestCheckResourceAttr("grafana_cloud_stack.test", "slug", resourceName),
+		resource.TestCheckResourceAttr("grafana_cloud_stack.test", "description", stackDescription),
```
julienduchesne, line 34: This will need to be set in the test config

fishgrimsby: Hey @julienduchesne, I have amended the test config as you rightly pointed out. However, I am struggling with the Cloud unit tests as I am getting the following `resource_cloud_stack_test.go:45: Step 2/7, expected an error but got none` but I also get this when running them against the master branch directly, so it doesn't appear new with my changes.

I'm not really sure if it is something to do with my org or perhaps my dev machine does not produce quite the same error to meet the regex

julienduchesne: Cloud tests are tricky to run. I'll try them out on our end

fishgrimsby: That would be much appreciated, thank you. I'm happy to try to fix anything that comes out of them.

julienduchesne (approved):
LGTM! All tests pass. Thanks!

PR #1326, opened 2024-01-31
Update SM docs with access policy information

mem:
API tokens were removed a while ago, so update this documentation to mention access policies, which is what's actually used.


julienduchesne (approved):
Thanks

PR #1342, opened 2024-02-09
SM: Higher default for multi HTTP checks

### LINKED ISSUE #1173: default timeout ( 3s ) is too low for multihttp checks
The default timeout on a synthetic monitoring check is 3000ms

https://github.com/grafana/terraform-provider-grafana/blob/5abae1fd4e9dbab819a4d515ba3481422bcac8b1/docs/resources/synthetic_monitoring_check.md?plain=1#L423-L428

For multi-http checks, this tends to be too short.  Multi-http checks are backed by k6 it can sometimes take 1 second to start the k6 process,  leaving less time for the check to complete before the timeout.  5 seconds would probably be a more appropriate default for multi-http checks.

julienduchesne:
As suggested here: https://github.com/grafana/terraform-provider-grafana/issues/1173 
Can't hurt to raise the defaults for those. Not going to raise all of them because it would cause too much drift. Multi-HTTP is newer and is the one that needs it so it's fine, I think

### DIFF
--- a/internal/resources/syntheticmonitoring/resource_check.go
+++ b/internal/resources/syntheticmonitoring/resource_check.go
@@ -16,6 +16,11 @@ import (
 	"github.com/grafana/terraform-provider-grafana/internal/common"
 )
 
+const (
+	checkDefaultTimeout          = 3000
+	checkMultiHTTPDefaultTimeout = 5000
+)
+
 var (
 
 	// Set variables for schemas used in multiple fields and/or used to transform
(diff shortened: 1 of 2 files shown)


peterschretlen:
Nice, thanks @julienduchesne 🙌

peterschretlen (approved):
LGTM, thanks for fixing this!  We were going to have the same problem with the upcoming introduction of k6 "scripted" checks,  so this fix is very timely 🙏

PR #1350, opened 2024-02-14
SSO: add custom fields to OAuth2 settings

dmihai:
This PR adds custom fields to the `oauth2_settings` field in `grafana_sso_settings` terraform resource. The custom fields must not have the same name with a field already defined in `oauth2_settings`.

With custom fields introduced, I noticed a strange behaviour when re-applying a resource that has custom fields. On the second apply, `d.Get(settingsKey)` returns 2 items (the first one is empty). This doesn't reproduce for resources that don't contain any custom fields. I made a hacky fix in the `getSettingsFromResourceData` func, please let me know how to improve it.


internal/resources/grafana/resource_sso_settings_test.go
```diff
@@ -167,3 +182,16 @@ const testConfigWithManySettings = `resource "grafana_sso_settings" "sso_setting
...
 }`
+
+const testConfigWithInvalidCustomField = `resource "grafana_sso_settings" "sso_settings" {
+  provider_name = "gitlab"
+  oauth2_settings {
+    client_id     = "first_gitlab_client_id"
+    client_secret = "first_gitlab_client_secret"
+    auth_url      = "https://gitlab.com/oauth/authorize"
+    token_url     = "https://gitlab.com/oauth/token"
+	custom        = {
```
mgyongyosi: nit:
```suggestion
    custom        = {
```

mgyongyosi (approved):

internal/resources/grafana/resource_sso_settings_test.go
dmihai: Fixed, thanks!

internal/resources/grafana/resource_sso_settings.go
```diff
@@ -244,6 +245,14 @@ var oauth2SettingsSchema = &schema.Resource{
 			Optional:    true,
 			Description: "String list of Team Ids. If set, the user must be a member of one of the given teams to log in. If you configure team_ids, you must also configure teams_url and team_ids_attribute_path.",
 		},
+		customFieldsKey: {
+			Type:        schema.TypeMap,
+			Optional:    true,
+			Description: "Custom fields to configure for OAuth2. Refer to the Grafana docs for a list of supported custom fields.",
```
julienduchesne: A link would be nice

internal/resources/grafana/resource_sso_settings.go
```diff
@@ -271,6 +280,12 @@ func ReadSSOSettings(ctx context.Context, d *schema.ResourceData, meta interface
 	}
 
 	settingsSnake := make(map[string]any)
+
+	customFieldsFromTfState := make(map[string]any)
```
julienduchesne, line 284: Are the field not all returned because they're sensitive?

internal/resources/grafana/resource_sso_settings.go
dmihai: Added an example of such a custom field and its link. I don't think we have a list of all custom fields I can link to.

internal/resources/grafana/resource_sso_settings.go
dmihai: All custom fields should be returned by the API, but the concern here is that custom fields are not separated from the main fields in the API response. Also the API might return other fields that we want to ignore and not consider them custom fields by default. That's why I wanted to check the existing custom fields from the terraform state and match them against the ones retrieved from the API.

julienduchesne: hmm ok gotcha. So only what's been written by TF is managed. Are there specific fields that we want to ignore? As it is, the read functionality is not great for a TF resource

julienduchesne (approved):
LGTM, I would perhaps test it a bit more if you keep the targeted read as it is

Currently, there is no test that checks that it is actually being sent to the API. Usually, that is checked by the import test (but not you can't do an import test because your custom fields are not read). 

To be sure that it's OK from both sides (TF and API), you'd need two things: 
- Test that the custom fields are set with `resource.TestCheckResourceAttr`
- Add another test function that reads the object directly from the API and checks that the custom fields are set. A test function that uses the client directly

internal/resources/grafana/resource_sso_settings.go
dmihai: No, unfortunately there aren't any specific fields we want to ignore from the API because the API considers the settings fields as an opaque `map[string]any`.

PR #1353, opened 2024-02-16
SSO: additional acc test for custom fields

dmihai:
This test addresses the comment I received on my previous PR: https://github.com/grafana/terraform-provider-grafana/pull/1350#pullrequestreview-1883444259.


internal/resources/grafana/resource_sso_settings_test.go
```diff
@@ -177,6 +239,18 @@ func testConfigForProvider(provider string, prefix string) string {
 }`, prefix, provider, urls)
 }
 
+const testConfigWithCustomFields = `resource "grafana_sso_settings" "sso_settings" {
+  provider_name = "github"
+  oauth2_settings {
+    client_id     = "client_id"
+    client_secret = "client_secret"
+    custom = {
+      custom_field = "custom1"
```
julienduchesne, line 249: hmm, can users use non snake case here? Someone is going to try it for sure and it should work IMO. Snake Case -> translate, Camel case -> do nothing

julienduchesne (approved):
Awesome! Thanks for that. Just one question about casing

internal/resources/grafana/resource_sso_settings_test.go
dmihai: Good idea, I added a custom field using camelCase. I had to tweak the read function a bit for it to work.

Issue #769, opened 2023-01-09
Contact point can't be deleted because is currently used by a notification policy, 500 error on TF apply


hkailantzis:
### Terraform Version
* Terraform: latest
* Terraform Grafana Provider: v1.32.0
* Grafana: 9.3.1

### Affected Resource(s)
Please list the resources as a list, for example:
* grafana_contact_point
* grafana_notification_policy

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
resource "grafana_contact_point" "contact_point" {
  
  for_each = nonsensitive(var.contact_points)

  name = each.value.name

  dynamic "opsgenie" {
    for_each = each.value.opsgenies == null ? tolist([]) : each.value.opsgenies

    content {
      api_key           = var.opsgenie_api_key
      url               = opsgenie.value.url
      override_priority = true
      send_tags_as      = "tags"
      auto_close        = true
    }
  }

  dynamic "email" {
    for_each = each.value.emails == null ? tolist([]) : each.value.emails

    content {
      addresses    = email.value.email_addresses
      single_email = email.value.single_email
    }
  }
}

resource "grafana_notification_policy" "notification_policy" {
  group_by      = ["alertname", "datasource"]
  contact_point = grafana_contact_point.contact_point["opsgenie"].name

  dynamic "policy" {
    for_each = grafana_contact_point.contact_point

    content {
      contact_point = policy.value.name
      continue      = true
      group_by      = []
    }
  }
}

**test.tfvars**

contact_points = {
  opsgenie = {
    name = "OpsgenieTEST"
    opsgenies = [
      {
        url = "opsgenie_url"
      }
    ]
  }
  email = {
    name = "team-email"
    emails = [
      {
        email_addresses = ["team-email@example.com"]
        single_email    = true
      }
    ]
  }
}
```

When trying to omit the email block, TF plan shows the following: 

```
- resource "grafana_contact_point" "contact_point" {
      - id   = "xxxxxxx" -> null
      - name = "team-email" -> null
      - email {
          - addresses               = [
              - "team-email@example.com",
            ] -> null
          - disable_resolve_message = false -> null
          - settings                = (sensitive value)
          - single_email            = true -> null
          - uid                     = "xxxxxxx" -> null
        }
    }
  # module.generic_contact_points.grafana_notification_policy.notification_policy will be updated in-place
  ~ resource "grafana_notification_policy" "notification_policy" {
[...]

PR #1361, opened 2024-02-21
Fix contact point notifier update

julienduchesne:
Helps https://github.com/grafana/terraform-provider-grafana/issues/769, not a full fix though (a contact point should always be deletable on its own)

When a contact point gets all of its notifiers deleted, it's functionally deleted in the API. 
This shouldn't happen when updating, so in order to prevent that, we can add new notifiers at the beginning of the update and delete the old ones only at the end of the process 

Doing things in that order ensures that there will always be one notifier active since there's a TF resource validation that there should be at least one notifier


JacobsonMT (approved):
LGTM :rocket:

PR #1380, opened 2024-02-26
POC: Standardize and verify ID formats

julienduchesne:
Currently, resource ID formats are all over the place. The separator is mostly `:`, but sometimes it's `_`, `/` or `;`

In this PR, I create a helper struct to manage IDs:
- It defines a default separator with retrocompatibility with the old formats. The old separators should be deprecated on the next major version
- It provides validation for the number of fields provided. Currently, it panics in most resources if you provide not enough fields
- It does auto-generation of the `import.sh` files

If this looks good, I can expand this to all resources and then validate that the ID is set this way for all resources. Providing consistent import and ID management for all resources!

### DIFF
--- a/internal/resources/cloud/resource_cloud_plugin_test.go
+++ b/internal/resources/cloud/resource_cloud_plugin_test.go
@@ -3,6 +3,7 @@ package cloud_test
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"testing"
 
 	"github.com/grafana/grafana-com-public-clients/go/gcom"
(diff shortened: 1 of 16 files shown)


internal/resources/cloud/resource_cloud_plugin_test.go
```diff
@@ -40,6 +41,27 @@ func TestAccResourcePluginInstallation(t *testing.T) {
...
+				ImportStateId:     fmt.Sprintf("%s:%s", stackSlug, pluginSlug),
+			},
+			// Test import with invalid ID
+			{
+				ResourceName:      "grafana_cloud_plugin_installation.test-installation",
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateId:     "noseparator",
+				ExpectError:       regexp.MustCompile("Error: id \"noseparator\" does not match expected format. Should be in the format: stackSlug:pluginSlug"),
+			},
```
julienduchesne, line 64: The helper is essentially tested here

K-Phoen (approved):

PR #1386, opened 2024-02-27
SM: Allow deleting probes that are in use

### LINKED ISSUE #534: Nested dependencies when deploying both probes and checks via Terraform
### Terraform Version
* Terraform: 1.2.0
* Terraform Grafana Provider: 1.24.0
* Grafana: 8.5.7

### Affected Resource(s)
* grafana_synthetic_monitoring_check
* grafana_synthetic_monitoring_probe

```hcl
resource "grafana_synthetic_monitoring_probe" "private" {
  name      = "a_name"
  latitude  = 0
  longitude = 0 
  region    = "APAC"
  public    = false
}

resource "grafana_synthetic_monitoring_check" "this" {
  job               = "AJob"
  target            = "https://google.com"
  enabled           = true
  probes            = grafana_synthetic_monitoring_probe.private.id

  settings {
    http {
      method              = "GET"
      fail_if_not_ssl     = false
      fail_if_ssl         = false
      ip_version          = "V4"
      no_follow_redirects = false
      valid_status_codes = [
        200,
      ]

      tls_config {}
    }
  }
}
```

### Debug Output

**Will provide if needed**

### Panic Output

**NONE**

### Expected Behavior
What should have happened?

The above Terraform is correct behavior so there is no issue with it. However if we follow this flow we will encounter the nested dependencies:

1. Change the name of the probe
2. Terraform will pick up the change and try to re-deploy the probe
3. New probe's id will be registered
4. Old probe will be deleted
5. Checks depends on probe to finish deleting and creating, but probe can't be deleted if the checks still depends on it.
    
   ```
   Error: probe delete request: status="403 Forbidden", msg="delete not allowed, probe <id> has checks assigned", err="delete not allowed"
   ```

### Actual Behavior
What actually happened?

Like the above but when changing name or force probe to be re-created.

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
Are there anything atypical about your accounts that we should know? For example: Running in EC2 Classic? Custom version of OpenStack? Tight ACLs?

**None**

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:

**None**

julienduchesne:
Closes https://github.com/grafana/terraform-provider-grafana/issues/534

Removes the probe from checks before deleting it
Checks must still have a probe at all times, but now, in the case where all probes are deleted, there will also be a relevant error message.

### DIFF
--- a/internal/resources/syntheticmonitoring/resource_probe.go
+++ b/internal/resources/syntheticmonitoring/resource_probe.go
@@ -163,14 +163,31 @@ func ResourceProbeUpdate(ctx context.Context, d *schema.ResourceData, meta inter
 
 func ResourceProbeDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	c := meta.(*common.Client).SMAPI
-	var diags diag.Diagnostics
 	id, _ := strconv.ParseInt(d.Id(), 10, 64)
-	err := c.DeleteProbe(ctx, id)
+
+	// Remove the probe from any checks that use it.
+	checks, err := c.ListChecks(ctx)
 	if err != nil {
 		return diag.FromErr(err)
 	}
+	for _, check := range checks {
+		for i, probeID := range check.Probes {
+			if probeID == id {
+				check.Probes = append(check.Probes[:i], check.Probes[i+1:]...)
+				if len(check.Probes) == 0 {
+					return diag.Errorf(`could not delete probe %d. It is the only probe for check %q.
+You must also taint the check, or assign a new probe to it before deleting this probe.`, id, check.Job)
+				}
+				if _, err := c.UpdateCheck(ctx, check); err != nil {
+					return diag.Errorf(`error while deleting probe %d, failed to remove it from check %q: %s.`, id, check.Job, err)
+				}
+				break
+			}
+		}
+	}
+
 	d.SetId("")
-	return diags
+	return diag.FromErr(c.DeleteProbe(ctx, id))
 }
 
 // makeProbe populates an instance of sm.Probe. We need this for create and
--- a/internal/resources/syntheticmonitoring/resource_probe.go
+++ b/internal/resources/syntheticmonitoring/resource_probe.go
@@ -173,11 +173,11 @@ func ResourceProbeDelete(ctx context.Context, d *schema.ResourceData, meta inter
 	for _, check := range checks {
 		for i, probeID := range check.Probes {
 			if probeID == id {
-				check.Probes = append(check.Probes[:i], check.Probes[i+1:]...)
-				if len(check.Probes) == 0 {
+				if len(check.Probes) == 1 {
 					return diag.Errorf(`could not delete probe %d. It is the only probe for check %q.
 You must also taint the check, or assign a new probe to it before deleting this probe.`, id, check.Job)
 				}
+				check.Probes = append(check.Probes[:i], check.Probes[i+1:]...)
 				if _, err := c.UpdateCheck(ctx, check); err != nil {
 					return diag.Errorf(`error while deleting probe %d, failed to remove it from check %q: %s.`, id, check.Job, err)
 				}
--- a/internal/resources/syntheticmonitoring/resource_probe_test.go
+++ b/internal/resources/syntheticmonitoring/resource_probe_test.go
@@ -94,6 +94,64 @@ func TestAccResourceProbe_recreate(t *testing.T) {
 	})
 }
 
+// Test that a probe that is used in a check can be deleted or recreated
+func TestAccResourceProbe_recreateProbeUsedInCheck(t *testing.T) {
+	testutils.CheckCloudInstanceTestsEnabled(t)
+
+	randomName := acctest.RandomWithPrefix("tf")
+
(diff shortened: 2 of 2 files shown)


peterschretlen:
This looks good to me.  if I understand correctly, we do this because the delete probe API will fail if there are any checks still using it. So this seems to be the right fix.   But I will swap myself as reviewer for a BE engineer on the team for proper review @ka3de maybe you can take a look with @The-9880 ?

internal/resources/syntheticmonitoring/resource_probe_test.go
```diff
@@ -94,6 +94,64 @@ func TestAccResourceProbe_recreate(t *testing.T) {
...
+				Config: testSyntheticMonitoringProbeAndCheck(randomName, "test1"),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet("grafana_synthetic_monitoring_probe.first", "id"),
+					resource.TestCheckResourceAttrSet("grafana_synthetic_monitoring_probe.second", "id"),
+					resource.TestCheckResourceAttrSet("grafana_synthetic_monitoring_check.http", "id"),
+					resource.TestCheckResourceAttr("grafana_synthetic_monitoring_probe.first", "name", randomName+"test1-first"),
+					resource.TestCheckResourceAttr("grafana_synthetic_monitoring_probe.second", "name", randomName+"test1-second"),
+				),
+			},
+			// Change the name of the probe
```
ka3de, line 116: I'm not familiar with Terraform providers implementation, so just to make sure.. I understand this test verifies that the check that was originally associated with both test probes (_first_ and _second_), is now associated with the "new" probes after the rename. Is that correct?

internal/resources/syntheticmonitoring/resource_probe.go
```diff
@@ -163,14 +163,31 @@ func ResourceProbeUpdate(ctx context.Context, d *schema.ResourceData, meta inter
...
+	checks, err := c.ListChecks(ctx)
 	if err != nil {
 		return diag.FromErr(err)
 	}
+	for _, check := range checks {
+		for i, probeID := range check.Probes {
+			if probeID == id {
+				check.Probes = append(check.Probes[:i], check.Probes[i+1:]...)
+				if len(check.Probes) == 0 {
+					return diag.Errorf(`could not delete probe %d. It is the only probe for check %q.
```
ka3de, line 177: I understand that when a probe has to be recreated, the way Terraform works does not allow to create the new probe first and then move the checks to the new probe before deleting the previous one?

internal/resources/syntheticmonitoring/resource_probe.go
```diff
@@ -163,14 +163,31 @@ func ResourceProbeUpdate(ctx context.Context, d *schema.ResourceData, meta inter
...
+	// Remove the probe from any checks that use it.
+	checks, err := c.ListChecks(ctx)
 	if err != nil {
 		return diag.FromErr(err)
 	}
+	for _, check := range checks {
+		for i, probeID := range check.Probes {
+			if probeID == id {
+				check.Probes = append(check.Probes[:i], check.Probes[i+1:]...)
+				if len(check.Probes) == 0 {
```
ka3de: nit: We could potentially move this verification up before doing the re-slice.

ka3de (commented):
The `ResourceProbeDelete` function implementation looks good to me, in the sense that it does what it's meant to do. Other than that I have a couple of questions due to my lack of knowledge on Terraform, but overall LGTM.

internal/resources/syntheticmonitoring/resource_probe.go
julienduchesne: The problem stems from the following:
- When "tainting" a resource, Terraform can either create it first and then destroy the old one (opt-in) or the opposite (default behavior).
- The "create" function does not have any knowledge that a "delete" came beforehand, or will come afterwards
- Creating the probe first would mean creating under a temporary name to avoid conflicts. This is generally bad practice because we're dealing with as-code here and any interruption in the process would leave unmanaged things with temporary names. Not to mention that it'd be impossible for the create function to actually know it has to use a temporary name (because of the previous point)

Destroying the probe first is the default behaviour and is the one I'm trying to fix here. IMO, it's alright to have the limitation that every check needs to have at least one probe assigned, which is why I'm closing the issue. We just need to do the best effort, and document (with errors) what the best way to go forward is for the user, if it does still fail.

internal/resources/syntheticmonitoring/resource_probe_test.go
julienduchesne: It checks that we can rename used probes, whether or not it recreates them. The original issue mentioned that renaming probes was an issue

internal/resources/syntheticmonitoring/resource_probe.go
julienduchesne: Done!

internal/resources/syntheticmonitoring/resource_probe.go
ka3de: Got it. Thanks for the explanation.

ka3de (approved):

PR #1395, opened 2024-03-02
Typed resource ID helper

julienduchesne:
Each resource's ID is composed of n (>=1) string or integer elements 
We can generalize this behavior and do the parsing/formatting in the helper function 
This is extracted from https://github.com/grafana/terraform-provider-grafana/pull/1391 and is part of a push to have standardized IDs for all resources, allowing for easier generation of TF code!

### DIFF
--- a/internal/resources/cloud/resource_cloud_stack_service_account_test.go
+++ b/internal/resources/cloud/resource_cloud_stack_service_account_test.go
@@ -3,7 +3,6 @@ package cloud_test
 import (
 	"context"
 	"fmt"
-	"strings"
 	"testing"
 
 	"github.com/grafana/grafana-com-public-clients/go/gcom"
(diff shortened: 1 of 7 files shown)


internal/resources/cloud/resource_cloud_stack_service_account_test.go
```diff
@@ -87,24 +86,14 @@ func testAccGrafanaAuthCheckServiceAccounts(stack *gcom.FormattedApiInstance, ex
 			return fmt.Errorf("failed to get service accounts: %w", err)
 		}
 
-		var foundSAs []string
```
julienduchesne, line 90: This test is broken from a recent change to Grafana Cloud. Some service accounts are auto-provisioned into new stacks

malcolmholmes (approved):

PR #1397, opened 2024-03-03
Self-contained ML package

julienduchesne:
Just like #1393 and #1396

- Makes the resource functions private (exposed via a single map attribute) rather than each resource individually
- Puts the client validation in the ML package


sd2k (approved):
Looks fine to me!

PR #1407, opened 2024-03-06
Alerting: Add support for notification settings for alert rules

yuri-tceretian:
This PR adds support for a new field in the alert rule definition, `notification_settings`. This field has been supported since Grafana 10.4.

NOTE:
1. As of now, it requires the feature flag `alertingSimplifiedRouting` to be enabled. Otherwise, the field will be accepted and stored but not used by Grafana.

Related to https://github.com/grafana/grafana/pull/81011


julienduchesne (approved):
Cool feature. LGTM!

PR #1415, opened 2024-03-12
Run tests on Grafana v11

julienduchesne:
Update other versions, drop v8 support


julienduchesne:
Gonna have to wait for 10.4.1, there's a bug that panics in 10.4.0

julienduchesne:
Still broken in v10.4.1 it seems

spinillos (approved):

PR #1422, opened 2024-03-14
ML: each Jobs test creates a datasource to depend on

gerboland:
Create a datasource and use it for each ML job. Avoids depending on magic datasource id/uids.

### DIFF
--- a/internal/resources/machinelearning/resource_job_test.go
+++ b/internal/resources/machinelearning/resource_job_test.go
@@ -18,15 +18,21 @@ func TestAccResourceJob(t *testing.T) {
 	testutils.CheckCloudInstanceTestsEnabled(t)
 
 	randomName := acctest.RandomWithPrefix("Test Job")
+	defaultIDRegexp := regexp.MustCompile(`^[a-zA-Z0-9-_]+$`)
 
 	var job mlapi.Job
 	resource.ParallelTest(t, resource.TestCase{
 		ProtoV5ProviderFactories: testutils.ProtoV5ProviderFactories,
-		CheckDestroy:             testAccMLJobCheckDestroy(&job),
+		CheckDestroy: resource.ComposeTestCheckFunc(
+			testAccMLJobCheckDestroy(&job),
+			testAccDatasourceCheckDestroy(),
+		),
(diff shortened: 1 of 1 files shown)


julienduchesne:
> The datasource ID of 10 is invalid, causing the ML job to fail with an error that impacts our SLO.
> 
> This changes it to 11 which is valid.

Can we provision the datasource in the test rather than using a magic number?

gerboland:
> Can we provision the datasource in the test rather than using a magic number?

Not a bad idea, I'll look into that. Would you know if there's prior art for that?

julienduchesne:
> > Can we provision the datasource in the test rather than using a magic number?
> 
> Not a bad idea, I'll look into that. Would you know if there's prior art for that?

You can create a datasource like this and reference it: https://github.com/grafana/terraform-provider-grafana/blob/main/internal/resources/grafana/resource_data_source_permission_test.go#L69-L82

gerboland:
Thanks, I'll get that going

internal/resources/machinelearning/resource_job_test.go
```diff
@@ -18,53 +18,61 @@ func TestAccResourceJob(t *testing.T) {
...
+			testAccMLJobCheckDestroy(&job),
+			testAccDatasourceCheckDestroy(),
+		),
 		Steps: []resource.TestStep{
 			{
-				Config: testutils.TestAccExampleWithReplace(t, "resources/grafana_machine_learning_job/job.tf", map[string]string{
+				// Note for the reader: these tests construct a datasource & a job, where the Job's datasource id is set by terraform.
+				Config: machinelearningDatasource + testutils.TestAccExampleWithReplace(t, "resources/grafana_machine_learning_job/job.tf", map[string]string{
 					"Test Job": randomName,
+					"= 10":     "= split(\":\", grafana_data_source.foo.id)[1]", // as it of form ^(0|1):[a-zA-Z0-9-_]+$, we only use later part
```
julienduchesne, line 35: It'd be good if ML supported uids as IDs are being deprecated but that's for another Pr

julienduchesne (approved):

PR #1432, opened 2024-03-21
Remove global resources var

julienduchesne:
The `provider` package now has a function that will expose all the resources from all subcategories of the provider 

This list is used to:
- Used in the import docs generation command
- Sent to both provider servers
- Used in the tf code generation command

***Review only the first commit, I merged https://github.com/grafana/terraform-provider-grafana/pull/1433, forgetting that it depended on this PR here 🤦***


gerboland (approved):
+1 from ML

PR #1441, opened 2024-03-22
Add `grafana_data_source_config` resource

### LINKED ISSUE #1408: [Feature Request]: grafana_data_source data allow non-existing datasource
### Feature Request

Currently the `data "grafana_data_source"` returns error 404 if the datasource doesn't exist.
This is problematic for us since the tempo and loki datasource require one another on some feature flags.

Would it be possible to allow `data "grafana_data_source"` to work with non-existing datasources and don't let terraform fail if the datasource doesn't exist?

miljanic:
This PR introduces the `grafana_data_source_config` resource. It will separate the creation of the Data Source from its configuration. This is especially useful in cases where multiple Data Sources configurations depend on each other, thus creating a circular dependency.

Closes #1408


internal/resources/grafana/resource_data_source_config.go
```diff
@@ -0,0 +1,195 @@
...
+				},
+				DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
+					if oldValue == "{}" && newValue == "" {
+						return true
+					}
+					return common.SuppressEquivalentJSONDiffs(k, oldValue, newValue, d)
+				},
+			},
+		},
+	}
```
julienduchesne: I assume these are exactly the same as in the resource_data_source.go file. They could be extracted to common vars

docs/resources/data_source_config.md
```diff
@@ -0,0 +1,80 @@
...
+  json_data_encoded = jsonencode({
+    tracesToLogsV2 = {
+      customQuery     = true
+      datasourceUid   = grafana_data_source.loki.uid
+      filterBySpanID  = false
+      filterByTraceID = false
+      query           = "|=\"$${__trace.traceId}\" | json"
+    }
+  })
+}
```
julienduchesne: I'm going to assume that these docs were written manually because this example is not in a separate file. It should be. The example will also be tested automatically at that point.

The docs should be generated with `go generate ./...`

internal/resources/grafana/resource_data_source_config.go
```diff
@@ -0,0 +1,195 @@
...
+		URL:             fetchedDataSource.URL,
+		User:            fetchedDataSource.User,
+		WithCredentials: fetchedDataSource.WithCredentials,
+	}
+	_, err = client.Datasources.UpdateDataSourceByUID(dataSourceUID, &body)
+
+	return diag.FromErr(err)
+}
+
+func readDatasourceConfig(d *schema.ResourceData, dataSource *models.DataSource) diag.Diagnostics {
```
julienduchesne: This code should be re-used in the main datasource code

julienduchesne:
Nice, thanks for picking this up. Some things need fixing but this is a good start

miljanic:
@julienduchesne I fixed all comments but the one related to the common code extraction. I understand its idea, but couldn't find an idiomatic way to do so. I'm still happy to do it if you give me an example.

julienduchesne:
> @julienduchesne I fixed all comments but the one related to the common code extraction. I understand its idea, but couldn't find an idiomatic way to do so. I'm still happy to do it if you give me an example.

I can take a look 👀

julienduchesne (approved):
Added a test, fixed up a few things and make a bunch of code common between the regular datasource resource. LGTM now! Thank you!

PR #1460, opened 2024-04-03
Docs: update docs for local development with Grafana

dmihai:
After setting up terraform to work with my local `terraform-provider-grafana` I found the following missing from the docs:
- `.terraformrc` file has to be in the user directory
- `terraform plan/apply` will use the local binary regardless of the version you specify for the Grafana provider

Updated the docs accordingly.


julienduchesne (approved):
thanks

PR #1461, opened 2024-04-03
Expose client creation and provider config

julienduchesne:
Also some renaming of things that annoyed me
This will be used in https://github.com/grafana/terraform-provider-grafana/pull/1436 
This is a no-op change but reduces the diff on the PR above


K-Phoen (approved):
Bite-sized PR, very nice :)

PR #1464, opened 2024-04-04
Fix `grafana_cloud_stack_service_account` resource to support creating service accounts without a basic role

### LINKED ISSUE #1463: [Bug]: Cannot create `grafana_cloud_stack_service_account` without basic role
### Terraform Version

v1.7.5

v2.14.3

### Grafana Version

Grafana Cloud

### Affected Resource(s)

* grafana_cloud_stack_service_account

```hcl
resource "grafana_cloud_stack_service_account" "test_service_account" {
  stack_slug = "my-org"
  name       = "test-service-account"
}
```

### Expected Behavior

Creation should be successful

### Actual Behavior

The following error is received:
```
 Error: 409 Conflict
│
│   with grafana_cloud_stack_service_account.test_service_account,
│   on test.tf line 1, in resource "grafana_cloud_stack_service_account" "test_service_account":
│  1: resource "grafana_cloud_stack_service_account" "test_service_account" {
```

### Steps to Reproduce

1. Run terraform apply.
2. Type `yes` to accept and apply the plan.

### Important Factoids

The GCOM api for creating stack service accounts uses the following URL: `/instances/{instanceId}/api/serviceaccounts`
If we do not want to assign a role to the service account, the `role` property must be set to `None`. If the `role` property is not set, the request will be rejected as it is required.

The problem is that the provider assumes role to be optional, and if set, should be one of `Viewer`, `Editor` or `Admin` per: https://github.com/grafana/terraform-provider-grafana/blob/main/internal/resources/cloud/resource_cloud_stack_service_account.go#L59

Therefore:
- If role is not set, it passes terraform validation, but the API rejects it.
- If role is set to `None`, terraform validation fails.

### References

_No response_

F21:
Closes #1463

### DIFF
--- a/docs/resources/cloud_stack_service_account.md
+++ b/docs/resources/cloud_stack_service_account.md
@@ -41,12 +41,12 @@ resource "grafana_cloud_stack_service_account" "cloud_sa" {
 ### Required
 
 - `name` (String) The name of the service account.
+- `role` (String) The basic role of the service account in the organization.
 - `stack_slug` (String)
 
 ### Optional
 
 - `is_disabled` (Boolean) The disabled status for the service account. Defaults to `false`.
-- `role` (String) The basic role of the service account in the organization.
 
 ### Read-Only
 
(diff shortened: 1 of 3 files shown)


julienduchesne (approved):
LGTM! Thanks!

PR #1468, opened 2024-04-04
Datasource: Use UID instead of numerical ID

julienduchesne:
The numerical ID is deprecated since v9. Let's remove it on next major version


Duologic (approved):

gerboland (approved):
Looks good from the ML side

PR #1495, opened 2024-04-16
`grafana_service_account`: Require role attribute

### LINKED ISSUE #1493: [Bug]: 500 Internal Server Error when `grafana_service_account.role` is omitted
### Terraform Version

1.6.2

2.17.0

### Grafana Version

11 (Grafana Cloud)

### Affected Resource(s)

* grafana_service_account

```hcl
resource "grafana_service_account" "my_service_account" {
  name = "test"
  lifecycle {
    // ignore_changes = [role] // avoids HTTP 500 error
  }
}
```

### Expected Behavior

No change

### Actual Behavior

Error: status: 500, body: {"error":"Server Error","message":"Internal Server Error - Please try again. If the issue persists please open a support ticket that notes the time in UTC."}

### Steps to Reproduce

`terraform apply` twice, the second time fails

### Important Factoids

On the second `terraform apply`, I would expect to see "None" or "No Basic Role" instead of "Viewer" 
```
resource "grafana_service_account" "annotations" {
        id          = "1:155"
        name        = "test"
      - role        = "Viewer" -> null
        # (2 unchanged attributes hidden)
    }
```

### References

_No response_

julienduchesne:
Closes https://github.com/grafana/terraform-provider-grafana/issues/1493 
The other option would've been to default to `None` BUT the `None` role is only available from 10.3, so this would not work for most Grafana versions

### DIFF
--- a/docs/resources/service_account.md
+++ b/docs/resources/service_account.md
@@ -30,12 +30,12 @@ resource "grafana_service_account" "admin" {
 ### Required
 
 - `name` (String) The name of the service account.
+- `role` (String) The basic role of the service account in the organization.
 
 ### Optional
 
 - `is_disabled` (Boolean) The disabled status for the service account. Defaults to `false`.
 - `org_id` (String) The Organization ID. If not set, the Org ID defined in the provider block will be used.
-- `role` (String) The basic role of the service account in the organization.
 
 ### Read-Only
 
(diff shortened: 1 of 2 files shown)


K-Phoen (approved):

hertzsprung:
This is a backwards incompatible change, will it be released with a major version bump?

julienduchesne:
> This is a backwards incompatible change, will it be released with a major version bump?

Not specifying it gave a 500. Requiring a role is not backwards incompatible IMO

PR #1512, opened 2024-04-25
Users: ignore case when diffing email and login

IevaVasiljeva:
**What**

This PR makes user login and email diffing case insensitive.

**Why**

We've changed the implementation on Grafana's side to be case insensitive when resolving user emails and logins.

This already causes a diff for newly created users if upper case characters are used for user logins or emails, and will cause an even larger diff when https://github.com/grafana/grafana/pull/86359 is merged, as we will be migrating all existing user logins and emails to lower case.

**How to test**

* use Grafana TF provider without this change;
* create a user with upper case characters in login/email;
* run `terraform apply` again and observe the diff;
* now use Grafana TF provider with this change and run `terraform apply` - no diff should appear.

User resource used:
```
resource "grafana_user" "test_user" {
  email    = "NEW_terraform_uSEr@test.com"
  login    = "nEW_terRAForm_user"
  password = "password"
}
```

Diff observed:
```
  # grafana_user.test_user will be updated in-place
  ~ resource "grafana_user" "test_user" {
      ~ email    = "new_terraform_user@test.com" -> "NEW_terraform_uSEr@test.com"
        id       = "15"
      ~ login    = "new_terraform_user@test.com" -> "nEW_terRAForm_user@test.com"
        # (3 unchanged attributes hidden)
    }
```


Jguer (approved):

internal/resources/grafana/resource_user_test.go
```diff
@@ -96,6 +108,16 @@ resource "grafana_user" "test" {
 }
 `
 
+const testAccUserConfig_mixedCase = `
```
eleijonmarck, line 111: could potentially add a case for all uppercase

only for all coverage

eleijonmarck (approved):

internal/resources/grafana/resource_user_test.go
IevaVasiljeva: Oh, sorry, didn't see this comment before merging. I think in this case it doesn't matter enough to raise another PR just for that.

PR #1513, opened 2024-04-25
Update oncall_schedule.md

gthompson280:
The Schedule HTTP API docs (https://grafana.com/docs/oncall/latest/oncall-api-reference/schedules/) that the oncall_schedule resource uses specify "web" as a valid type value.

However, this isn't an available option for the TF resource due to drift or conflict issues. It would be good to clarify the valid values ("ical" and "calendar") since this differs from the Schedule HTTP API docs.


julienduchesne (approved):

matiasb (approved):
Thanks :+1:

PR #1516, opened 2024-04-27
Deprecate global `org_id` config

julienduchesne:
This attribute is not useful because:
- The attribute was already added on each resource, so it doesn't remove any functionality
- It's irrelevant to all parts of the provider other than Grafana resources, possibly causing confusing since there's also an org concept on cloud resources
- Even for Grafana, it does not apply to all resources. For example, `users` are not org-scoped. By setting it on a resource level, this is much better defined
- When using it globally, it prevents looping on multiple orgs. Ex: Create dashboards for a list of orgs. Providers cannot be created dynamically


K-Phoen (approved):

ishworg:
This PR isn't a deprecation. It's a removal. I would encourage the authors in the future to drop it in the `ProviderConfig` first. Then remove it *after* announcing the deprecation to help users migrate(users have varying capacity)...

PR #1518, opened 2024-04-27
Deprecate old ID separators

julienduchesne:
In previous versions, all legacy IDs were deprecated in favor of a common `:` 

Anyone running the latest version should have had all their resources migrated to the new separator so we can remove all the legacy ones 

This PR will be merged in the next major version


K-Phoen (approved):
:rocket:

PR #1529, opened 2024-05-02
Remove all deprecated attributes and resources

julienduchesne:
Preparing the next major release, this PR removes deprecated attributes and resources 
Their use was also replaced with their non-deprecated alternative

**Note: this is a pretty big PR, but there's nothing really crazy in there. Mostly removing some attributes and where they are plugged in. The one thing of note is a workaround I had to do in the annotations resource: https://github.com/grafana/terraform-provider-grafana/pull/1529/files#diff-2557a290b338e0264c77cd6aead40eb6907ee01698f66757b51e281a72df841aR134-R154**


gerboland (approved):
Looks good from the ML-team! Thank you

internal/resources/oncall/data_source_action.go
joeyorlando, line 1: agree with getting rid of this, just curious if the next release be a major version bump? (as this feels like a breaking change)

joeyorlando (approved):
LGTM from OnCall side of things. Just one question

internal/resources/oncall/data_source_action.go
julienduchesne: Yup. That will be in the 3.0.0

julienduchesne: I'll be merging all of the `v3` milestone PRs at the same time. When I'm ready to do the major release

eleijonmarck:
@julienduchesne is there a estimate as yo when the v3 of terraform will be out?

julienduchesne:
Monday, I think

PR #1530, opened 2024-05-02
Add data source for oncall integration

karl-johan-grahn:
https://github.com/stakater/grafana-cloud-ansible-operator creates and managed oncall integrations, however integrations don't exist as data sources in the provider so cannot be referenced as such when creating oncall routes for example


internal/resources/oncall/data_source_integrations_test.go
```diff
@@ -0,0 +1,35 @@
...
+				Config:      testAccDataSourceIntegrationConfig(integrationID),
+				ExpectError: regexp.MustCompile(`couldn't find an integration`),
+			},
+		},
+	})
+}
+
+func testAccDataSourceIntegrationConfig(integrationID string) string {
+	return fmt.Sprintf(`
+data "grafana_oncall_integration" "test-acc-integration" {
```
julienduchesne: I think the test should create an integration and then fetch it back

examples/resources/grafana_oncall_route/resource.tf
```diff
@@ -2,13 +2,19 @@ data "grafana_oncall_slack_channel" "example_slack_channel" {
   name = "example_slack_channel"
 }
 
+data "grafana_oncall_integration" "example_integration_data_source" {
+  id = "CABCDEFGHI321"
+}
+
```
julienduchesne: hmm, I'd remove this one. The link between integration and route is already in this example file

internal/resources/oncall/data_source_integrations.go
```diff
@@ -0,0 +1,49 @@
...
+		ReadContext: withClient[schema.ReadContextFunc](dataSourceIntegrationRead),
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The integration ID.",
+			},
+			"name": {
+				Type:        schema.TypeString,
+				Computed:    true,
```
julienduchesne: It'd be nice if that could be specified as well. Input just the name -> get the ID and vice-versa

internal/resources/oncall/data_source_integrations.go
```diff
@@ -0,0 +1,49 @@
+package oncall
```
julienduchesne: This file should be named `data_source_integration.go`

examples/resources/grafana_oncall_route/resource.tf
karl-johan-grahn: Alright removed it

internal/resources/oncall/data_source_integrations.go
karl-johan-grahn: Renamed them

internal/resources/oncall/data_source_integrations.go
karl-johan-grahn: The API https://grafana.com/docs/oncall/latest/oncall-api-reference/integrations/#get-integration just accepts an ID so I don't know how we would go about with that, would need advice if that would be a strict requirement

internal/resources/oncall/data_source_integrations_test.go
karl-johan-grahn: I've made an attempt

karl-johan-grahn:
Please run tests and review the tests

karl-johan-grahn:
What about the docs, should I generate them and commit the generated files? Not clear if the pipeline will do that or if it's on me

julienduchesne:
Merged through https://github.com/grafana/terraform-provider-grafana/pull/1532. Thank you!

PR #1541, opened 2024-05-07
APIkey: Creating a apikey return a error for the provider

eleijonmarck:
**why**
We are removing the ability to create apikeys from grafanas api in favor of service accounts and service account tokens.

**what**
This removes the call to grafana and returns an error saying it is deprecated.


IevaVasiljeva (approved):

julienduchesne:
This resource is going away in v3, removed here: https://github.com/grafana/terraform-provider-grafana/pull/1529. I will close this because it's a breaking change and I need to do another v2 release before this

PR #1553, opened 2024-05-09
feat(slo): support folder field

### LINKED ISSUE #1527: [Feature Request]: Allow defining Folder for SLOs
### Feature Request

Could the provider be updated to support defining the Folder the SLO dashboard will live in? I see it's currently available in the API Spec https://grafana.com/docs/grafana-cloud/alerting-and-irm/slo/api/slo-api.yaml

Thanks!

inikotoran:
### Summary

In order for users to be able to define folder for SLO resource, we need to upgrade github.com/grafana/slo-openapi-client to latest version. This mainly require changing the usage of `Slo` struct to `SloV00Slo`. The new struct is generated by OpenAPI spec and since the API request function expect the versioned struct, it would be easier to use the new struct directly.

Related to the folder input, the new field is called `folder_uid` and it's optional.

#### Local

The following tests has been conducted locally and applied in Grafana Cloud Enterprise account.
- [x] Can create SLO without defining `folder_uid`
- [x] Can create SLO with `folder_uid`
- [x] Throws error when folder UID is invalid


internal/resources/slo/resource_slo.go
```diff
@@ -51,6 +51,12 @@ Resource manages Grafana SLOs.
 				Description:  `Description is a free-text field that can provide more context to an SLO.`,
 				ValidateFunc: validation.StringLenBetween(0, 1024),
 			},
+			"folder": {
```
julienduchesne:
```suggestion
			"folder_uid": {
```

Can we call it that way to make it clear that we expect the `grafana_folder.<>.uid` attribute?

inikotoran: hi @julienduchesne thank your for your review. It make sense, I've updated it here: 7bb5292da7c52d4fbef79a4f1359ef7dadb48ad3

inikotoran: going to fix the test locally, edit: fixed here: ea5d82571ede312142a444b7686784c8fa967098

inikotoran: @julienduchesne is it mandatory to add tests for the `folder_uid` input? If that's the case, I'm thinking to follow [resource_dashboard_test](https://github.com/grafana/terraform-provider-grafana/blob/main/internal/resources/grafana/resource_dashboard_test.go#L316-L336) implementation by creating folder first

julienduchesne:
👋 For some reason, the tests don't pass anymore when these changes are applied. I will investigate more tomorrow 👀

julienduchesne:
LGTM! Thanks!

julienduchesne (approved):

PR #1561, opened 2024-05-13
Update synthetic monitoring frequency value validation

tomaszgawliksa:
**Summary**
This aims to update changes introduced in this PR: https://github.com/grafana/terraform-provider-grafana/pull/409/files#diff-2881232238e1b6d6b1e702da359aa74b21307f47f450d817a97f1fe9cf10f74bR480 

When using the Grafana UI (Grafana v11.1.0-69950 (a4bb4c8400)) to create a new Synthetic Check, the range for the frequency is between 10 seconds and 1 hour. 

The terraform validation function range is between 1 second and 120 seconds. 
image
image

It is possible to create a check with Terraform and update it in the UI:
```
  # grafana_synthetic_monitoring_check.uptime_checks["test"] will be updated in-place
  ~ resource "grafana_synthetic_monitoring_check" "this" {
      ~ frequency          = 3600000 -> 60000
```

### DIFF
--- a/internal/resources/syntheticmonitoring/resource_check.go
+++ b/internal/resources/syntheticmonitoring/resource_check.go
@@ -676,7 +676,7 @@ multiple checks for a single endpoint to check different capabilities.
 				Type:         schema.TypeInt,
 				Optional:     true,
 				Default:      60000,
-				ValidateFunc: validation.IntBetween(10000, 3600000),
+				ValidateFunc: validation.IntBetween(1000, 3600000),
 			},
 			"timeout": {
 				Description: "Specifies the maximum running time for the check in milliseconds. " +
--- a/internal/resources/syntheticmonitoring/resource_check.go
+++ b/internal/resources/syntheticmonitoring/resource_check.go
@@ -672,7 +672,7 @@ multiple checks for a single endpoint to check different capabilities.
 			},
 			"frequency": {
 				Description: "How often the check runs in milliseconds (the value is not truly a \"frequency\" but a \"period\"). " +
-					"The minimum acceptable value is 10 second (10000 ms), and the maximum is 1 hour (3600000 ms).",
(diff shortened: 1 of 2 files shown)


julienduchesne (approved):
You're right. It does crash with 1000. LGTM!

peterschretlen:
I looked at frequencies in production and we do have some generated through terraform that are < 10s.  I think to maintain compatibility with those checks we should keep the lower bound at 1s,  otherwise this becomes a breaking change.

julienduchesne:
> I looked at frequencies in production and we do have some generated through terraform that are < 10s. I think to maintain compatibility with those checks we should keep the lower bound at 1s, otherwise this becomes a breaking change.

Oh, good to know. I assumed that since it crashed, it meant that <10s frequencies weren't possible. Let's keep the min at 1s then 👍

internal/resources/syntheticmonitoring/resource_check.go
```diff
@@ -672,11 +672,11 @@ multiple checks for a single endpoint to check different capabilities.
 			},
 			"frequency": {
 				Description: "How often the check runs in milliseconds (the value is not truly a \"frequency\" but a \"period\"). " +
-					"The minimum acceptable value is 1 second (1000 ms), and the maximum is 120 seconds (120000 ms).",
+					"The minimum acceptable value is 10 second (10000 ms), and the maximum is 1 hour (3600000 ms).",
 				Type:         schema.TypeInt,
 				Optional:     true,
 				Default:      60000,
-				ValidateFunc: validation.IntBetween(1000, 120000),
+				ValidateFunc: validation.IntBetween(10000, 3600000),
```
julienduchesne:
```suggestion
				ValidateFunc: validation.IntBetween(1000, 3600000),
```

keep at 1s for retrocompatibility

julienduchesne (changes requested):

julienduchesne (approved):

PR #1565, opened 2024-05-14
Update auth key

jdbaldry:
Do these other cases need to be updated?

```
examples/provider/provider-cloud.tf
38     auth = grafana_cloud_stack_service_account_token.cloud_sa.key

examples/provider/provider.tf
3      auth = var.grafana_auth

internal/resources/cloud/resource_cloud_stack_service_account_test.go
83   					auth = grafana_cloud_stack_service_account_token.management_token.key
105  					auth = grafana_cloud_stack_service_account_token.management_token.key
131  					auth = grafana_cloud_stack_service_account_token.management_token.key

testdata/integration/main.tf
12     auth = "admin:admin"

cmd/generate/testdata/testblocks.tf
2      auth = "admin:admin"
```

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -19,7 +19,7 @@ The changelog for this provider can be found here: <https://github.com/grafana/t
 ```terraform
 provider "grafana" {
   url  = "http://grafana.example.com/"
-  auth = var.grafana_auth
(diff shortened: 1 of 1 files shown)


jdbaldry:
More context has revealed that these docs are correct.

PR #1590, opened 2024-05-29
Config Generation: Add type and name filter

julienduchesne:
Allows generating only a subset of resources, or all resources containing a certain string in their ID (or a combination of that)


K-Phoen (approved):
Nice!

Issue #1534, opened 2024-05-06
[Bug]: Error: PutAlertRuleGroup (status 500): {}


jwavoet:
### Terraform Version

1.8.2

### Terraform Grafana Provider Version

2.18.0

### Grafana Version

9.4.7

### Affected Resource(s)

* grafana_rule_group

### Terraform Configuration Files

```hcl
resource "grafana_rule_group" "my_alert_rule" {
  provider         = grafana.provider
  name             = "Development-Customer"
  folder_uid       = grafana_folder.customer_internal_folder.uid
  interval_seconds = 60

  rule {
    name           = "Development-CPU%"
    for            = "5m"
    condition      = "B"
    no_data_state  = "NoData"
    exec_err_state = "Error"
    annotations = {
      "__dashboardUid__" = "Ia8HCJGTh"
      "__panelId__"      = "2"
    }
    is_paused = false

    data {
      ref_id     = "CPU"
      query_type = ""
      relative_time_range {
        from = 3600
        to   = 0
      }
      datasource_uid = "jO5EHHfIk"
      model = jsonencode({
        "accountId" : "123456789000",
        "alias" : "",
        "datasource" : {
          "type" : "cloudwatch",
          "uid" : "jO5EHHfIk"
        },
        "dimensions" : {
          "InstanceId" : [
            "i-12345678900000000"
          ]
        },
        "expression" : "",
        "id" : "",
        "intervalMs" : 1000,
        "label" : "",
        "logGroups" : [],
        "matchExact" : true,
        "maxDataPoints" : 43200,
        "metricEditorMode" : 0,
        "metricName" : "CPUUtilization",
        "metricQueryType" : 0,
        "namespace" : "AWS/EC2",
        "period" : "",
        "queryMode" : "Metrics",
        "refId" : "CPU",
        "region" : "eu-west-1",
        "sqlExpression" : "",
        "statistic" : "Average"
      })
    }

    data {
      ref_id     = "A"
      query_type = ""
      relative_time_range {
        from = 3600
        to   = 0
      }
      datasource_uid = "__expr__"
      model          = <<EOT
{
  "conditions": [
    {
      "evaluator": {
        "params": [],
        "type": "gt"
      },
      "operator": {
        "type": "and"
      },
      "query": {
        "params": [
          "A"
        ]
      },
      "reducer": {
        "params": [],
        "type": "last"
      },
      "type": "query"
    }
  ],
  "datasource": {
    "type": "__expr__",
[...]

PR #1610, opened 2024-06-05
Rule group detect rule name conflicts 

WeberKuo:
Based on Issue #1534, it has been clarified that the non-duplicative aspect should apply to `rule names` rather than `rule group names`.

The modifications in PR #1550 will result in disallowing the existence of duplicate `rule group names` within different folders. However, this should be permissible.

This adjustment will entail checking for duplicate `rule names` within the same `rule group` and also verifying if duplicate `rule names` exist for rules already present on `Grafana`.

Upon testing, it has been observed that blocking identical rule names will only occur within the same rule group. 
It is indeed permissible for rule groups in different folders to have duplicate `rule names`.

### DIFF
--- a/internal/resources/grafana/resource_alerting_rule_group.go
+++ b/internal/resources/grafana/resource_alerting_rule_group.go
@@ -339,17 +339,30 @@ func readAlertRuleGroup(ctx context.Context, data *schema.ResourceData, meta int
 
 func putAlertRuleGroup(ctx context.Context, data *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client, orgID := OAPIClientFromNewOrgResource(meta, data)
+
+	respAlertRules, err := client.Provisioning.GetAlertRules()
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	if data.IsNewResource() {
+		// Check if a rule group with the same name already exists. The API either:
+		// - overwrites the existing rule group if it exists in the same folder, which is not expected of a TF provider.
+
+		for _, rule := range respAlertRules.Payload {
+			name := data.Get("name").(string)
+			if *rule.RuleGroup == name && *rule.FolderUID == folder {
+				return diag.Errorf("rule group with name %q already exists", name)
+			}
+		}
+	}
+
 	group := data.Get("name").(string)
 	folder := data.Get("folder_uid").(string)
 	interval := data.Get("interval_seconds").(int)
 
 	packedRules := data.Get("rule").([]interface{})
 	rules := make([]*models.ProvisionedAlertRule, 0, len(packedRules))
-
-	respAlertRules, err := client.Provisioning.GetAlertRules()
-	if err != nil {
-		return diag.FromErr(err)
-	}
 
 	for i := range packedRules {
 		rule, err := unpackAlertRule(packedRules[i], group, folder, orgID)
--- a/internal/resources/grafana/resource_alerting_rule_group.go
+++ b/internal/resources/grafana/resource_alerting_rule_group.go
@@ -351,6 +351,7 @@ func putAlertRuleGroup(ctx context.Context, data *schema.ResourceData, meta inte
 
 		for _, rule := range respAlertRules.Payload {
 			name := data.Get("name").(string)
+			folder := data.Get("folder").(string)
 			if *rule.RuleGroup == name && *rule.FolderUID == folder {
 				return diag.Errorf("rule group with name %q already exists", name)
 			}
--- a/internal/resources/grafana/resource_alerting_rule_group.go
+++ b/internal/resources/grafana/resource_alerting_rule_group.go
@@ -351,7 +351,7 @@ func putAlertRuleGroup(ctx context.Context, data *schema.ResourceData, meta inte
 
 		for _, rule := range respAlertRules.Payload {
 			name := data.Get("name").(string)
-			folder := data.Get("folder").(string)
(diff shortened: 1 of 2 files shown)


julienduchesne (commented):
I think the current message is pretty clear:

```
grafana_rule_group.test3: Creating...
│ Error: [PUT /v1/provisioning/folder/{FolderUID}/rule-groups/{Group}][400] putAlertRuleGroupBadRequest {"message":"failed to insert alert rules: [alerting.alert-rule.conflict] conflicting alert rule found [rule_uid: 'cdnv1d3h4x88wa', title: 'My Random Walk Alert', namespace_uid: 'blabla-child']: rule title under the same organisation and folder should be unique"}
```

The purpose of the check (within `data.IsNewResource`) is to avoid "overwriting" rule groups. Since the API call is a PUT, Terraform had no way to check that a rule group already exists. So I don't think it should be removed

Though there is an error in it. It is indeed possible to have the same rule group name in different folders so line 354 should probably be `if *rule.RuleGroup == name && *rule.FolderUID == folder {`

WeberKuo:
This was my negligence. I will revert it.

internal/resources/grafana/resource_alerting_rule_group.go
```diff
@@ -363,11 +363,26 @@ func putAlertRuleGroup(ctx context.Context, data *schema.ResourceData, meta inte
...
+			if *r.Title == *rule.Title {
+				return diag.Errorf("rule with name %q is defined more than once", *rule.Title)
+			}
+		}
+
+		for _, r := range respAlertRules.Payload {
+			if r.UID != rule.UID && *r.Title == *rule.Title && *r.FolderUID == *rule.FolderUID {
+				return diag.Errorf("rule with name %q already exists in the alert group", *rule.Title)
+			}
+		}
```
julienduchesne, line 385: I'm not sure this is useful at all. The API returns a pretty clear error message when there are duplicated rules

WeberKuo: I tried to use the same rule name under two different alert groups.
It returned an error: 
```
Error: [PUT /v1/provisioning/folder/{FolderUID}/rule-groups/{Group}] PutAlertRuleGroup (status 500): {}
```
on Grafana 9.5.19.

julienduchesne: Hmm, gotcha. Later versions have better error handling it seems

internal/resources/grafana/resource_alerting_rule_group_test.go
```diff
@@ -384,6 +378,84 @@ resource "grafana_rule_group" "second" {
 	})
 }
 
+func TestAccAlertRule_ruleNameConflict(t *testing.T) {
```
julienduchesne, line 382: This test doesn't run:
```
=== CONT  TestAccAlertRule_nameConflict
    resource_alerting_rule_group_test.go:301: Step 1/1, expected an error with pattern, no match on: Error running pre-apply refresh: exit status 1
        
        Error: Duplicate resource "grafana_rule_group" configuration
        
          on terraform_plugin_test.tf line 41:
          41: resource "grafana_rule_group" "test" {
        
        A grafana_rule_group resource named "test" was already declared at
        terraform_plugin_test.tf:11,1-37. Resource names must be unique per type in
        each module.
    testing_new.go:74: Error retrieving state, there may be dangling resources: exit status 1
        
        Error: Duplicate resource "grafana_rule_group" configuration
        
          on terraform_plugin_test.tf line 41:
          41: resource "grafana_rule_group" "test" {
        
        A grafana_rule_group resource named "test" was already declared at
        terraform_plugin_test.tf:11,1-37. Resource names must be unique per type in
        each module.
--- FAIL: TestAccAlertRule_nameConflict (0.55s)
```

internal/resources/grafana/resource_alerting_rule_group_test.go
```diff
@@ -384,6 +378,84 @@ resource "grafana_rule_group" "second" {
...
+				hide          = false
+				intervalMs    = 1000
+				maxDataPoints = 43200
+				refId         = "A"
+			})
+		}
+	}
+}
+				`, name),
+				ExpectError: regexp.MustCompile(`rule with name "My Alert Rule" is defined more than oncendnndndndn`),
```
julienduchesne: That's probably wrong

julienduchesne (commented):
You can run the test locally with `GRAFANA_VERSION=10.4.3 TESTARGS='-run="TestAccAlertRule_ruleNameConflict"' make testacc-oss-docker`

internal/resources/grafana/resource_alerting_rule_group_test.go
julienduchesne: Oh wait, that's the one that was already there. The issue is the same though

internal/resources/grafana/resource_alerting_rule_group_test.go
WeberKuo: I fixed the bug in the unit test, but some 'Get Secrets' in jobs cannot be retrieved. How should I handle this?

julienduchesne (approved):
LGTM! Thanks!

internal/resources/grafana/resource_alerting_rule_group_test.go
julienduchesne: It's fine. These are just enterprise and cloud tests. All the tests that you added/modified in this PR have passed.

PR #1616, opened 2024-06-05
Common Resource and DataSource framework

julienduchesne:
Just like it was done for resources, expose the datasource as structs that wrap their name and schema 
This will allow us to add the category to the these objects and get rid of this: https://github.com/grafana/terraform-provider-grafana/blob/main/tools/subcategories.json


joeyorlando (approved):

sd2k (approved):
:+1: from ML

PR #1631, opened 2024-06-11
[Fix] Adds check to ensure that the Grafana-Terraform-Provider header is set

### LINKED ISSUE #1630: [Fix] Returns an Error if the Grafana-Terraform-Provider Header is missing for Requests to the SLO API
To prevent regression due to missing headers from the Terraform Provider, we add a check directly in the TF Provider that ensures that `resourceSloRead` calls always have a `readOnly.Provenance` field of "terraform". 

`resourceSloRead` is called after all CREATE/UPDATEs made by the Terraform Provider, so it is reasonable to expect that the modified Slo would have a `readOnly.Provenance` field of `terraform`.

If this is missing, it's likely because the header was removed from within the Terraform Provider.

elainevuong:
To prevent regression due to missing request headers from the Terraform Provider, we add a check in the TF Provider Tests that ensures that following a successful CREATE or UPDATE, that a SLO returned from the READ call to the SLO API has a `provenance` field of `terraform`. 

This is **done directly in the TF Provider Tests** `resource_slo_test` and does **NOT** affect the actual function of the Terraform Provider, which means it does not interfere with IMPORT functionality.

Fixes #1630

### DIFF
--- a/internal/resources/slo/resource_slo.go
+++ b/internal/resources/slo/resource_slo.go
@@ -305,12 +305,18 @@ func resourceSloRead(ctx context.Context, d *schema.ResourceData, client *slo.AP
 
 	req := client.DefaultAPI.V1SloIdGet(ctx, sloID)
 	slo, _, err := req.Execute()
-
 	if err != nil {
 		return apiError("Unable to read SLO - API", err)
 	}
 
-	setTerraformState(d, *slo)
+	err = setTerraformState(d, *slo)
+	if err != nil {
+		diags = append(diags, diag.Diagnostic{
+			Severity: diag.Error,
+			Summary:  "Missing Terraform provenance request header",
+			Detail:   err.Error(),
+		})
+	}
 
 	return diags
 }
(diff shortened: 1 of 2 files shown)


elainevuong:
error vid - missing the `Grafana Terraform Provider` header

https://github.com/grafana/terraform-provider-grafana/assets/73451363/9f161a85-8a8e-450e-afe3-f0765f0f4968

elainevuong:
success vid

https://github.com/grafana/terraform-provider-grafana/assets/73451363/08bad5ee-269e-474d-bd02-2ed81d8ddedc

julienduchesne (commented):
This prevents someone from importing a SLO and taking it over with Terraform

ellisda:
> This prevents someone from importing a SLO and taking it over with Terraform

Is that because the `resourceSloRead` function is called during import as well? Is there a unit or integration test that we can add such a check to, outside of the production code path?

julienduchesne:
Yep, read is called on imports. The only way to test this would be to have a new acceptance test that creates a SLO without the header and then imports it (plenty of import examples in other TF tests)

elainevuong:
~~I'll update this - and add the test @julienduchesne suggested - a new acceptance test that creates a SLO without the header and then imports it. thanks for the comments!~~

ended up doing this check directly in the TF Provider Tests, and not in the Provider itself so it shouldn't affect the Import functionality. doing the import tests took a bit more time to think through - I think it would require changes to the TF SLO Schema, since we don't currently have the `readOnly` field included in our SLO Schema anywhere. decided to just test for the header directly in our TF Provider Tests instead

julienduchesne (approved):

Issue #1620, opened 2024-06-07
[Bug]: "prevent_destroy_if_not_empty" ignores dashboards in child folder


Sakib37:
### Terraform Version

Terraform v1.5.5

### Terraform Grafana Provider Version

3.1.0

### Grafana Version

11.1.0

### Affected Resource(s)

* grafana_folder

The [prevent_destroy_if_not_empty](https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/folder#prevent_destroy_if_not_empty) flag can only protect from unexpected deletion of a folder when there is dashboard inside the parent folder. 

However, with new sub-folder feature, we can create subfolder inside a parent folder. If we have dashboards inside a child folder, the `prevent_destroy_if_not_empty` does not prevent to destroy the parent folder. 

### Expected Behavior

For any folder, `prevent_destroy_if_not_empty` checks if there is any dashboard in the subfolders as well and prevent deletion of a folder in this kind of situation

### Actual Behavior

`prevent_destroy_if_not_empty` cannot detect if there is any dashboard inside a child folder. So folder deletion deletes the parent folder including all this child folders as well as the dashboards inside the child folder.

### Steps to Reproduce

-  Create a folder by terraform and also enable the `prevent_destroy_if_not_empty` parameter.
- Manually create few child folder inside the parent folder 
- Manually create some dashbaords inside the child folders 
- Now try to delete the parent folder via terraform.

julienduchesne:
Fixed here: https://github.com/grafana/terraform-provider-grafana/pull/1635. Will be released by EOW

PR #1635, opened 2024-06-12
Prevent deletion on nested folders

malcolmholmes:
As described in #1620, if a folder contains a non-Terraform managed nested folder, `prevent_destroy_if_not_empty` will not protect the folder from deletion.

This PR fixes that. As well as looking for dashboards in the folder prior to deletion, it also checks the folder for inner folders.

### DIFF
--- a/internal/resources/grafana/resource_folder.go
+++ b/internal/resources/grafana/resource_folder.go
@@ -155,6 +155,7 @@ func DeleteFolder(ctx context.Context, d *schema.ResourceData, meta interface{})
 	client, _, uid := OAPIClientFromExistingOrgResource(meta, d.Id())
 	deleteParams := folders.NewDeleteFolderParams().WithFolderUID(uid)
 	if d.Get("prevent_destroy_if_not_empty").(bool) {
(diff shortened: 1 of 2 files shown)


internal/resources/grafana/resource_folder.go
```diff
@@ -155,19 +155,33 @@ func DeleteFolder(ctx context.Context, d *schema.ResourceData, meta interface{})
...
+		}
+		if len(searchResp.GetPayload()) > 0 {
+			for _, folder := range searchResp.GetPayload() {
+				dashboardAndFolderNames = append(dashboardAndFolderNames, folder.Title)
+			}
+		}
+		if len(dashboardAndFolderNames) > 0 {
+			return diag.Errorf("folder %s is not empty and prevent_destroy_if_not_empty is set. It contains the following dashboards and/or folders: %v", uid, dashboardAndFolderNames)
+		}
+
```
julienduchesne: If you remove the type from those searches, can you do a single API call?

malcolmholmes: Yes, you're right.

PR #1653, opened 2024-06-19
add `grafana_oncall_user_notification_rule` resource

### LINKED ISSUE #1193: Allow configuring users through Terraform
Use-case: large org onboards to OnCall. They would like to pre-fill notification policies & phone numbers.

We don't allow editing **Users** though API or TF, only to list: https://grafana.com/docs/oncall/latest/oncall-api-reference/users/

We allow editing **Personal Notifications** through API but we don't allow editing through Terraform: https://grafana.com/docs/oncall/latest/oncall-api-reference/personal_notification_rules/

joeyorlando:
Related to the Terraform portion of https://github.com/grafana/oncall/issues/4410

Closes https://github.com/grafana/oncall/issues/1193

Once a new provider version is released, I will open a PR in https://github.com/grafana/crossplane-provider-grafana to generate the related Crossplane resources

Tested on my local stack all of the following cases and all work as expected:
- creating several `grafana_oncall_user_notification_rule`, various `type`s and `position`s
- updates
- deletions
- importing
- force recreates - changing either the `user_id` or `important` value for a resource, should delete the old resource and create a new one


julienduchesne:
This is going to be an annoying comment but we're trying to get rid of the old TF framework (which is what this PR is using). New resources look like this: https://github.com/grafana/terraform-provider-grafana/blob/main/internal/resources/cloud/resource_cloud_org_member.go

joeyorlando:
thanks for pointing that out @julienduchesne, will take a look 👍

joeyorlando:
@julienduchesne done!

internal/resources/oncall/resource_user_notification_rule_test.go
```diff
@@ -0,0 +1,87 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckOnCallUserNotificationRuleResourceExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, "user_id", userID),
+					resource.TestCheckResourceAttr(resourceName, "position", "1"),
+					resource.TestCheckResourceAttr(resourceName, "duration", "300"),
+					resource.TestCheckResourceAttr(resourceName, "type", "wait"),
+					resource.TestCheckResourceAttr(resourceName, "important", "false"),
+				),
+			},
+			// TODO: add some more test cases...
```
julienduchesne: An import test case is an easy one to add:

```
{
  Import: true,
  ImportVerify: true
}
```

julienduchesne (commented):
Other than the failing linter + an import test, this looks great! Thanks for redoing everything in the new plugin framework. Much appreciated.

internal/resources/oncall/resource_user_notification_rule.go
```diff
@@ -0,0 +1,328 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+var (
+	resourceUserNotificationRuleName = "grafana_oncall_user_notification_rule"
+
+	userNotificationRuleTypeOptions = []string{
```
vstpme, line 28: should MS Teams be in this list?

internal/resources/oncall/resource_user_notification_rule.go
```diff
@@ -0,0 +1,328 @@
...
+			},
+			"user_id": schema.StringAttribute{
+				MarkdownDescription: "User ID",
+				Required:            true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"position": schema.Int64Attribute{
+				MarkdownDescription: "Personal notification rules execute one after another starting from position=0. Position=-1 will put the escalation policy to the end of the list. A new escalation policy created with a position of an existing escalation policy will move the old one (and all following) down on the list.",
```
vstpme: `Position=-1 will put the escalation policy to the end of the list` this [doesn't work](https://github.com/grafana/oncall/blob/038d62788f66c5123174bd0efe41fa0ecd97c671/engine/common/ordered_model/serializer.py#L47) for Terraform (if there are multiple notification rules to be created, Terraform will issue multiple concurrent requests and there's no way to guarantee that the `-1` rule will actually be the last one).

joeyorlando: Good catch! Addressed in https://github.com/grafana/terraform-provider-grafana/pull/1653/commits/bdc35ec84cfbb5be2ca268da7996038996c1df38

internal/resources/oncall/resource_user_notification_rule.go
joeyorlando: Addressed in https://github.com/grafana/terraform-provider-grafana/pull/1653/commits/bdc35ec84cfbb5be2ca268da7996038996c1df38 (I'll also update our public API docs to make a mention that this is available as well (only in Grafana Cloud))

joeyorlando: https://github.com/grafana/oncall/pull/4628/files#diff-7225c8889a4ef0527f6d724edcb8482d7f3f89e41e8e6e68e393214d33f8aae1R38

julienduchesne (approved):
👍

PR #1660, opened 2024-06-26
feat(slo): Update Terraform provider to include option to set min Failures

Leo-DiCara:
We recently have merged new functionality which has updated the SLO CRD. This update brings the terraform provider in line with the new capabilities. 

-Doc Update
-New Test for advanced Options

Part of https://github.com/grafana/slo/issues/2070


internal/resources/slo/data_source_slo.go
```diff
@@ -212,6 +216,18 @@ func unpackAlertingMetadata(metaData slo.SloV00AlertingMetadata) []map[string]in
 	return retAlertMetaData
 }
 
+func unpackAdvancedOptions(options slo.SloV00AdvancedOptions) []map[string]int {
```
ellisda: question: Should this return type be `[]map[string]interface{}` like other `unpack...` functions?

I don't know if this matters

internal/resources/slo/resource_slo.go
```diff
@@ -221,6 +221,22 @@ Resource manages Grafana SLOs.
 								},
 							},
 						},
+						"advanced_options": {
+							Type:        schema.TypeList,
+							MaxItems:    1,
```
ellisda, line 226: question: are there pros/cons to omitting this `MaxItems` field?

I'm thinking about what happens if we add another field under advancedOptions, months from now. Do we end up with the same chicken and egg issue where older terraform provider yells about new fields that it doesn't know about.

internal/resources/slo/resource_slo.go
```diff
@@ -552,6 +568,20 @@ func packAlerting(tfAlerting map[string]interface{}) slo.SloV00Alerting {
 		SlowBurn:    &tfSlowBurn,
 	}
 
+	// All options in advanced options will be optional
+	// Adding a second feature will need to make a better way of checking what is there
+	if failures := tfAlerting["advanced_options"]; failures != nil {
+		lf := failures.([]interface{})
```
ellisda:
```suggestion
	if advancedOptions, ok := tfAlerting["advanced_options"].([]interface{}); ok {
```

it seems that the blocks above this all follow the same pattern with type assertion on the "if" line. Can we not follow that pattern here? Note: the suggestion is using `[]interface{}` not `[]int` (type depends on whether you take suggestion above)

internal/resources/slo/resource_slo.go
```diff
@@ -552,6 +568,20 @@ func packAlerting(tfAlerting map[string]interface{}) slo.SloV00Alerting {
...
+		if len(lf) > 0 {
+			lf2 := lf[0].(map[string]interface{})
```
ellisda: issue: this will panic if the runtime value of `lf[0]` fails the type assertion.

Don't you need to use the `lf2, ok := lf[0].(map[string]interface{})` version and check `ok`?

ellisda (commented):
looks pretty good (though I'm not a good reviewer here). a couple of style questions to align with existing code

internal/resources/slo/resource_slo.go
Leo-DiCara: I assumed this was the number of advanced_options, not the number of sub items when I did this, but your right it shouldn't have a max items restriction.

internal/resources/slo/data_source_slo.go
Leo-DiCara: Elaine told me to use this and it worked. I originally had it the interface way. I can try it the other way and see if it passes the tests?

internal/resources/slo/resource_slo.go
Leo-DiCara: Interesting point. Yeah I should fix that.

Leo-DiCara: We actually do this exact same thing for the alerting struct as a whole, the terraform can panic if it doesn't match the expected format. Think I should switch that too?

internal/resources/slo/resource_slo.go
Leo-DiCara: For example the top level alerting has MaxItems set to 1.
`			"alerting": {
				Type:     schema.TypeList,
				MaxItems: 1,
				Optional: true,
				...<other stuff>...
			}`

internal/resources/slo/data_source_slo.go
ellisda: I'd trust Elaine. I was just seeing the bits of code above and below. more type information is better for reader at least

internal/resources/slo/resource_slo.go
ellisda: yes, if it's easy. If we can handle more gracefully than panic, we should.

where else are we doing it though? The only uncheecked type assertions I see in this function are on line 574 and 576

Leo-DiCara: I actually misunderstood. the alerting struct handles it all gracefully. I've replicated it now  here.

elainevuong (approved):
nice job - LGTM!

spinillos (approved):

malcolmholmes (approved):

internal/resources/slo/resource_slo_test.go
```diff
@@ -150,6 +160,30 @@ func testAlertingExists(expectation bool, rn string, slo *slo.SloV00Slo) resourc
 	}
 }
 
+func testAdvancedOptionsExists(expectation bool, rn string, slo *slo.SloV00Slo) resource.TestCheckFunc {
```
julienduchesne, line 163: Heh, I guess this works but this could be tested without an additional test function. An import test would've tested that things are written correctly too (it wipes the state and checks that all attributes would be set the same on an import)

julienduchesne (approved):
LGTM!

PR #1662, opened 2024-07-01
Cloud Provider AWS CloudWatch Scrape Jobs scaffolding

tristanburgess:
This resolves https://github.com/grafana/grafana-aws-app/issues/371

This PR creates some initial code scaffolding for new AWS CloudWatch Scrape Job resources for the Cloud Provider AWS App. See [this design doc](https://docs.google.com/document/d/1Cu98HukviDahdQ2cSc5fpURxQ75oh4iblyzDryeqLdE/edit#heading=h.5sybau7waq2q) for more details.

This PR also migrates the AWS Accounts resources onto (and has the Scrape Job resources already using) the [Terraform Plugin Framework](https://developer.hashicorp.com/terraform/plugin/framework), which seems to be the eventual way this repo is going. This is the new framework for developing Terraform providers, compared to SDKv2, and opens up to more options in the future for improving our schema, such as using [Nested Attributes](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes/list-nested) [instead of blocks](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/blocks), if `terraform-provider-grafana` is updated someday to use the v6 Plugin Protocol (from its current v5).

```
GRAFANA_URL=http://localhost:3000 GRAFANA_AUTH=admin:admin make testacc
```
passes locally

I also tested running TF apply to CRUD the resources tied to test data, tied to a "real" AWS sandbox account TF resource per the TF config in https://docs.google.com/document/d/1fpkhEsevdhSLy0GV-9IrN-4wrolGN3mRhtD4XPXcYdg/edit#heading=h.qogpfojwu1sj


docs/data-sources/cloud_provider_aws_account.md
```diff
@@ -17,9 +17,18 @@ data "grafana_cloud_stack" "test" {
   slug = "gcloudstacktest"
```
tristanburgess, line 17: all files in `docs/` are autogenerated using `go generate`. they are built from the resource schemas + description strings added in `internal/resources/cloudprovider` files, as well as the config examples in `examples/resources/grafana_cloud_provider_aws*` and `examples/data-sources/grafana_cloud_provider_aws*`

internal/common/cloudproviderapi/client.go
```diff
@@ -80,6 +80,31 @@ func (c *Client) DeleteAWSAccount(ctx context.Context, stackID string, accountID
...
+	Regions               []string
+	ServiceConfigurations []AWSCloudWatchServiceConfiguration
+}
+type AWSCloudWatchServiceConfiguration struct {
+	Name                        string
+	Metrics                     []AWSCloudWatchMetric
+	ScrapeIntervalSeconds       int64
+	ResourceDiscoveryTagFilters []AWSCloudWatchTagFilter
+	TagsToAddToMetrics          []string
+	IsCustomNamespace           bool
```
kgeckhart: I wonder if we can model the custom namespace jobs a bit cleaner here 🤔 

ATM a custom namespace job doesn't actually support the two tag features above this because there's no resource discovery and no resource discovery = no tags. It's a bit of bleh detail for a consumer of this to have to know about.

tristanburgess: Agreed on that point, good to know that support doesn't carry over, and makes sense. We might be best off implementing a separate resource + data sources for custom namespace jobs, then. Another idea could be to automatically designate the service name as a custom namespace on the backend if it doesn't match anything in our service maps, but that could be a bit of a surprising behavior as well.

I'll try to think of any other ideas, as well.

kgeckhart: > Another idea could be to automatically designate the service name as a custom namespace on the backend if it doesn't match anything in our service maps, but that could be a bit of a surprising behavior as well.

The fun part with that is some customers have realized that they can setup a custom namespace for a service we don't have defined quite yet. It would be fun if in one terraform run it's a custom and the next it's not 😆 .

Ooo another thing I realized that's going to be a PITA for a consumer is CustomNamespace configs use the CloudWatch metric name while a supported service uses prom metric names. Perhaps we might want to move to the CloudWatch metric names for both. We know we also want to start persisting the CloudWatch name vs the prom name internally so this will help us out in the long run.

But all of this is to say we should probably split https://github.com/grafana/terraform-provider-grafana/pull/1662/files/0807df467fe5e1dc6dc755e2df617cba7fd10c46#diff-10897d974440fb02a00222ce085ea6194ec2cc20e0c99969022cbe7a797e58d4R89 in to two fields. We could call the one `CustomNamespaces` and the other `Services`, wdyt?

tristanburgess: The metric names I'm not 100% sure, do you think it would be confusing at all that the input to our TF provider is different than what we have listed in [our docs (the Prom names)](https://grafana.com/docs/grafana-cloud/monitor-infrastructure/aws/cloudwatch-metrics/services/)? I suppose all of the CW metric names are listed in docs as well, so it should be okay. Our Prom metric names duplicate the service names anyways, so could be less annoying to type up the CW names. I can update the field descriptions to mention that our API expects the CW metric names.

The field split makes sense to me, I can work that

kgeckhart: If I'm honest I'm not sure why it was chosen to use expose the prom metric names. I think at one point it was helpful for a customer to see it so they could know what to look for but now we have the explore query links instead.

We can update our docs to publish both names so they are both readily available in our docs. If we use the CloudWatch names with terraform then customers who are looking at the CloudWatch docs for supported metrics aren't DOA or having to do a conversion on the fly themselves.

tristanburgess: Yeah I like that idea, publishing both makes sense.

tristanburgess:
Seems that acceptance tests will fail until https://github.com/grafana/terraform-provider-grafana/pull/1660 is merged. Getting
```
 json: unknown field "advancedOptions"
```
from
```
RUN   TestAccExamples/SLO/data-sources/grafana_slos/data-source.tf
```

docs/data-sources/cloud_provider_aws_cloudwatch_scrape_jobs.md
kgeckhart, line 1: Just wondering what the use case is for for all of our resources also existing as data sources? I can kinda see the use case for accounts where someone might want to reference an account they created in one terraform module in another but am not quite seeing the value of scrape job being a data source. What am I missing?

tristanburgess: I guess it's not required, but in my experience with Terraform, I've typically seen corresponding data sources made available for resources. Data sources in general are about getting data about resources that aren't necessarily managed by Terraform. One example I could imagine could be to configure a scrape job in the UI as a "template", and then use a data source to retrieve it and copy its data into a set of resources, which might be faster to work with for e.g. service and metrics selection, although I know we're soon building our own UI experience to provide TF snippets for this stuff.

I implemented it because it didn't take much time to do so, but agree it might not get used. If we would rather start more bare and only introduce if it's asked for, that makes sense/sounds fine and I can remove them. This PR's commit history can always be referred to later as needed.

kgeckhart: Honestly, I was just wondering and will totally leave it up to you. It's already implemented so might as well leave it since it's going to use the same client API calls anyway.

kgeckhart (approved):
:shipit:

internal/common/cloudproviderapi/client.go
thepalbi: Just popping in here for future reference: docs have been updated to include both

PR #1667, opened 2024-07-08
Standardize setting empty ID

julienduchesne:
Two things:
- No need to do it in delete functions
- Use `common.WarnMissing` so that we have a consistent diagnostic and log


mem (approved):
Looks good to me for the SM part!

sd2k (approved):
LGTM from ML :+1:

joeyorlando (approved):
LGTM from oncall side 👍

malcolmholmes (approved):

PR #1685, opened 2024-07-15
Check oncall rolling_users shift does not include empty group

matiasb:
Related to https://github.com/grafana/oncall/issues/4603

### DIFF
--- a/internal/resources/oncall/resource_shift_test.go
+++ b/internal/resources/oncall/resource_shift_test.go
@@ -47,6 +47,13 @@ func TestAccOnCallOnCallShift_basic(t *testing.T) {
 				),
 			},
 			{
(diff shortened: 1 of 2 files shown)


internal/resources/oncall/resource_shift_test.go
```diff
@@ -46,6 +46,13 @@ func TestAccOnCallOnCallShift_basic(t *testing.T) {
 					resource.TestCheckResourceAttr("grafana_oncall_on_call_shift.test-acc-on_call_shift", "frequency", "hourly"),
 				),
 			},
+			{
+				Config: testAccOnCallOnCallShiftEmptyRollingUsers(scheduleName, shiftName),
```
julienduchesne, line 51: Could we also have another test that does test the new error message?

julienduchesne (approved):

PR #1692, opened 2024-07-17
Config Generation: Add some missing listers

julienduchesne:
Added a tiny tool to list resources that don't have a lister

- SA
- Team
- SM Checks

### DIFF
--- a/internal/resources/syntheticmonitoring/resource_check.go
+++ b/internal/resources/syntheticmonitoring/resource_check.go
@@ -790,28 +790,26 @@ multiple checks for a single endpoint to check different capabilities.
 		"grafana_synthetic_monitoring_check",
 		resourceCheckID,
 		schema,
-	)
+	).WithLister(listChecks)
 }
 
-// TODO: Fix lister
-// .WithLister(listChecks)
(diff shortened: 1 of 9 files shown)


internal/resources/syntheticmonitoring/resource_check.go
```diff
@@ -790,28 +790,26 @@ multiple checks for a single endpoint to check different capabilities.
...
-// 		return nil, err
-// 	}
-
-// 	var ids []string
-// 	for _, check := range checkList {
-// 		ids = append(ids, strconv.FormatInt(check.Id, 10))
-// 	}
-// 	return ids, nil
-// }
+func listChecks(ctx context.Context, client *common.Client, data any) ([]string, error) {
```
peterschretlen, line 796: Out of curiosity, what was it that needed to be fixed - the lister code does not seem to have changed?

peterschretlen (approved):
LGTM,  just a question on the context of the change

internal/resources/syntheticmonitoring/resource_check.go
julienduchesne: I don't know. The generator has gone through lots of changes since this was commented out. It works now 🤷

PR #1704, opened 2024-07-22
API: update url.UserPassword arguments to use TrimSpace

rluisr:
Resolving the problem of an Invalid Header error when reading `auth` using the terraform `file` function, where newlines etc. are inserted at the end of the credential file in formats such as Prettier, VSCode(insertFinalNewline).


julienduchesne (approved):
That seems reasonable to me. Thanks!

PR #1719, opened 2024-07-25
Address `file name too long` error when passing certs as literal values

lornest:
Hey team,

I came across this issue when attempting to create a PR to add the TLS attributes to the Crossplane provider's provider config (which will follow this fix). 

When testing my changes locally, I was encountering an error where my certs (literal values) were causing the Terraform client creation to fail with an error `file name too long`. After some debugging, it seems that if the cert literal is longer than the system file path limit (256 Windows, 255 Mac, ~4000 on Linux), the error is not handled by `os.ErrNotExist`. I wrote a test to expose this and have created a PR for a fix.

Note, this fix will only work for Unix like systems (I'm running Apple Sillicon/darwin_arm64), I don't think Windows has an equivalent to `syscall.ENAMETOOLONG` in Go. I hummed and hawed about the best way to handle this, so am happy to change based on any feedback. One option, for example, is to be slightly more 'heavy handed' and just check for any error of type `*PathError` returned by `os.Stat`. However, this has the downside that it'll create temp files for other errors such as insufficient privileges, which will then cause failures when creating resources as the 'cert' in the file will be nonsense. 

Let me know what you think.

### DIFF
--- a/pkg/provider/configure_clients.go
+++ b/pkg/provider/configure_clients.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"os"
 	"strings"
+	"syscall"
 	"time"
 
 	onCallAPI "github.com/grafana/amixr-api-go-client"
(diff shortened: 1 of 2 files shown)


lornest:
Update: file path character limits are

- Mac 255
- Windows 256 (but WSL ~32,768 apparently!)
- Linux 4096

Likely this issue won't occur on Linux machines (although in theory CA certs can be unlimited in length), so mostly a Mac (local) or Windows issue.

Again, happy to hear thoughts and feedback, but I'm now wondering if checking the GOOS and validating length based on the OS might be a tighter solution to this.

julienduchesne:
Great stuff. Thanks for fixing this and for the tests!

PR #1736, opened 2024-07-31
Alerting: Add support for recording rules

rwwiv:
This PR adds support for Grafana managed recording rules (nested field `record` in alert rule definition).

Related to https://github.com/grafana/grafana/pull/87779


internal/resources/grafana/resource_alerting_rule_group_test.go
```diff
@@ -675,6 +675,34 @@ func TestAccAlertRule_NotificationSettings(t *testing.T) {
 	})
 }
 
+func TestAccRecordingRule(t *testing.T) {
+	testutils.CheckOSSTestsEnabled(t, ">=11.2.0")
```
julienduchesne: Can you switch that to run in a cloud instance instead as 11.2 isn't released yet

```suggestion
	testutils.CheckCloudInstanceTestsEnabled(t) # TODO: Use 11.2 when released
```

internal/resources/grafana/resource_alerting_rule_group_test.go
```diff
@@ -675,6 +675,34 @@ func TestAccAlertRule_NotificationSettings(t *testing.T) {
...
+		CheckDestroy:             alertingRuleGroupCheckExists.destroyed(&group, nil),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccRecordingRule(name, metric, "A"),
+				Check: resource.ComposeTestCheckFunc(
+					alertingRuleGroupCheckExists.exists("grafana_rule_group.my_rule_group", &group),
+					resource.TestCheckResourceAttr("grafana_rule_group.my_rule_group", "name", name),
+					resource.TestCheckResourceAttr("grafana_rule_group.my_rule_group", "rule.#", "1"),
+					resource.TestCheckResourceAttr("grafana_rule_group.my_rule_group", "rule.0.name", "My Random Walk Alert"),
+					resource.TestCheckResourceAttr("grafana_rule_group.my_rule_group", "rule.0.for", "2m0s"),
```
alexweav: How does this test pass? For doesn't exist for recording rules, the server should be ignoring the field and returning `0` back to the provider here?

internal/resources/grafana/resource_alerting_rule_group_test.go
```diff
@@ -675,6 +675,34 @@ func TestAccAlertRule_NotificationSettings(t *testing.T) {
...
+					resource.TestCheckResourceAttr("grafana_rule_group.my_rule_group", "rule.0.data.0.model", "{\"hide\":false,\"refId\":\"A\"}"),
+					resource.TestCheckResourceAttr("grafana_rule_group.my_rule_group", "rule.0.data.0.record.metric", metric),
+					resource.TestCheckResourceAttr("grafana_rule_group.my_rule_group", "rule.0.data.0.record.from", "A"),
+				),
```
alexweav, line 704: Can we also get corresponding tests for Err/NoData state and the other alert rule fields? Even if it just asserts it's properly ignored.

Maybe the provider should reject such rules, that define these fields on recording rules?

internal/resources/grafana/resource_alerting_rule_group.go
```diff
@@ -706,3 +733,36 @@ func unpackNotificationSettings(p interface{}) (*models.AlertRuleNotificationSet
...
+}
+
+func unpackRecord(p interface{}) *models.Record {
+	if p == nil {
+		return nil
+	}
+	list := p.([]interface{})
+	if len(list) == 0 {
+		return nil
+	}
```
spinillos: Maybe add safe check here?
```suggestion
	list, ok := p.([]interface{})
	if !ok || len(list) == 0 {
		return nil
	}
```

spinillos (approved):

PR #1739, opened 2024-08-01
Teams: Allow specifying team members and admins

IevaVasiljeva:
**What**

* change the code to use team membership bulk update method (introduced in https://github.com/grafana/grafana/pull/87441);
* allow specifying member and admin level team memberships.

Note that this will reset team admin access back to team member for any users who have been provisioned as team members through terraform, and whose permissions have been manually updated to team admin. To avoid this, remove these users from `members` list and add them to `admins` list.

Fixes https://github.com/grafana/identity-access-team/issues/697

### DIFF
--- a/docs/data-sources/team.md
+++ b/docs/data-sources/team.md
@@ -43,10 +43,12 @@ data "grafana_team" "from_name" {
 
 ### Read-Only
 
+- `admins` (Set of String) A set of email addresses corresponding to users who should be given admin membership
+to the team. Note: users specified here must already exist in Grafana.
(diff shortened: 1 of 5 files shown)


julienduchesne:
This is going to be tricky because this API is only in Grafana 11 but the provider supports two versions down

IevaVasiljeva:
> This is going to be tricky because this API is only in Grafana 11 but the provider supports two versions down

Oh, good point. I didn't consider that. I don't think we're going to backport the API, so I'll close this PR for now and we can pick it back up when we're on Grafana 13 😬

Issue #1750, opened 2024-08-06
[Bug]: grafana_organization requires two applies when only 1 organization exist in Grafana


PovilasV1:
### Terraform Version

v1.9.3

### Terraform Grafana Provider Version

3.7.0

### Grafana Version

11.1.3

### Affected Resource(s)

grafana_organization, grafana_user

### Terraform Configuration Files

```
resource "grafana_user" "staff" {
  email    = "staff.name@example.com"
  name     = "Staff Name"
  login    = "staff"
  password = "my-password"
  is_admin = false
}

resource "grafana_user" "staff1" {
  email    = "staff1.name@example.com"
  name     = "Staff1 Name"
  login    = "staff1"
  password = "my-password"
  is_admin = false
}


resource "grafana_user" "admin" {
  email    = "admin@hostinger.com"
  name     = "admin@hostinger.com"
  login    = "admin@hostinger.com"
  password = "my-password"
  is_admin = true
}

resource "grafana_organization" "org" {
  name         = "Hostinger"
  create_users = false
  admin_user   = "admin"
  admins = [
    "admin@hostinger.com"
  ]
  editors = [
    "staff.name@example.com",
    "staff1.name@example.com",
  ]
  viewers = [
  ]

  depends_on = [grafana_user.staff,
    grafana_user.staff1
  ]
}

resource "grafana_organization" "org1" {
  name         = "Hostinger1"
  create_users = false
  admin_user   = "admin"
  admins = [
    "admin@hostinger.com"
  ]
  editors = [
    "staff.name@example.com",
    "staff1.name@example.com",
  ]
  viewers = [
  ]

  depends_on = [grafana_user.staff,
    grafana_user.staff1
  ]
}
```

### Expected Behavior

When creating a new user in Grafana using the terraform grafana_user resource, and then editing their role in the organization using the grafana_organization resource, the changes should be applied successfully in a single terraform apply command.

### Actual Behavior

If you have only 1 organization and would like to manage users within that organization using Terraform, you need to run Terraform apply twice for each user to get the desired role. The user created with the grafana_user resource gets the role that is set in auto_assign_org_id and terraform does not update it in the first apply.

### Steps to Reproduce

Trying to add a user to Grafana and then adding it to an organization:
```
Terraform will perform the following actions:

  # grafana_organization.org will be updated in-place
  ~ resource "grafana_organization" "org" {
      ~ editors              = [
          + "staff1.name@example.com",
[...]

PR #1752, opened 2024-08-07
Add checks if a user is already in the org due to auto_assign_org_id

PovilasV1:
related to https://github.com/grafana/terraform-provider-grafana/issues/1750

If the user is added to org by auto_assign_org_id, the provider doesn't see it and tries to add it again, skipping the role update. 

There might be a better solution, but I don't have experience with Terraform providers and Golang.


Duologic:
@grafana/identity-squad Is this a reasonable change? Should we move forward with this for issue #1750 ?

PR #1764, opened 2024-08-09
Config Generation: Use new categories enum instead of resource prefixes

julienduchesne:
Clears up a TODO
Also added a util to get a map of all resource types


Duologic (approved):
Nice, very clean.

PR #1784, opened 2024-08-29
fix(SLO): Handle resource 404 gracefully

joeblubaugh:
Terraform providers should react to a 404 as a "remove from tf state" command, but the SLO resource was not handling this correctly. This change fixes that in the SLO read path.


julienduchesne (approved):
LGTM, modulo that lint failure

julienduchesne:
Oh wait the test failed 🤔

joeblubaugh:
Yep - I'm still trying to figure out why there's an empty plan returned in the second step.

joeblubaugh:
@julienduchesne the SLO errors are fixed - there was a performance problem on deletes that was causing the tests to be too slow. This is fixed. There's a problem with the ML tests that I think is basically due to the current state of the instance.

julienduchesne (approved):
LGTM!

PR #1785, opened 2024-08-30
SSO: Add support for LDAP in SSO settings

dmihai:
This PR adds support for LDAP to the `grafana_sso_settings` resource. LDAP support in Grafana is currently experimental (hidden behind the `ssoSettingsLDAP` flag) and will be available starting with Grafana v11.3.

The unit test from this PR passed locally.

Solves https://github.com/grafana/identity-access-team/issues/715.

### DIFF
--- a/docs/resources/sso_settings.md
+++ b/docs/resources/sso_settings.md
@@ -3,13 +3,13 @@
 page_title: "grafana_sso_settings Resource - terraform-provider-grafana"
 subcategory: "Grafana OSS"
 description: |-
-  Manages Grafana SSO Settings for OAuth2 and SAML. Support for SAML is currently in preview, it will be available in Grafana Enterprise starting with v11.1.
+  Manages Grafana SSO Settings for OAuth2, SAML and LDAP. Support for LDAP is currently in preview, it will be available in Grafana starting with v11.3.
   Official documentation https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/HTTP API https://grafana.com/docs/grafana/latest/developers/http_api/sso-settings/
 ---
 
 # grafana_sso_settings (Resource)
 
-Manages Grafana SSO Settings for OAuth2 and SAML. Support for SAML is currently in preview, it will be available in Grafana Enterprise starting with v11.1.
+Manages Grafana SSO Settings for OAuth2, SAML and LDAP. Support for LDAP is currently in preview, it will be available in Grafana starting with v11.3.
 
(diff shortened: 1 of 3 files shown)


docs/resources/sso_settings.md
```diff
@@ -72,17 +72,87 @@ resource "grafana_sso_settings" "saml_sso_settings" {
 
 ### Required
 
-- `provider_name` (String) The name of the SSO provider. Supported values: github, gitlab, google, azuread, okta, generic_oauth, saml.
+- `provider_name` (String) The name of the SSO provider. Supported values: github, gitlab, google, azuread, okta, generic_oauth, saml, ldap.
 
 ### Optional
 
+- `ldap_settings` (Block Set, Max: 1) The LDAP settings set. Required for the ldap provider. (see [below for nested schema](#nestedblock--ldap_settings))
```
mgyongyosi, line 79:
```suggestion
- `ldap_settings` (Block Set, Max: 1) The LDAP settings set. Required for the LDAP provider. (see [below for nested schema](#nestedblock--ldap_settings))
```

docs/resources/sso_settings.md
```diff
@@ -72,17 +72,87 @@ resource "grafana_sso_settings" "saml_sso_settings" {
...
+
+Required:
+
+- `host` (String) The LDAP server host.
+- `search_base_dns` (List of String) An array of base DNs to search through.
+- `search_filter` (String) The user search filter, for example "(cn=%s)" or "(sAMAccountName=%s)" or "(uid=%s)".
+
+Optional:
+
+- `attributes` (Map of String) The LDAP server attributes. The following attributes can be configured: email, member_of, name, surname, username.
```
mgyongyosi, line 118: I'm not 100% sure, but I think either the `email` or `username` must be specified.

mgyongyosi (approved):
Stellar job. Left two suggestions.

docs/resources/sso_settings.md
dmihai: I wanted to keep `ldap` in lowercase here to match the value of the `provider_name` property. It's the same for the other providers (look at the saml description).

I understand it's a bit confusing to see `LDAP` in uppercase in the first sentence and then `ldap` in the second sentence. But the first one refers to the protocol name, while the second one should be the value users have to put in the `provider_name`. WDYT?

docs/resources/sso_settings.md
dmihai: This is not enforced on the server side and I think we shouldn't do it on the client.

docs/resources/sso_settings.md
mgyongyosi: Sounds reasonable, thanks!

PR #1789, opened 2024-09-04
cloudprovider: fix update calls for accounts and scrape jobs

tristanburgess:
This PR does two things:

1. for accounts, have the API-computed resource ID propagate the value state into the planned state. This allows us to call the Accounts API with only the planned state, and also simplifies Terraform diffs for customers (so it doesn't look like the ID is changing every time, which it isn't).
2. for scrape jobs, we can just use the planned state directly. I had confused this with the configured state (by the user), but this state typically omits some fields e.g. the optional `enabled` field. Just for some background, the planned state is ultimately a combination (listed here in order of ascending precedence) of the previous state, the user's configured state, and unknowns that must be fulfilled by the API.

Tested updating fields for an AWS account and CW scrape job locally, and after importing.


tristanburgess:
the test that's failing is unrelated, reached out to the owning team:
https://raintank-corp.slack.com/archives/C01A26KK49M/p1725547686946989

internal/resources/cloudprovider/resource_aws_cloudwatch_scrape_job.go
```diff
@@ -309,32 +309,26 @@ func (r *resourceAWSCloudWatchScrapeJob) Read(ctx context.Context, req resource.
...
-	var stateData awsCWScrapeJobTFModel
-	diags := req.State.Get(ctx, &stateData)
-	resp.Diagnostics.Append(diags...)
-	if resp.Diagnostics.HasError() {
-		return
-	}
-	var configData awsCWScrapeJobTFModel
-	diags = req.Config.Get(ctx, &configData)
+	var planData awsCWScrapeJobTFModel
+	diags := req.Plan.Get(ctx, &planData)
```
thepalbi: so to understand, there's three things here:
- req.State => data in the tf state (eg last applied)
- req.Config => what the hcl code describes right now at the moment of plan / apply
- req.Plan => mixin of the above, and what terraform show's as planned / intends to apply.

In another words, state is the before, config is the desired, and "`plan = state + config`"?

internal/resources/cloudprovider/resource_aws_account.go
```diff
@@ -82,6 +82,9 @@ func (r *resourceAWSAccount) Schema(ctx context.Context, req resource.SchemaRequ
 			"resource_id": schema.StringAttribute{
 				Description: "The ID given by the Grafana Cloud Provider API to this AWS Account resource.",
 				Computed:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
```
thepalbi: what's this again? didn't quite get it

thepalbi (commented):
some general questions to get my mind more in tf land

internal/resources/cloudprovider/resource_aws_cloudwatch_scrape_job.go
tristanburgess: yes exactly, with some minor exceptions such as the plan modifier decorator we talked about below which gives precedence to the current state's value rather than the configured state for certain Unknown values.

internal/resources/cloudprovider/resource_aws_account.go
tristanburgess: let me put a doc link as a comment on this one since it's kind of obscure: https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification#usestateforunknown

The resource_id should never change for an account, it can't be updated, it's assigned by our API. However, the schema for this attribute must be marked `Computed` because its value can only come from the API, it can't be specified by the requestor.

 Because of `Computed`, with no value being assigned by the requestor, every single modification to an account resource in Terraform would also mark the `resource_id` as "(known after apply)" in the plan. But this is unnecessary, because we know that the `resource_id` can never actually change. 
 
This framework-standard modifier makes it so that whenever the `resource_id` is unknown (which is on creation, and every time an existing resource is to be updated), it will fall back to the previously known state value. Therefore, on creation, the state is nonexistent for the resource, so indeed it will show "(known after apply)" in the plan, but after creation, on modification, the resource_id is known, so it will just keep assuming that same resource_id value in the plan without changing.

tristanburgess:
re-requesting review because we can just merge this into the feature branch despite the known unrelated test failing, as long as this looks good to you.

thepalbi (approved):

thepalbi:
> re-requesting review because we can just merge this into the feature branch despite the known unrelated test failing, as long as this looks good to you.

Yes sorry for the delay, was reading around docs on plugin development to get a better understanding on modifiers and else.

PR #1794, opened 2024-09-06
fix: use distinct names for ML forecast and outlier alerts

sd2k:
The alerts may be created at the same time which results in a conflict,
as encountered [here](https://github.com/grafana/terraform-provider-grafana/pull/1789#issuecomment-2332452686).


gitdoluquita (approved):

sd2k:
Closing as this seems to be an issue with our API not being idempotent - see https://github.com/grafana/machine-learning/pull/4794. I think we may need to make another fix after that's merged and deployed, let's see.

PR #1797, opened 2024-09-09
Resource: Add LBAC for datasources `data_source_config_lbac_rules` 

eleijonmarck:
**why/what**
We have seen that users would love to provision their lbac rules using terraform. this implements a experimental feature that will be highly developed. this is the first part of LBAC rules resource for terraform.

**dependancies:**
- https://github.com/grafana/terraform-provider-grafana/pull/1857 ✅ 
- https://github.com/grafana/terraform-provider-grafana/pull/1856 ✅ 
- https://github.com/grafana/grafana-enterprise/pull/7449 ✅ 
- https://github.com/grafana/terraform-provider-grafana/pull/1905 ✅ 
- https://github.com/grafana/grafana-enterprise/pull/7469

ability to run `terraform plan, terraform apply and terraform destroy of the rules resource`

Example terraform:
```hcl
resource "grafana_data_source" "test" {
  type = "loki"
  name = "loki"
  url  = "http://localhost:3100"
  basic_auth_enabled  = true
  basic_auth_username = "username"

  lifecycle {
    ignore_changes = [json_data_encoded, http_headers]
  }
}

resource "grafana_team" "test" {
	name = "test"
}

resource "grafana_data_source_config_lbac_rules" "test" {
    datasource_uid = grafana_data_source.test.uid
    rules = jsonencode({
        "${grafana_team.test.uid}" = [
            "{ foo != \"bar\", foo !~ \"baz\" }",
            "{ foo = \"qux\" }"
        ]
    })
}
```

### why is this difficult
lbac for datasources terraform resource, why is it a bit tricky?

the bidirection of `lbac_rules -> data_source:jsonData`, meaning we are updating the jsondata with another resource impacting another resource
Image

---
### proposals and thought process

there is no ID for the json field `teamHttpHeaders`
- so the state is only “stored” in the datasource as a whole, therefore leaning on copying the `data_source_config `resource more
**this is a resource that will be heavily used**
- i would like to have the internal state for terraform (some customers create about 500 rules and more), which clogs the datasource, i would like to copy the setup from `grafana_data_source_permission_item` for this. as it suits well for how we would like to setup this resource
- instead we make use of the api as it is. it is bulk update anyway. so we assume that all of the teams will be provided in the resource of `resource_data_source_config_lbac_rules`

**proposal of all rules combined into one resource**

**benefits**
- one api call to grafana
- mimicing the get/update apis

**tradeoffs**
- clonky json encoding (could be fixed)
- not ideal for teams managing their own set of rules

```hcl
resource "grafana_data_source_config_lbac_rules" "test" {
    datasource_uid = grafana_data_source.test.uid
	org_id = grafana_team.test.org_id
    rules = jsonencode({
        "${grafana_team.test.id}" = [
            "{ foo != \"bar\", foo !~ \"baz\" }",
            "{ foo = \"qux\" }"
        ]
    })
}
```

**proposal of each rule being a resource**

**benefits**
- each team rules, would be able to be owned by a team. this is preferred from a user perspective

**tradeoffs**
- too many api calls to grafana, if we have 500 rules we would have 500 api calls to the datasource endpoint

```hcl
resource "grafana_data_source_config_lbac_rule" "test" {
    datasource_uid = grafana_data_source.test.uid
    org_id = grafana_team.test.org_id
    team_id = "1"
    rules = [
            "{ foo != \"bar\", foo !~ \"baz\" }",
            "{ foo = \"qux\" }"
        ]
}
```

**proposal of embedding it into a resource and then add it to the json**

**benefits**:
- easier to implement

**tradeoff**
- management of lbac rules would lie inside of the json and we would not be able to move the storage of the lbac rules

It's a tradeoff of where you want the complexity to be. In Grafana with a new way to store this, or in the provider extracting it from json data, and possibly running into versioning issues.

I guess the less complex option is to leave everything as-is, isn't it? Have people manage those new fields in the existing json data? Perhaps provide a datasource helper for this, if the value is not trivial to build/represent

```hcl
data "grafana_datasource_team_headers" "helper" { ... }

resource "grafana_datasource" "test" {
  ...
  json_data_encoded = jsonencode({
     teamHeaders = data.grafana_datasource_team_headers.helper.value
     otherfields
  })
}
```

run tests via

```bash
GRAFANA_VERSION=11.1.0 
GRAFANA_URL=http://localhost:3000 
GRAFANA_AUTH=admin:admin 
TESTARGS="-run TestAccDataSourceLBAC_inOrg" make testacc-enterprise
```

The `teamHttpHeaders` resource is a datasource json field.

[epic](https://github.com/grafana/identity-access-team/issues/669)

basically this introduced a way to add/update a datasource json field
[lbac-api](https://github.com/grafana/grafana-enterprise/pull/6166/files#diff-4db5aa26eef3c740b6d4585fc98a773832db4d77da286de0bbf72e3717ba9636R40)

here is the doc for it. https://grafana.com/docs/grafana/latest/administration/data-source-management/teamlbac/

---
## next steps

improvements:
- [ ] API of the resource could be made so that there is no need of escaped JSON. To enforces types and schema

1.	Create a template file (logql_query.tmpl):
```txt
{f="alice", b="bob", a="abc"}
```

```hcl
resource "grafana_data_source_config_lbac_rules" "test" {
    datasource_uid = grafana_data_source.test.uid
    rules = jsonencode({
        "1" = [
            "{ foo != \"bar\", foo !~ \"baz\" }",
            "{ foo = \"qux\" }"
        ],
        "2" = [
            templatefile("${path.module}/logql_query.tmpl", {})
        ]
    })
}

```[tasklist]
- [ ] fix the lifecycle of the jsondata within grafana
  - [ ] make sure you cannot set teamhttpheaders from the json api (updateDatasource), ONLY through lbac/team api
  - [ ] magic diffsuprressfunc to ignore changes on json_data_encoded for the specific field teamHttpHeaders
```


undefinedhuman:
Hey there, is there an ETA when this feature will be shipped? Or is there anything that still needs to be done?

eleijonmarck:
@undefinedhuman we are working on the backend side of things for this. So this will wait for clear resource handling

undefinedhuman:
Thank you very much!

Robsta86:
What's the status of this? I was using teamHttpHeaders to configure lbac via the grafana_datasource_config, but that seems to be broken after these 2 are merged:

#1857 
#1856 

```  Error: teamHttpHeaders is a reserved key and cannot be used in JSON data. Use the data_source_config_lbac_rules resource instead ```

Now I am no longer able to manage my loki lbac rules via terraform, resulting in new onboarded teams to have full query access to my log datasource.

File: examples/resources/grafana_data_source_config_lbac_rules/resource.tf [hunk 1]
```diff
@@ -0,0 +1,33 @@
...
+  basic_auth_enabled  = true
+  basic_auth_username = "username"
+
+  json_data_encoded = jsonencode({
+    authType = "default"
+    ## basicAuthPassword = "<>"
+  })
+}
+
+# TODO: add back 5th of dec
```
IevaVasiljeva: This commended out code block also appears in [the docs](https://github.com/grafana/terraform-provider-grafana/pull/1797/files#diff-699c776ea911e1496728900cde3d15c92f34800e5f2701eec97295020257f320R43-R57), so we should really remember to uncomment the code before we do the next Terraform release.

Or maybe just track this as an action item in an issue rather than having it here.

internal/resources/grafana/resource_data_source_config_lbac_rules.go
```diff
@@ -0,0 +1,221 @@
...
+	rulesMap := make(map[string][]string)
+	err := json.Unmarshal([]byte(data.Rules.ValueString()), &rulesMap)
+	if err != nil {
+		resp.Diagnostics.AddError("Invalid rules JSON", fmt.Sprintf("Failed to parse rule s: %v", err))
+	}
+
+	apiRules := make([]*models.TeamLBACRule, 0, len(rulesMap))
+	for teamUID, rules := range rulesMap {
+		apiRules = append(apiRules, &models.TeamLBACRule{
+			TeamID:  "",
```
IevaVasiljeva: I think you can skip setting team ID altogether if you set UID, right?

internal/resources/grafana/resource_data_source_config_lbac_rules.go
```diff
@@ -0,0 +1,221 @@
...
+	data = resourceDataSourceConfigLBACRulesModel{
+		ID:            types.StringValue(datasourceUID),
+		DatasourceUID: types.StringValue(datasourceUID),
+		Rules:         types.StringValue(string(rulesJSON)),
+	}
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
+}
+
+func (r *resourceDataSourceConfigLBACRules) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
```
IevaVasiljeva, line 184: `Create` and `Update` have the same logic, could pull it out into a separate function and reuse.

internal/resources/grafana/resource_data_source_config_lbac_rules.go
```diff
@@ -0,0 +1,221 @@
...
+		return
+	}
+
+	data.ID = types.StringValue(datasourceUID)
+	data.DatasourceUID = types.StringValue(datasourceUID)
+	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
+}
+
+func (r *resourceDataSourceConfigLBACRules) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
+	resp.Diagnostics.AddWarning("Operation not supported", "Delete operation is not supported for LBAC rules")
```
IevaVasiljeva: We support removing LBAC rules on Grafana side, right? I don't think we can have a resource that can't be deleted.

IevaVasiljeva (commented):
Mostly looks good, but I've added a few questions.

internal/resources/grafana/resource_data_source_config_lbac_rules.go
eleijonmarck: ------------
basically terraform states that it is deleted, but it was not deleted on the grafana side of things. il implement the delete

examples/resources/grafana_data_source_config_lbac_rules/resource.tf
eleijonmarck: il track it in an item instead 👍

eleijonmarck: https://github.com/grafana/identity-access-team/issues/1047

internal/resources/grafana/resource_data_source_config_lbac_rules.go
eleijonmarck: refactoring into a separate function, as it will also be used for `delete`

eleijonmarck: i am adding a comment at the top to address that we are not dealing with a CRUD but a GET/UPDATE api

docs/resources/data_source_config_lbac_rules.md
```diff
@@ -0,0 +1,77 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "grafana_data_source_config_lbac_rules Resource - terraform-provider-grafana"
+subcategory: "Grafana Enterprise"
+description: |-
+  Manages LBAC rules for a data source.
+  !> Warning: The resource is experimental and will be subject to change. This resource manages the entire LBAC rules tree, and will overwrite any existing rules.
+  Official documentation https://grafana.com/docs/grafana/latest/administration/data-source-management/teamlbac/HTTP API https://grafana.com/docs/grafana/latest/developers/http_api/datasource_lbac_rules/
+  This resource requires Grafana >=11.4.0.
```
IevaVasiljeva: I think we actually need to set it to 11.5.0 - 11.4.0 will not include the changes that we need ([see this internal thread](https://raintank-corp.slack.com/archives/C010WU9Q5TJ/p1732112639018819)).

examples/resources/grafana_data_source_config_lbac_rules/resource.tf
```diff
@@ -0,0 +1,32 @@
...
+resource "grafana_data_source" "test" {
+  type                = "loki"
+  name                = "loki-from-terraform"
+  url                 = "https://mylokiurl.net"
+  basic_auth_enabled  = true
+  basic_auth_username = "username"
+
+  json_data_encoded = jsonencode({
+    authType = "default"
+    ## basicAuthPassword = "<>"
```
IevaVasiljeva: Is this line needed?

internal/resources/grafana/resource_data_source_config_lbac_rules.go
```diff
@@ -0,0 +1,237 @@
...
+		resourceDataSourceConfigLBACRulesID,
+		resourceStruct,
+	)
+}
+
+type LBACRule struct {
+	TeamID  types.String   `tfsdk:"team_id"`
+	TeamUID types.String   `tfsdk:"team_uid"`
+	Rules   []types.String `tfsdk:"rules"`
+}
```
IevaVasiljeva: Is this actually used?

internal/resources/grafana/resource_data_source_config_lbac_rules.go
```diff
@@ -0,0 +1,237 @@
...
+			"id": schema.StringAttribute{
+				Computed: true,
+			},
+			"datasource_uid": schema.StringAttribute{
+				Required:    true,
+				Description: "The UID of the datasource.",
+			},
+			"rules": schema.StringAttribute{
+				Required:    true,
+				Description: "JSON-encoded LBAC rules for the data source. Map of team IDs to lists of rule strings.",
```
IevaVasiljeva:
```suggestion
				Description: "JSON-encoded LBAC rules for the data source. Map of team UIDs to lists of rule strings.",
```

internal/resources/grafana/resource_data_source_config_lbac_rules_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"reflect"
+	"testing"
+
+	"github.com/grafana/terraform-provider-grafana/v3/internal/testutils"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccDataSourceConfigLBACRules_basic(t *testing.T) {
+	testutils.CheckEnterpriseTestsEnabled(t, ">=11.4.0")
```
IevaVasiljeva: Should also be 11.5.0.

IevaVasiljeva (approved):
Nice, I've added some small comments, but once those are fixed it looks ready to me 👌

internal/resources/grafana/resource_data_source_config_lbac_rules.go
eleijonmarck: wait it is unused, i should really start to make my structs private on first creation. how do you spot this? with the linter

examples/resources/grafana_data_source_config_lbac_rules/resource.tf
eleijonmarck: it's not needed at all. just showing that you need to have a password to connect with the datasource from basicAuth

eleijonmarck:
Tested feature fully ✅ :

- rejects the `teamHttpHeaders` field in json when trying to update a json field in the datasource resource.
- do not change state when data source was provisioned via terraform and users manually adds a rule / removes rules
- correctly adds resource
- correctly delete resource
- correctly updates/changes the resource, if user adds more rules

---
This feature would be issued together with the release of Grafana version 11.5 in 28th of January

PR #1800, opened 2024-09-10
fix: Use a service account token instead of an API key

dmihai:
The unit test `TestAccFolder_createFromDifferentRoles` had a dependency on API keys which are deprecated. Also the endpoint for creating API keys has been removed.

In this PR, the test has been updated to work with service account tokens instead.


eleijonmarck (approved):
⚡

PR #1802, opened 2024-09-12
docs: use custom templates for ML resource docs

sd2k:
Using custom templates, similar to SLO and Synthetic Monitoring,
allows us to include multiple examples in the documentation of
each resource, along with comments for each resource.


sd2k:
The CI failures are due to some SLO tests failing; I've [raised it in Slack.](https://raintank-corp.slack.com/archives/C03G40X1ZK6/p1726135839437849)

gerboland (approved):

PR #1828, opened 2024-10-02
Add common provider headers to cloud-provider-client

thepalbi:
This PR adds provider headers the cloud-provider-client. This are basically:
- `Grafana-Terraform-Provider`: true if request coming from provider, allows us in the API to understand a request comes from the provider
- `Grafana-Terraform-Provider-Version`: contains version of provider, might be useful to propagate it through context and log it to debug provider issues

Also, once we have this, we can identify from which client a request is coming from, and then, store the "provenance" of a resource. Using here the same concept SLO uses https://github.com/grafana/slo/blob/346637b4ce113db1bc910f46fc397c72943e3426/pkg/api/slo.go#L410, which sounds just right.

### DIFF
--- a/internal/common/cloudproviderapi/client.go
+++ b/internal/common/cloudproviderapi/client.go
@@ -14,9 +14,10 @@ import (
 )
 
 type Client struct {
-	authToken string
-	apiURL    url.URL
-	client    *http.Client
+	authToken      string
+	apiURL         url.URL
+	client         *http.Client
+	defaultHeaders map[string]string
 }
 
 const (
@@ -24,7 +25,7 @@ const (
 	defaultTimeout = 90 * time.Second
 )
 
-func NewClient(authToken string, rawAPIURL string, client *http.Client) (*Client, error) {
+func NewClient(authToken string, rawAPIURL string, client *http.Client, defualtHeaders map[string]string) (*Client, error) {
 	parsedAPIURL, err := url.Parse(rawAPIURL)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse Cloud Provider API url: %w", err)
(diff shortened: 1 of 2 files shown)


internal/common/cloudproviderapi/client.go
```diff
@@ -14,17 +14,18 @@ import (
...
+	defaultHeaders map[string]string
 }
 
 const (
 	defaultRetries = 3
 	defaultTimeout = 90 * time.Second
 )
 
-func NewClient(authToken string, rawAPIURL string, client *http.Client) (*Client, error) {
+func NewClient(authToken string, rawAPIURL string, client *http.Client, defualtHeaders map[string]string) (*Client, error) {
```
Duologic:
```suggestion
func NewClient(authToken string, rawAPIURL string, client *http.Client, defaultHeaders map[string]string) (*Client, error) {
```

internal/common/cloudproviderapi/client.go
```diff
@@ -38,9 +39,10 @@ func NewClient(authToken string, rawAPIURL string, client *http.Client) (*Client
 	}
 
 	return &Client{
-		authToken: authToken,
-		apiURL:    *parsedAPIURL,
-		client:    client,
+		authToken:      authToken,
+		apiURL:         *parsedAPIURL,
+		client:         client,
+		defaultHeaders: defualtHeaders,
```
Duologic:
```suggestion
		defaultHeaders: defaultHeaders,
```

Duologic (changes requested):

thepalbi:
@Duologic changes applied!

Duologic (approved):

tristanburgess (approved):

PR #1829, opened 2024-10-02
cloudprovider: update CW scrape job regions field description text

tristanburgess:
Update description to specify that job regions must be a subset of the associated account resource's regions.

### DIFF
--- a/internal/resources/cloudprovider/resource_aws_cloudwatch_scrape_job.go
+++ b/internal/resources/cloudprovider/resource_aws_cloudwatch_scrape_job.go
@@ -96,7 +96,7 @@ func (r *resourceAWSCloudWatchScrapeJob) Schema(ctx context.Context, req resourc
 				Required:    true,
 			},
 			"regions": schema.SetAttribute{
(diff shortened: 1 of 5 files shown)


internal/resources/cloudprovider/resource_aws_cloudwatch_scrape_job.go
```diff
@@ -96,7 +96,7 @@ func (r *resourceAWSCloudWatchScrapeJob) Schema(ctx context.Context, req resourc
 				Required:    true,
 			},
 			"regions": schema.SetAttribute{
-				Description: "A set of AWS region names that this CloudWatch Scrape Job applies to.",
+				Description: "A set of AWS region names that this CloudWatch Scrape Job applies to. This must be a subset of the regions that are configured in the associated AWS Account resource.",
```
thepalbi, line 99: should we alter this to be called `regions_override`, and change the behaviour to allow null and empty values as we discused yesterday?

thepalbi (approved):
lgtm, disgeragard the comment

PR #1836, opened 2024-10-08
grafana_oncall_escalation: Add `declare_incident` option

matiasb:
Related to https://github.com/grafana/irm/issues/4519


internal/resources/oncall/resource_escalation.go
```diff
@@ -216,6 +217,21 @@ func resourceEscalation() *common.Resource {
...
+					"notify_on_call_from_schedule",
+					"persons_to_notify",
+					"persons_to_notify_next_each_time",
+					"notify_to_team_members",
+					"notify_if_time_from",
+					"notify_if_time_to",
+					"action_to_trigger",
+				},
+				Description: "The severity of the incident for declare_incident type step.",
+			},
```
joeyorlando, line 234: the value of this is essentially an enum, right? If so, I'd suggest adding a `ValidateFunc` here, like [this](https://github.com/grafana/terraform-provider-grafana/pull/1836/files#diff-f690fc0017aaf007bb7b6f788fc54da577d98cc99a7960d9def49d579a8f2ae9R67-R72):
```go
ValidateFunc: validation.StringInSlice(severityOptions, false),
Description:  fmt.Sprintf("The severity of the incident for declare_incident type step. Can be %s", severityOptionsVerbal),
```

joeyorlando (approved):

internal/resources/oncall/resource_escalation.go
matiasb: We have similar enumerations for each possible param, but each list the specific conflicting params which are different in each case. I think for now I would prefer having the list explicit and readable for each param? (I can take a look at defining multiple enums and refactoring in a separate PR too)

joeyorlando: ohh I see, sounds more involved, was just curious 👍

PR #1838, opened 2024-10-09
SSO: Convert numeric durations to strings in SSO settings

dmihai:
When importing SAML settings into Terraform, it is possible that some duration fields coming from the API to have a numeric type. This will generate an error because their type in Terraform is string:

```[ERROR] setting state: saml_settings.0.max_issue_delay: '' expected type 'string', got unconvertible type 'float64', value: '3e+11'```

This fix will convert all numeric durations from the settings map to strings.

### DIFF
--- a/internal/resources/grafana/resource_sso_settings.go
+++ b/internal/resources/grafana/resource_sso_settings.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/url"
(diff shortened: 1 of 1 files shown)


mgyongyosi (approved):
Thanks, it works like a charm! Tested locally and the `max_issue_delay` and `metadata_valid_duration` are handled correctly. 🎉

PR #1849, opened 2024-10-18
Add enabled to metrics endpoint scrape job resource

fridgepoet:
Thanks to Tristan's work on https://github.com/grafana/cloud-onboarding/pull/7841, we can make use of the "enabled" field on resource Create and Update.
This PR puts back the ability to define the boolean "enabled" field for metrics endpoint scrape jobs. 

Fixes https://github.com/grafana/cloud-onboarding/issues/7842

### DIFF
--- a/internal/resources/connections/resource_metrics_endpoint_scrape_job.go
+++ b/internal/resources/connections/resource_metrics_endpoint_scrape_job.go
@@ -12,6 +12,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/booldefault"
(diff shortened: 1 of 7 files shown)


internal/resources/connections/resource_metrics_endpoint_scrape_job.go
```diff
@@ -80,6 +81,12 @@ func (r *resourceMetricsEndpointScrapeJob) Schema(ctx context.Context, req resou
 					stringplanmodifier.RequiresReplace(),
 				},
 			},
+			"enabled": schema.BoolAttribute{
```
matthewnolf, line 84: There is also the `disabled_reason` field, which I believe we can add as a computed field?

fridgepoet: Following up what we discussed:
We'll shelve this for later. Captured some thoughts here, please feel free to add to this if I missed anything https://github.com/grafana/cloud-onboarding/issues/7854

matthewnolf (approved):

PR #1859, opened 2024-10-25
Update OnCall client setup to support service account auth

matiasb:
Related to https://github.com/grafana/irm/issues/997

Requires https://github.com/grafana/amixr-api-go-client/pull/24 (and a client release + updating dep here)


joeyorlando (approved):
overall LGTM. Should we update the provider documentation on service account vs oncall api token auth here? Do you think this warrants a major-version bump of the provider?

matiasb:
> overall LGTM. Should we update the provider documentation on service account vs oncall api token auth here? Do you think this warrants a major-version bump of the provider?

Yeah, I will be updating some docs here before merging (still need to release the updated API client and push some changes here, besides I want to do some additional testing too). About the provider version, not sure who/how that is decided (in any case we are not breaking any compatibility just extending the auth support).

Also I would like to update some [other docs](https://grafana.com/docs/grafana-cloud/developer-resources/infrastructure-as-code/terraform/) I found re OnCall and Terraform once I confirm everything working as expected.

PR #1869, opened 2024-10-29
cloudproviderapi client: remove unnecessary pointer receivers

tristanburgess:


thepalbi (changes requested):
We should keep the pointer types in here, since clients / services / or structs whose purpose is not just a data container, should be a pointer type. That allows all the code to perform the calls over a pointer to the instance, rather than copying around all it's contents. At least, that's the practice I'm used to, and that we use in all our codebases.

PR #1881, opened 2024-11-01
cloudprovider: aws acceptance tests

tristanburgess:
this PR implements https://github.com/grafana/grafana-csp-app/issues/63
it adds acceptance tests for major provider functionality for creating Grafana AWS Account resources and CloudWatch Scrape Jobs for the Grafana Cloud Provider app.

### DIFF
--- a/internal/resources/cloudprovider/data_source_aws_cloudwatch_scrape_job_test.go
+++ b/internal/resources/cloudprovider/data_source_aws_cloudwatch_scrape_job_test.go
@@ -1,32 +1,61 @@
 package cloudprovider_test
 
 import (
+	"context"
 	"fmt"
+	"os"
 	"testing"
 
+	"github.com/grafana/terraform-provider-grafana/v3/internal/common"
+	"github.com/grafana/terraform-provider-grafana/v3/internal/common/cloudproviderapi"
 	"github.com/grafana/terraform-provider-grafana/v3/internal/testutils"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/stretchr/testify/require"
 )
 
 func TestAccDataSourceAWSCloudWatchScrapeJob(t *testing.T) {
-	t.Skip("Skipping test until we have a valid test case")
 	testutils.CheckCloudInstanceTestsEnabled(t)
+
+	// Uses a pre-existing account resource so that we don't need to create a new account resource for every test run.
+	accountID := os.Getenv("GRAFANA_CLOUD_PROVIDER_TEST_AWS_ACCOUNT_RESOURCE_ID")
+	require.NotEmpty(t, accountID, "GRAFANA_CLOUD_PROVIDER_TEST_AWS_ACCOUNT_RESOURCE_ID must be set")
+
+	roleARN := os.Getenv("GRAFANA_CLOUD_PROVIDER_AWS_ROLE_ARN")
+	require.NotEmpty(t, roleARN, "GRAFANA_CLOUD_PROVIDER_AWS_ROLE_ARN must be set")
+
+	stackID := os.Getenv("GRAFANA_CLOUD_PROVIDER_TEST_STACK_ID")
+	require.NotEmpty(t, stackID, "GRAFANA_CLOUD_PROVIDER_TEST_STACK_ID must be set")
+
+	// Make sure the account exists and matches the role ARN we expect for testing
+	client := testutils.Provider.Meta().(*common.Client).CloudProviderAPI
+	gotAccount, err := client.GetAWSAccount(context.Background(), stackID, accountID)
+	require.NoError(t, err)
+	require.Equal(t, roleARN, gotAccount.RoleARN)
+
+	var gotJob cloudproviderapi.AWSCloudWatchScrapeJobResponse
+
+	jobName := "test-job" + acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum)
 
 	resource.ParallelTest(t, resource.TestCase{
 		ProtoV5ProviderFactories: testutils.ProtoV5ProviderFactories,
-		CheckDestroy: func() resource.TestCheckFunc {
-			return func(s *terraform.State) error {
-				return nil
-			}
-		}(),
 		Steps: []resource.TestStep{
 			{
-				Config: awsCloudWatchScrapeJobDataSourceData(),
+				Config: awsCloudWatchScrapeJobResourceData(stackID,
+					jobName,
+					false,
+					accountID,
+					regionsString(testAWSCloudWatchScrapeJobData.RegionsSubsetOverride),
+					servicesString(testAWSCloudWatchScrapeJobData.Services),
+					customNamespacesString(testAWSCloudWatchScrapeJobData.CustomNamespaces),
+				) + awsCloudWatchScrapeJobDataSourceData(stackID, jobName),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr("data.grafana_cloud_provider_aws_cloudwatch_scrape_job.test", "stack_id", testStackID),
-					resource.TestCheckResourceAttr("data.grafana_cloud_provider_aws_cloudwatch_scrape_job.test", "name", testAWSCloudWatchScrapeJobData.Name),
-					resource.TestCheckResourceAttr("data.grafana_cloud_provider_aws_cloudwatch_scrape_job.test", "aws_account_resource_id", testAWSCloudWatchScrapeJobData.AWSAccountResourceID),
+					checkAWSCloudWatchScrapeJobResourceExists("grafana_cloud_provider_aws_cloudwatch_scrape_job.test", stackID, &gotJob),
+					resource.TestCheckResourceAttr("data.grafana_cloud_provider_aws_cloudwatch_scrape_job.test", "stack_id", stackID),
+					resource.TestCheckResourceAttr("data.grafana_cloud_provider_aws_cloudwatch_scrape_job.test", "name", jobName),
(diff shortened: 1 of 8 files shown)


internal/resources/cloudprovider/data_source_aws_cloudwatch_scrape_job_test.go
```diff
@@ -57,19 +64,41 @@ func TestAccDataSourceAWSCloudWatchScrapeJob(t *testing.T) {
...
 	})
 }
 
-func awsCloudWatchScrapeJobDataSourceData() string {
+func awsCloudWatchScrapeJobDataSourceData(stackID string, jobName string) string {
 	data := fmt.Sprintf(`
 data "grafana_cloud_provider_aws_cloudwatch_scrape_job" "test" {
 	stack_id = "%[1]s"
-	name = "%[2]s"
+	name = grafana_cloud_provider_aws_cloudwatch_scrape_job.test.name
```
thepalbi, line 97: shouldn't this be the `jobName` argument?

internal/resources/cloudprovider/data_source_aws_cloudwatch_scrape_jobs_test.go
```diff
@@ -2,75 +2,61 @@ package cloudprovider_test
...
+		CheckDestroy: checkAWSCloudWatchScrapeJobResourceDestroy(testCfg.stackID, &gotJob),
 	})
 }
 
-func awsCloudWatchScrapeJobsDataSourceData() string {
+func awsCloudWatchScrapeJobsDataSourceData(stackID string) string {
 	data := fmt.Sprintf(`
 data "grafana_cloud_provider_aws_cloudwatch_scrape_jobs" "test" {
 	stack_id = "%[1]s"
+	depends_on = [grafana_cloud_provider_aws_cloudwatch_scrape_job.test]
```
thepalbi, line 56: isn't this some kind of circular dependency?

internal/resources/cloudprovider/resource_aws_account_test.go
```diff
@@ -1,75 +1,116 @@
...
+			return fmt.Errorf("account id not set")
+		}
+
+		client := testutils.Provider.Meta().(*common.Client).CloudProviderAPI
+		gotAccount, err := client.GetAWSAccount(context.Background(), stackID, accountID)
+		if err != nil {
+			return fmt.Errorf("error getting account: %s", err)
+		}
+
+		*account = gotAccount
```
thepalbi, line 80: why in here overwrite the ptr value with the getted account?

internal/resources/cloudprovider/resource_aws_account_test.go
```diff
@@ -1,75 +1,116 @@
...
+					resource.TestCheckResourceAttr("grafana_cloud_provider_aws_account.test", "regions.0", account.Regions[0]),
+					resource.TestCheckResourceAttr("grafana_cloud_provider_aws_account.test", "regions.1", account.Regions[1]),
+					resource.TestCheckResourceAttr("grafana_cloud_provider_aws_account.test", "regions.2", account.Regions[2]),
 				),
 			},
+			{
+				ResourceName:      "grafana_cloud_provider_aws_account.test",
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
```
thepalbi, line 47: what's this bit?

internal/resources/cloudprovider/utils_test.go
```diff
@@ -0,0 +1,162 @@
...
+func servicesString(svcs []cloudproviderapi.AWSCloudWatchService) string {
+	if len(svcs) == 0 {
+		return ""
+	}
+	b := new(bytes.Buffer)
+	for _, svc := range svcs {
+		fmt.Fprintf(b, "\n\t\t")
+		fmt.Fprintf(b, `{
+			name = "%[1]s",
+			metrics = [%[2]s],
```
thepalbi, line 63: arent we using this notation for metrics, or are both supported:
```
    metric {
      name        = "CPUUtilization"
      statistics  = ["Average"]
    }

    metric {
      name        = "StatusCheckFailed"
      statistics  = ["Maximum"]
    }
```

internal/resources/cloudprovider/resource_aws_cloudwatch_scrape_job_test.go
```diff
@@ -173,33 +225,83 @@ func TestAccResourceAWSCloudWatchScrapeJob(t *testing.T) {
...
+		if !ok {
+			return fmt.Errorf("resource not found: %s\n %#v", rn, s.RootModule().Resources)
+		}
+
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("resource id not set")
+		}
+
+		parts := strings.SplitN(rs.Primary.ID, ":", 2)
+		if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
```
thepalbi, line 255: should we perform some assertion over the first part of the key?

internal/resources/cloudprovider/data_source_aws_cloudwatch_scrape_job_test.go
tristanburgess: IMO no, because we pass `jobName` into the scrape job resource to create, then reference that resource in this data source block, which makes Terraform depend on the resource to be created before doing the lookup.

internal/resources/cloudprovider/data_source_aws_cloudwatch_scrape_jobs_test.go
tristanburgess: it is not, because the Terraform resource block definition for this test's config, which comes from https://github.com/grafana/terraform-provider-grafana/blob/439626c56f114eca1f8b216d980e18b15c20130d/internal/resources/cloudprovider/resource_aws_account_test.go#L104 doesn't depend on any of the data source block's attributes. This `depends_on` is needed for that reason; without it, the data source would likely not pick up the resource's job because it'd be free to execute concurrently with the resource creation.

internal/resources/cloudprovider/resource_aws_account_test.go
tristanburgess: this is a pattern that I copied from other packages in this repo, e.g. https://github.com/grafana/terraform-provider-grafana/blob/ad25130f6220d560d0247c98b9767b8e6960ad08/internal/resources/cloud/resource_cloud_stack_test.go#L221

the idea is that we want to check the API for the resource to no longer exist in order to ensure that it is destroyed via https://github.com/grafana/terraform-provider-grafana/blob/439626c56f114eca1f8b216d980e18b15c20130d/internal/resources/cloudprovider/resource_aws_account_test.go#L49 in order to do that, we need the ID of the resource. 

The problem is that the way the tests are set up is declarative: the Terraform config comes from https://github.com/grafana/terraform-provider-grafana/blob/439626c56f114eca1f8b216d980e18b15c20130d/internal/resources/cloudprovider/resource_aws_account_test.go#L32 and then a number of checks are run against that resource config after applied before it is destroyed at the end. Thus one way that is used throughout this repo to pass the resource ID into the `CheckDestroy` function is to have one of the test steps write the API object to a variable outside of the scope of the `resource.ParallelTest` instance https://github.com/grafana/terraform-provider-grafana/blob/439626c56f114eca1f8b216d980e18b15c20130d/internal/resources/cloudprovider/resource_aws_account_test.go#L26 so that it can be referenced by the step executed later (CheckDestroy) so that function can use the resource ID

internal/resources/cloudprovider/utils_test.go
tristanburgess: the purposes of these util functions are to convert the API structs into Terraform-encoded strings, e.g. [Terraform's object format](https://developer.hashicorp.com/terraform/language/expressions/type-constraints#object) that can be plugged into `for_each` blocks e.g. https://github.com/grafana/terraform-provider-grafana/blob/439626c56f114eca1f8b216d980e18b15c20130d/internal/resources/cloudprovider/resource_aws_cloudwatch_scrape_job_test.go#L308

I could've alternatively done a string template in go via a function or something to repeat the blocks myself directly in the resource definition. In my original PR implementing the provider from July or so, I set up the scrape job examples with these dynamic blocks so that users would be guided to plug in data from simple arrays instead of copying the block formats verbatim, and wanted the tests to match that pattern. I decided to keep the test structure implemented this way even though we simplified the format of the examples.

internal/resources/cloudprovider/resource_aws_cloudwatch_scrape_job_test.go
tristanburgess: yeah that's fair, will add just to be sure

internal/resources/cloudprovider/resource_aws_account_test.go
tristanburgess: it's implemented here https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/resource/testing_new_import_state.go

described fairly well here https://stackoverflow.com/a/69010198 and in TF docs https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/import, but basically this will clear the TF working directory/statefile from the previous steps, without destroying the resources etc, and then test importing the resource indicated by the ResourceName. `ImportStateVerify` holds onto the state resulting from the end of the previous steps and makes sure that the imported state matches the previous state.

internal/resources/cloudprovider/resource_aws_account_test.go
thepalbi: sounds good!

internal/resources/cloudprovider/data_source_aws_cloudwatch_scrape_jobs_test.go
thepalbi: ahhh got it!

internal/resources/cloudprovider/data_source_aws_cloudwatch_scrape_job_test.go
thepalbi: Imagine it's answer with same as below

thepalbi (approved):
SGTM

PR #1895, opened 2024-11-08
Teams: `resource_team` Add UID to teams for reads

eleijonmarck:
Adds the UID to be read from the resource
This enables lbac_rules to read the UIDs from teams.

Enables adding the resource `data_source_config_lbac_rules` - https://github.com/grafana/grafana-enterprise/pull/7469

### DIFF
--- a/docs/data-sources/team.md
+++ b/docs/data-sources/team.md
@@ -52,6 +52,7 @@ to the team. Note: users specified here must already exist in Grafana.
 - `team_sync` (List of Object) Sync external auth provider groups with this Grafana team. Only available in Grafana Enterprise.
 	* [Official documentation](https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-team-sync/)
 	* [HTTP API](https://grafana.com/docs/grafana/latest/developers/http_api/team_sync/) (see [below for nested schema](#nestedatt--team_sync))
+- `team_uid` (String) The team uid assigned to this team by Grafana.
 
 <a id="nestedatt--preferences"></a>
 ### Nested Schema for `preferences`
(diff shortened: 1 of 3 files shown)


IevaVasiljeva (commented):
Did you test what `plan` looks like for a team that was created before this change was introduced? I wonder if we should ignore the diff if UID was not set and then gets set for the first time, as this is not something that we expect users to do anything about, so showing a diff might be confusing.

eleijonmarck:
> Did you test what `plan` looks like for a team that was created before this change was introduced? I wonder if we should ignore the diff if UID was not set and then gets set for the first time, as this is not something that we expect users to do anything about, so showing a diff might be confusing.


# existing teams

nothing changes for existing teams

# new created teams for the plan will look like:

## previously
  // grafana_team.test-2 will be created
  + resource "grafana_team" "test-2" {
      + id      = (known after apply)
      + name    = "Test team 2"
      + team_id = (known after apply)
    }

## new introduce change
  // grafana_team.test-2 will be created
  + resource "grafana_team" "test-2" {
      + id       = (known after apply)
      + name     = "Test team 2"
      + team_id  = (known after apply)
      + team_uid = (known after apply)
    }

eleijonmarck:
it is the way to get the resource to get a good example of how to use the uid

```hcl
resource "grafana_data_source_config_lbac_rules" "test" {
    datasource_uid = grafana_data_source.test.uid
    rules = jsonencode({
        "${grafana_team.test.team_uid}" = [
            "{ foo != \"bar\", foo !~ \"baz\" }",
            "{ foo = \"qux\" }"
        ]
    })
    depends_on = [
        grafana_team.test,
        grafana_data_source.test
    ]
}
```

IevaVasiljeva (approved):
> nothing changes for existing teams

Ah nice! I thought that Terraform might show a diff, as team UID will "change" for the existing resources. But good to hear that it's not the case 👌

PR #1905, opened 2024-11-13
Chore: `go.mod` Update to latest openapi spec 

eleijonmarck:
Updating to the openapi-client from - https://github.com/grafana/grafana-openapi-client-go/commit/9cb2bbfeb8a38e28dc253534529b0e08b03c59c4


IevaVasiljeva:
Thank you so much for helping with the OpenAPI dependency update @spinillos 🏅

eleijonmarck:
@spinillos is a life savor

internal/resources/grafana/contentTypeRoundTripper.go
```diff
@@ -0,0 +1,36 @@
+package grafana
```
julienduchesne: File name should be snake_case and I think that if it's only used in tests, then it could live in a _test.go file

julienduchesne (approved):

internal/resources/grafana/contentTypeRoundTripper.go
spinillos: Ups!! Too much java during past week...

PR #1907, opened 2024-11-14
Add support for the Azure credentials resource

andriikushch:
This PR introduces the `grafana_cloud_provider_azure_credential` resource and data source to the Grafana Terraform provider. The new resource allows users to manage Azure credentials within their Grafana Cloud stack. The data source enables users to retrieve information about existing Azure credentials.


docs/resources/cloud_provider_azure_credential.md
```diff
@@ -0,0 +1,47 @@
...
+- `name` (String) The name of the Azure Credential.
+- `stack_id` (String) The StackID of the Grafana Cloud instance. Part of the Terraform Resource ID.
+- `tenant_id` (String) The tenant ID of the Azure Credential.
+
+### Read-Only
+
+- `id` (String) The Terraform Resource ID. This has the format "{{ stack_id }}:{{ resource_id }}".
+- `resource_id` (String) The ID given by the Grafana Cloud Provider API to this AWS Account resource.
+
+## Import
```
andriikushch, line 41: Import is not implemented.

internal/common/cloudproviderapi/client.go
```diff
@@ -193,6 +193,68 @@ func (c *Client) DeleteAWSCloudWatchScrapeJob(ctx context.Context, stackID strin
 	return nil
 }
 
+type AzureCredential struct {
```
kgeckhart, line 196: Should we model the `MetricCollectionConfiguration` at all or is it left off because it's part of the API response and terraform doesn't need to manage it/be aware of iit?

kgeckhart (approved):

internal/resources/cloudprovider/data_source_azure_credential.go
```diff
@@ -0,0 +1,144 @@
...
+			},
+			"resource_id": schema.StringAttribute{
+				Description: "The ID given by the Grafana Cloud Provider API to this Azure Credential resource.",
+				Required:    true,
+			},
+			"name": schema.StringAttribute{
+				Description: "The name of the Azure Credential.",
+				Required:    true,
+			},
+			"client_id": schema.StringAttribute{
```
tristanburgess, line 65: For a Terraform data source, we should only mark fields required if the user _must_ supply values for them to be able to look up a unique resource. All other fields should be marked only as `Computed: true`, see e.g. 

https://github.com/grafana/terraform-provider-grafana/blob/main/internal/resources/cloudprovider/resource_aws_account.go 

versus

https://github.com/grafana/terraform-provider-grafana/blob/main/internal/resources/cloudprovider/data_source_aws_account.go

internal/resources/cloudprovider/resource_azure_credential.go
```diff
@@ -0,0 +1,350 @@
...
+
+type resourceAzureCredentialModel struct {
+	ID                 types.String `tfsdk:"id"`
+	Name               types.String `tfsdk:"name"`
+	TenantID           types.String `tfsdk:"tenant_id"`
+	ClientID           types.String `tfsdk:"client_id"`
+	StackID            types.String `tfsdk:"stack_id"`
+	ClientSecret       types.String `tfsdk:"client_secret"`
+	ResourceID         types.String `tfsdk:"resource_id"`
+	ResourceTagFilters []TagFilter  `tfsdk:"resource_tag_filter"`
```
tristanburgess: I've recently found out that if we use Go slice types directly here, instead of the Terraform type e.g. [List](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/types/list), you can get errors when a value is planned by Terraform with a state other than the normal state e.g. `Unknown`. See https://github.com/grafana/cloud-onboarding/issues/8084. I'm still working through what the right pattern is myself, possibly handing off to @thepalbi on Wednesday, but this will need to be fixed here as well.

internal/resources/cloudprovider/resource_azure_credential.go
```diff
@@ -0,0 +1,350 @@
...
+				Required:    true,
+			},
+			"client_secret": schema.StringAttribute{
+				Description: "The client secret of the Azure Credential.",
+				Required:    true,
+				Sensitive:   true,
+			},
+		},
+		Blocks: map[string]schema.Block{
+			"resource_tag_filter": schema.ListNestedBlock{
```
tristanburgess: https://github.com/grafana/terraform-provider-grafana/blob/9fb114825d75198a812a3bf6cfd3827d4d613d66/internal/resources/cloudprovider/resource_aws_cloudwatch_scrape_job.go#L167 calls a similar field `resource_discovery_tag_filter`, should we make things consistent, or is it an Azure thing?

internal/resources/cloudprovider/azure_credential_test.go
```diff
@@ -0,0 +1,151 @@
...
+	"testing"
+
+	"github.com/grafana/terraform-provider-grafana/v3/internal/testutils"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+// Tests both managed resource and data source
+func TestAcc_AzureCredential(t *testing.T) {
+	resourceID := "3"
+	// Mock the Connections API response for Create, Get, and Delete
```
tristanburgess: comment leftover from copy

internal/resources/cloudprovider/resource_azure_credential.go
andriikushch: done

examples/data-sources/grafana_cloud_provider_azure_credential/data-source.tf
```diff
@@ -0,0 +1,32 @@
...
+    value = "value-2"
+  }
+}
+
+
+data "grafana_cloud_provider_azure_credential" "test" {
+  stack_id    = grafana_cloud_provider_azure_credential.test.stack_id
+  resource_id = grafana_cloud_provider_azure_credential.test.resource_id
+
+  resource_discovery_tag_filter {
```
tristanburgess: does this + the docs need to be updated to remove these blocks from the data source example?

internal/resources/cloudprovider/data_source_azure_credential.go
```diff
@@ -0,0 +1,144 @@
...
+				Computed:    true,
+			},
+			"client_secret": schema.StringAttribute{
+				Description: "The client secret of the Azure Credential.",
+				Computed:    true,
+				Sensitive:   true,
+			},
+		},
+		Blocks: map[string]schema.Block{
+			"resource_discovery_tag_filter": schema.ListNestedBlock{
```
tristanburgess, line 80: are these filters required to look up a unique resource? I would think that just the stack and resource ID are sufficient, because it looks like the client `GetAzureCredential` function only needs the stack ID and resource ID. I think the inner fields need to be marked with Computed instead of Required, see e.g. https://github.com/grafana/terraform-provider-grafana/blob/04bb8636bdfc8b919b84631fe42a349d86fea8f0/internal/resources/cloudprovider/data_source_aws_cloudwatch_scrape_job.go#L61

tristanburgess: if so, then it looks like `resource_discovery_tag_filter` needs to be retrieved and set below in the Read function from the API response as well

examples/data-sources/grafana_cloud_provider_azure_credential/data-source.tf
andriikushch: Removed 👍

internal/resources/cloudprovider/data_source_azure_credential.go
andriikushch: Done

tristanburgess (approved):

PR #1909, opened 2024-11-15
SM Checks: Select probes automatically

julienduchesne:
Crossplane doesn't have datasources, so we have to specify probe IDs manually 
This will allow us to specify a number of probes to be selected randomly

I've updated the basic examples so that they use this attribute. The advanced examples still use explicit probes

### DIFF
--- a/docs/resources/synthetic_monitoring_check.md
+++ b/docs/resources/synthetic_monitoring_check.md
@@ -26,15 +26,11 @@ multiple checks for a single endpoint to check different capabilities.
 ### DNS Basic
 
 ```terraform
-data "grafana_synthetic_monitoring_probes" "main" {}
-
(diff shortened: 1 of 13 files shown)


peterschretlen (commented):
No objections to the change itself.  Looks ok to me, though a comment:

I can see a case for making this simple when people don't care where the probes are located.  

Usually there are some criteria around probe selection though:  US only / Europe only;  private probes only.  Particular if customers have a mix of public and private probes, the random selection may not see much use, or we may see requests to limit it somehow to a subset.

julienduchesne:
> Usually there are some criteria around probe selection though: US only / Europe only; private probes only. Particular if customers have a mix of public and private probes, the random selection may not see much use, or we may see requests to limit it somehow to a subset.

The `probes` attribute keeps working the same as before, so users can select the probes as they wish with it

PR #1922, opened 2024-11-26
Upgrade amixr-api-go-client package to 0.0.17

sagarp337:
1. upgrading the **amixr-api-go-clien**t package transitive dependent package **go-retryablehttp** v0.6.6 has a vulnerability. The issue has been addressed in amixr-api-go-client v0.0.17
2. Upgrading all go packages


Duologic:
Could you limit this PR to only update the amixr lib? No need to update all go packages.

PR #1923, opened 2024-11-27
Add http compression option to synthetic monitoring check resource

peterschretlen:
This option - which validates check the compression algorithm used in the response - is available in the UI and the API, but was not exposed in the terraform provider. 

Fixes https://github.com/grafana/synthetic-monitoring/issues/184

The absence of a compression value, compression = "", and compression=none mean the same thing: "no compression validation is being done".

But we need to chose one of these representations when reading and writing http check settings.  In this PR I opted for removing the compression value altogether when compression = none.  This preserves the current behaviour if no compression value is specified (and the integration tests fail if we represent it with compression = none ).

### DIFF
--- a/internal/resources/syntheticmonitoring/resource_check.go
+++ b/internal/resources/syntheticmonitoring/resource_check.go
@@ -980,7 +980,9 @@ func resourceCheckRead(ctx context.Context, d *schema.ResourceData, c *smapi.Cli
 			"cache_busting_query_param_name":    chk.Settings.Http.CacheBustingQueryParamName,
 		})
 
-		if chk.Settings.Http.Compression != nil {
+		// The default compression "none" is the same as omitting the value.
+		// Since this value is usually not explicitly set, omit when set to "none"
+		if chk.Settings.Http.Compression != sm.CompressionAlgorithm_none {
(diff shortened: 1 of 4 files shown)


internal/resources/syntheticmonitoring/resource_check.go
```diff
@@ -341,6 +341,11 @@ var (
 				Optional:    true,
 				Elem:        syntheticMonitoringCheckSettingsHTTPHeaderMatch,
 			},
+			"compression": {
+				Description: "Check fails if the response body is not compressed using this compression algorithm. One of `none`, `identity`, `br`, `gzip`, `deflate`.",
+				Type:        schema.TypeString,
+				Optional:    true,
```
spinillos: You can use `ValidateFunc` to ensure that you are receiving an expected value. This is an [example](https://github.com/grafana/terraform-provider-grafana/blob/main/internal/resources/grafana/resource_report.go#L118).

peterschretlen: Thank you @spinillos!  I've added a validation function,  example output of running the plan with an invalid compression setting:

image

spinillos (approved):

PR #1943, opened 2024-12-06
SLO tests: make less flaky.

joeblubaugh:
* Runs SLO tests serially, making plan interference less likely.
* Adds a short delay after an out-of-band SLO delete, to allow background workers to clean up the resources. This should reduce flakiness when planning after the deletion.


Duologic (approved):

elainevuong (approved):
LGTM

PR #1945, opened 2024-12-06
Fix CleanResourceName for IDs starting with a - characted

froque:
This should make IDs compatible with Terraform syntax for identifiers:

> Identifiers
> 
> Argument names, block type names, and the names of most Terraform-specific constructs like resources, input variables, etc. are all identifiers.
> 
> Identifiers can contain letters, digits, underscores (_), and hyphens (-). The first character of an identifier must not be a digit, to avoid ambiguity with literal numbers.
> 
> For complete identifier rules, Terraform implements [the Unicode identifier syntax](http://unicode.org/reports/tr31/), extended to include the ASCII hyphen character -.

https://developer.hashicorp.com/terraform/language/syntax/configuration#identifiers

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ provider_installation {
 Acceptance tests require a running instance of Grafana. You can either handle
 running an instance of Grafana yourself or use `docker-compose`.
 
-If you choose `docker-compose`, run `make testacc-docker`. This is the simplest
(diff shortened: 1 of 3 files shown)


Duologic (commented):
Can you add a description?

froque:
> Can you add a description?

done

PR #1949, opened 2024-12-10
Doc: Update `service_account_token` example

GregSharpe1:
Ensure the `stack_slug` is within the example within the documentation as it's required.


Duologic:
Can you sign the CLA?

Duologic (approved):

volcanonoodle (dismissed):

volcanonoodle:
Thank you for the fix! 🙏 
I just remembered that the docs are autogenerated from the examples, so the file to modify is actually 
https://github.com/grafana/terraform-provider-grafana/blob/55da61efbb974696f6f5b1b5fbf9e737922ae71f/examples/resources/grafana_cloud_stack_service_account_token/resource.tf#L9-L12
Then you can run `go generate` and the docs will be updated  🙂

PR #1952, opened 2024-12-12
cloudprovider: fix list base types for AWS scrape jobs

tristanburgess:
This fixes a bug where the provider returns an error when trying to define the service blocks using a for_each construct, and feeding the for_each with a complex type defined in variables. Terraform seems to treat this situation as needing to compute the list result at apply-time, but the provider currently implemented lists using Go slice types, resulting in the following error:

```
╷
│ Error: Value Conversion Error
│
│   with module.monitoring.grafana_cloud_provider_aws_cloudwatch_scrape_job.main,
│   on modules/test/main.tf line 16, in resource "grafana_cloud_provider_aws_cloudwatch_scrape_job" "main":
│   16: resource "grafana_cloud_provider_aws_cloudwatch_scrape_job" "main" {
│
│ An unexpected error was encountered trying to build a value. This is always an error in the provider. Please report the following to the provider
│ developer:
│
│ Received unknown value, however the target type cannot handle unknown values. Use the corresponding `types` package type or a custom type that
│ handles unknown values.
│
│ Path:
│ Target Type: []cloudprovider.awsCWScrapeJobServiceTFModel
│ Suggested Type: basetypes.ListValue
╵
```

This PR changes the Terraform model structs to use Terraform types for Lists, which support Terraform state values such as Unknown.


internal/resources/cloudprovider/models.go
```diff
@@ -41,30 +42,82 @@ type awsCWScrapeJobTFDataSourceModel struct {
...
-	ResourceDiscoveryTagFilters []awsCWScrapeJobTagFilterTFModel `tfsdk:"resource_discovery_tag_filter"`
-	TagsToAddToMetrics          types.Set                        `tfsdk:"tags_to_add_to_metrics"`
+	Name                        types.String `tfsdk:"name"`
+	Metrics                     types.List   `tfsdk:"metric"`
+	ScrapeIntervalSeconds       types.Int64  `tfsdk:"scrape_interval_seconds"`
+	ResourceDiscoveryTagFilters types.List   `tfsdk:"resource_discovery_tag_filter"`
+	TagsToAddToMetrics          types.Set    `tfsdk:"tags_to_add_to_metrics"`
+}
+
+func (m awsCWScrapeJobServiceTFModel) attrTypes() map[string]attr.Type {
```
thepalbi, line 57: what are this for?

thepalbi (approved):

internal/resources/cloudprovider/models.go
tristanburgess: this is for plugging into Terraform schema validation in the conversion functions where the inner type of the List is the generic TF Object type. The map lets TF validate conformity with the Go-type-specific schema, if that makes sense.

thepalbi: Yep, makes sense, thanks for explaining!

Issue #1953, opened 2024-12-13
Failed to create Clients - Token required


SeamusGrafana:
### Terraform Grafana Provider Version

3.15.0

### Grafana Version

Cloud

### Affected Resource(s)

All ? (Anything that uses the Provider with a Stack Token that is known after apply)

### Terraform Configuration Files

```hcl
# Grafana Provider
## Grafana Cloud Provider Configuration
provider "grafana" {
  alias                     = "cloud"
  cloud_access_policy_token = var.grafana_cloud_key
}

## Grafana Cloud Stack Provider Configuration
provider "grafana" {
  alias                     = "stack"
  auth                      = module.grafana_cloud_stack.grafana_cloud_stack_sa_terraform_key
  cloud_access_policy_token = var.grafana_cloud_key
  sm_access_token           = module.grafana_cloud_stack.grafana_cloud_stack_sm_access_token
  sm_url                    = module.grafana_cloud_stack.grafana_cloud_stack_sm_url
  url                       = module.grafana_cloud_stack.grafana_cloud_stack_url
}
```

Module:

```hcl
# Create the Grafana Cloud Stack
resource "grafana_cloud_stack" "stack" {
  labels      = var.grafana_cloud_stack_labels
  name        = var.grafana_cloud_stack_slug
  region_slug = var.grafana_cloud_stack_region
  slug        = var.grafana_cloud_stack_slug
}

# Create a Service Account and Token/Key for the Stack
resource "grafana_cloud_stack_service_account" "terraform_sa" {
  stack_slug = grafana_cloud_stack.stack.slug

  name        = "sa-terraform-as-code"
  role        = "Admin"
  is_disabled = false
}
resource "grafana_cloud_stack_service_account_token" "terraform_sa" {
  stack_slug = grafana_cloud_stack.stack.slug

  name               = "sa-terraform-as-code-key"
  service_account_id = grafana_cloud_stack_service_account.terraform_sa.id
}

# Create Access Policy and Token for Synthetic Monitoring
resource "grafana_cloud_access_policy" "sm_metrics_publish" {
  region       = var.grafana_cloud_stack_region
  name         = "${var.grafana_cloud_stack_slug}-sm-publisher"
  display_name = "[${var.grafana_cloud_stack_slug}] - Synthetic Monitoring Metric Publisher"

  scopes = ["logs:write", "metrics:write", "stacks:read", "traces:write"]

  realm {
    type       = "stack"
[...]

PR #1958, opened 2024-12-16
Enable oncall client setup using service account auth

matiasb:
Related to https://github.com/grafana/terraform-provider-grafana/issues/1953

Updated OnCall API client to allow empty string tokens on setup (see https://github.com/grafana/amixr-api-go-client/pull/27).


docs/index.md
```diff
@@ -201,6 +201,46 @@ resource "grafana_oncall_escalation" "example_notify_step" {
...
 ```
 
+Alternatively, you can also configure the provider block by setting `url`
+to your Grafana URL and `auth` to a service account token:
+
+```terraform
+// Step 1: Configure provider block.
+provider "grafana" {
+  alias = "oncall"
+  url   = "http://grafana.example.com/"
```
joeyorlando, line 211: interesting, so we no longer need to make the Terraform provider "aware" of the OnCall API url?

joeyorlando (approved):

docs/index.md
matiasb: Not really, but there was always a default value for it (`prod-us-central-0`), so you can omit setting it in some cases (see existing [docs](https://registry.terraform.io/providers/grafana/grafana/latest/docs#managing-grafana-oncall)). It would be great not to need this URL, but there isn't an easy way to figure it out atm (afaict).

joeyorlando: > It would be great not to need this URL, but there isn't an easy way to figure it out atm

In hosted grafana land we could _maybe_ call the GCOM region API to fetch your stack's oncall regional API url? but yeah this won't support OSS

spinillos (approved):

Issue #1928, opened 2024-12-02
grafana_rule_group rule block ordering reassigns rule UIDs


JacobsonMT:
### Terraform Version

1.10.0

### Terraform Grafana Provider Version

3.13.1

### Grafana Version

main (11.4.0-pre)

### Affected Resource(s)

* grafana_rule_group

### Terraform Configuration Files

```hcl
resource "grafana_rule_group" "rule_group_527ddb6599bf18b3" {
  name             = "1m"
  folder_uid       = grafana_folder.my_folder.uid
  interval_seconds = 60

  rule {
    name      = "example alert"
    is_paused = true
    condition = "A"
    data {
      ref_id = "A"
      datasource_uid = "__expr__"
      relative_time_range {
        from = 600
        to   = 0
      }
      model = "{}"
    }
  }
  rule {
    name      = "example alert 2"
    is_paused = true
    condition = "B"
    data {
      ref_id = "B"
      datasource_uid = "__expr__"
      relative_time_range {
        from = 600
        to   = 0
      }
      model = "{}"
    }
  }
}
```

### Expected Behavior

Modifications to a `grafana_rule_group` that alter the ordering of rules would preferably retain the connection between a rule and its computed UID. This may be difficult to do in general given how alert rules are not resources and are instead modelled as `LIST` type blocks under a group. So, at the very least the following should hold true:

- Modifications to alert rule ordering should not unintentionally allow a rule to obtain the UID of an existing rule.
- There should be a way to (at least optionally) ensure a rule definition does not change UIDs when rules are reordered. 

### Actual Behavior

Modifying rule ordering in a group can unintentionally reassign a rule's UID to that of a different rule in the group. There are multiple ways this can happen, some examples:

- Deleting a rule in the middle of a group: will reassign the UID all rules that follow it to that of the rule in it's `idx-1`
- Adding a rule in the middle of a group: will give the new rule the UID of the previous rule at that index and reassign all following rules to that of the rule in it's `idx+1`. The last rule will obtain a newly generated UID.
- Swapping two rules: will swap their UIDs.

### Steps to Reproduce

1. Apply the above provided state.
2. Note the UIDs of the created rules.
3. Delete the `example alert` block.
4. Apply
5. Note that `example alert 2` rule now has the UID of the deleted `example alert` rule.

### Important Factoids

[...]

PR #1971, opened 2024-12-20
Fix make alert rule uid settable to avoid accidental overwriting of existing rule history

moustafab:
Unfortunately this is the only viable workaround. It will require users to manually set the uid for alert rules similar to how it is done in file provisioning.

Addresses #1928


spinillos (approved):

JacobsonMT:
The change looks good to me, however I think there is still a bug that will cause this to be an incomplete fix.

Reproduction:
- Create a group, two rules will do
- Give the rules uids.
- Swap the order of any two rules definitions.
- Apply will show a diff (correct) and apply successfully (correct) but the diff will remain afterwards (bug).

Likely there is a bug in the API that is causing the group to not actually re-order itself. This shouldn't necessarily prevent the merge of this change, but we should probably add a test case to cover this since tests passed here and probably shouldn't have.

Other (likely related) cases also fail, such as:
- Adding a new rules to a group that immediately have UIDs (not adding them after the fact)

moustafab:
Hopefully https://github.com/grafana/grafana/pull/99858 should address the issues @JacobsonMT pointed out, but the question is then how should we merge and release this so that we can clarify that setting the `uid` in configuration relies on the version that has the fixes.

PR #1999, opened 2025-01-22
Cloud: Check if cloud configuration exists

spinillos:
It adds a check if the `GrafanaCloudAPI` configuration is set to early fail and avoid to panic if its nil.

Also it adds checks to ensure that the datasource implements the necessary configuration. Information [here](https://developer.hashicorp.com/terraform/tutorials/providers-plugin-framework/providers-plugin-framework-data-source-read#implement-data-source-client-functionality).


aangelisc (approved):
LGTM!

PR #2000, opened 2025-01-23
Reject recording rules with incompatible fields

rwwiv:
This PR introduces logic to validate that conflicting fields are not used for recording rules, since these values are stripped by Grafana and will cause plans to always show diffs for these fields.

Note: Due to limitations in the SDK this PR moves some validation out of the schema into `CreateContext` in order to allow for us to check for conflicting fields, which is not possible on a `ListType`

### DIFF
--- a/internal/resources/grafana/resource_alerting_rule_group.go
+++ b/internal/resources/grafana/resource_alerting_rule_group.go
@@ -795,7 +795,3 @@ func unpackRecord(p interface{}) *models.Record {
 	}
 	return res
 }
-
-func isValidRecordRule(rule *models.ProvisionedAlertRule) {
-
-}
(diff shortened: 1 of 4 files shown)


cindy:
Can you fill out the PR description? Otherwise, this looks good to me 👍

internal/resources/grafana/resource_alerting_rule_group.go
```diff
@@ -604,6 +609,30 @@ func unpackRuleData(raw interface{}) ([]*models.AlertQuery, error) {
...
+		}
+	}
+
+	// condition is required if the rule is not a recording rule
+	if rule.Condition == nil {
+		return fmt.Errorf(`"condition" is required`)
+	}
+	return nil
+}
+
```
moustafab: I think you can achieve this via [schema attributes](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#Schema) like `ConflictsWith` and `ExactlyOneOf` which will give you the `terraform validate` function and language server some teeth to help users

moustafab: Unfortunately sdkv2 limitations make this impossible on elements of a list type 😢

moustafab (approved):
LGTM, looks like ci docs check is failing because jq has a new website

PR #2001, opened 2025-01-24
feat: Add Frontend Observability provider

kpelelis:
This commit adds the groundwork to support Frontend Observability as code. More specificailly, the change contains

1. A Go HTTP client that consumes the Apps API (Create/Read/Update/Delete)
2. Data Source and Resource version of the Frontend Observability App entity.


Alex3k:
Hey @kpelelis this is awesome thank you! I have been after this for about a year :) 

I noticed the [thing about getting the frontend o11y API URL](https://github.com/grafana/terraform-provider-grafana/blob/a0a29c06087b52955e60a5a57f29ede3bdea0fad/docs/index.md#obtaining-frontend-observability-api-url). Would it be possible to update the grafana cloud provider to return this instead? Otherwise the frontend o11y terraform provider can’t be used when creating a [grafana cloud stack](https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/cloud_stack) within the same terraform, it can only be used in situations with an existing stack which isn’t ideal for the whole everything as code paradigm

kpelelis:
Hey @Alex3k thank you so much for your feedback. I updated the terraform API that it uses as minimal data as possible. Check the examples and how to set the feature up

docs/resources/frontend_o11y_app.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Example Usage
+
+```terraform
+data "grafana_cloud_stack" "teststack" {
+  provider    = grafana.cloud
+  name        = "gcloudstacktest"
+}
+
+resource "grafana_frontend_o11y_app" "test-app" {
```
Alex3k, line 21: Is it possible to set `Combine K6 with RUM data` in this resource also for each frontend o11y app?

docs/resources/frontend_o11y_app.md
```diff
@@ -0,0 +1,58 @@
...
+  provider        = grafana.cloud
+  stack_id        = data.grafana_cloud_stack.teststack.id
+  name            = "test-app"
+  allowed_origins = ["https://grafana.com"]
+
+  extra_log_attributes = {
+    "terraform" : "true"
+  }
+
+  settings = {
```
Alex3k, line 31: I couldn't see any docs on what the possible settings are for this block - is that something we plan to add? or maybe a link to some other docs that indicate what options are available?

Alex3k:
> Hey @Alex3k thank you so much for your feedback. I updated the terraform API that it uses as minimal data as possible. Check the examples and how to set the feature up

This is awesome! I like that you can use `cloud_access_policy_token` as an alternative to `frontend_o11y_api_access_token` but can still use two separate tokens if you like. Very helpful. 

I added a few comments around the settings block of the `grafana_frontend_o11y_app` resource.

I'd love to give this a run through and see how our current automation fits into this. When it's ready can you let me know so I can have a play?

docs/resources/frontend_o11y_app.md
kpelelis: I added documentation on the settings that we currently have available (More to be released soon ;) )

docs/resources/frontend_o11y_app.md
kpelelis: Yes! Its possible from the settings using the "combineLabData" setting

internal/common/frontendo11yapi/client.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+	return apps, nil
+}
+
+func (c *Client) GetApp(ctx context.Context, baseURL string, stackID int64, appID int64) (App, error) {
+	path := fmt.Sprintf("%s/app/%d", pathPrefix, appID)
+	var app App
+	err := c.doAPIRequest(ctx, baseURL, stackID, http.MethodGet, path, nil, &app)
+	if err != nil {
+		return App{}, fmt.Errorf("failed to get faro apps: %w", err)
```
spinillos:
```suggestion
		return App{}, fmt.Errorf("failed to get faro app: %w", err)
```

internal/common/frontendo11yapi/client.go
```diff
@@ -0,0 +1,233 @@
...
+	req.Header.Add("Content-Type", "application/json")
+	req.Header.Add("User-Agent", c.userAgent)
+
+	resp, err := c.client.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to do request: %w", err)
+	}
+
+	bodyContents, err := io.ReadAll(resp.Body)
+	resp.Body.Close()
```
spinillos: Shouldn't be a defer?

t00mas (approved):

internal/common/frontendo11yapi/client.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+	if !(resp.StatusCode >= 200 && resp.StatusCode <= 299) {
+		if resp.StatusCode == 404 {
+			return ErrNotFound
+		}
+		if resp.StatusCode == 401 {
+			return ErrUnauthorized
+		}
+		return fmt.Errorf("status: %d", resp.StatusCode)
+	}
```
spinillos: Its a bit complex to read, maybe this?

```suggestion
	if resp.StatusCode < http.StatusOK || resp.StatusCode >= http.StatusMultipleChoices {
	        switch resp.StatusCode {
		case http.StatusNotFound:
			return ErrNotFound
		case http.StatusUnauthorized:
			return ErrUnauthorized
		default:
			return fmt.Errorf("status: %d", resp.StatusCode)
		}
	}
```

kpelelis: I am personally in favour of using hardcoded values to indicate that an error is within the 2xx space as I don't remember by hand that Multiple Choices is 300. Other than that I will switch this to a switch, looks cleaner indeed

internal/resources/frontendo11y/resource_frontend_o11y_app.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+}
+
+func (r *resourceFrontendO11yApp) getStackCluster(ctx context.Context, stackID string) (string, error) {
+	stack, res, err := r.gcomClient.InstancesAPI.GetInstance(ctx, stackID).Execute()
+	if err != nil {
+		return "", err
+	}
+
+	if res.StatusCode >= 500 {
```
spinillos, line 121:
```suggestion
	if res.StatusCode >= http.StatusInternalServerError {
```

spinillos (approved):

PR #2006, opened 2025-01-28
[WIP] PoC for managing Grafana App Platform resources

radiohead:
### What

This PR is a PoC for managing Grafana App Platform resources via Terraform. It implements a generic GAP resource, generic client and implements two new resources – dashboards and playlists.

The dashboard resource also supports validation (using `grafana/foundation-sdk` and linting using `grafana/dashboard-linter`.

This is still WIP and parts of this code will be move to other repositories (e.g. generated dashboards code will be moved to `grafana/grafana`, resource clients to `grafana/grafana-app-sdk` and so on).

### Why

To add support for GAP resources and improve the existing Grafana dashboards resource specifically.

The code for specific resources should ideally be auto-generated by something (likely something similar or same as in https://github.com/grafana/terraform-provider-schemas/tree/main/gen).


go.mod
radiohead, line 1: I've added this as a temporary solution, to isolate all required changes in one place. However, this ended up requiring moving the `common.Client` from `internal` to `pkg`, because it wasn't otherwise possible to re-use it in this submodule.

If we are going to stick to the submodule approach, we'll need to think about a cleaner way to address the changes to the `common.Client`.

appplatform/pkg/terraform/resources/playlist.go
radiohead, line 1: This is 90% there, just needs validation for item types (since those are an enum).

appplatform/pkg/terraform/resource.go
radiohead, line 1: This is the main piece of work here, would be nice to get feedback on it.

appplatform/pkg/generated/resource/dashboard/v0alpha1/dashboard_spec_gen.go
radiohead, line 1: These will be moved and imported from `grafana/grafana` (same way as playlists are imported right now).

appplatform/pkg/client/namespaced.go
radiohead, line 1: @IfSentient I'm thinking about adding these 2 clients (resource & namespaced) to app SDK, thoughts?

radiohead:
@Duologic @K-Phoen @IfSentient this is still WIP but if you have some time I'd appreciate general feedback for this!

Especially about the submodule approach – I'd like to know if we should stick with it or not.

appplatform/pkg/client/namespaced.go
IfSentient: I think that makes sense, I think we might want to rename the `NamespacedClient`, as it's actually a Namespaced `ResourceClient`, and https://github.com/grafana/grafana-app-sdk/issues/194 will likely need to introduce a namespaced general client/registry.

radiohead: Great! I'm happy to rename / change things slightly – what's here is definitely not final.

Duologic:
On first scan:
* It looks like the refactor for the client could fit into a separate PR as it doesn't seem to change how it works.
* I don't have experience with the submodule approach, not sure what the trade-offs are.
* Can the dashboard schema be pulled in as a dependency from grafana/grafana instead of vendoring? From reading the description, I'd expect the SDK and linter need to know about the schema, not necessarily the provider directly, am I misunderstanding something?

radiohead:
Thank you for taking a look!

> On first scan:
> 
> * It looks like the refactor for the client could fit into a separate PR as it doesn't seem to change how it works.
> * I don't have experience with the submodule approach, not sure what the trade-offs are.

If you are OK with _not_ using a submodule it will make client refactoring unnecessary – the only reason I've done it was to make it importable from a submodule.

Let me know if dropping the submodule is fine and I'll update accordingly.

> * Can the dashboard schema be pulled in as a dependency from grafana/grafana instead of vendoring? From reading the description, I'd expect the SDK and linter need to know about the schema, not necessarily the provider directly, am I misunderstanding something?

Yes, I'm working on moving the dashboard schemas to `grafana/grafana` and importing them the same way playlist schemas are currently imported.

Duologic:
I don't really understand which problem the submodule is solving tbh. What are the trade offs of dropping the submodule?

radiohead:
> I don't really understand which problem the submodule is solving tbh. What are the trade offs of dropping the submodule?

The main reason I've introduced one was to isolate the changes and the dependencies. I am 100% happy to drop it.

radiohead:
Closing in favour of https://github.com/grafana/terraform-provider-grafana/pull/2075

PR #2008, opened 2025-01-29
Add client nil checks to avoid panics

spinillos:
Same as: https://github.com/grafana/terraform-provider-grafana/pull/1999 but for the missing clients.

Probably we could do something nicer since it follows the same pattern everywhere. But at the moment we should stop the bleeding.


sd2k (approved):
LGTM from ML

twu (approved):

PR #2014, opened 2025-01-31
[WIP]feat(bigTent slo): BigTent SLO tests

Leo-DiCara:
Adds tests to ensure the terraform provider for SLOs can built big tent SLOs.

### DIFF
--- a/examples/resources/grafana_slo/resource_big_tent.tf
+++ b/examples/resources/grafana_slo/resource_big_tent.tf
@@ -1,0 +1,119 @@
+resource "grafana_slo" "big_tent_test" {
+  name        = "Terraform Testing - Big Tent"
+  description = "Terraform Description - Big Tent"
+  query {
+    freeform {
+      query = jsonencode([
+{
+aggregation: "Sum",
+alias: "",
+application: "57831",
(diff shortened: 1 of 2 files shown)


examples/resources/grafana_slo/resource_big_tent.tf
```diff
@@ -0,0 +1,119 @@
...
+query: {
+params: []
+},
+reducer: {
+params: [],
+type: "avg"
+},
+type: "query"
+}
+],
```
ellisda, line 69:
```suggestion
conditions: [
{
evaluator: {
params: [
0,
0
],
type: "gt"
},
operator: {
type: "and"
},
query: {
params: []
},
reducer: {
params: [],
type: "avg"
},
type: "query"
}
],
```
we don't need these part for "math" expressions. We keep copy/pasting it... but we should stop

examples/resources/grafana_slo/resource_big_tent.tf
```diff
@@ -0,0 +1,119 @@
...
+datasource: {
+name: "Expression",
```
ellisda, line 71:
```suggestion
```

same as above, this isn't needed

PR #2016, opened 2025-01-31
feat(bigTent slo): Custom Validate function for big tent SLO schema

Leo-DiCara:
**Please Don't merge this until all work in https://github.com/grafana/slo/issues/2697 is complete. We want to ensure that the SLO Plugin API is merged and running in prod before we merge terraform-provider changes.**

This adds a custom validate function for queries that checks to ensure basic form of big SLO JSON is being followed before we allow the user to roundtrip it to the API. This checks for:

- Basic JSON structure
- refID field
- datasource field
- uid and type subfields in datasource

Closes https://github.com/grafana/slo/issues/2702, https://github.com/grafana/slo/issues/2876, https://github.com/grafana/slo/issues/2875


internal/resources/slo/resource_slo.go
```diff
@@ -104,14 +107,16 @@ Resource manages Grafana SLOs.
...
+										ValidateDiagFunc: ValidateBigTent(),
 									},
 									"total_metric": {
-										Type:        schema.TypeString,
-										Description: `Metric for total events (denominator)`,
-										Required:    true,
+										Type:             schema.TypeString,
+										Description:      `Metric for total events (denominator)`,
+										Required:         true,
+										ValidateDiagFunc: ValidateBigTent(),
```
elainevuong: clarification: all of our tests seem to be modifying the freeform `query` field. 
do we also still need to validate the `success_metric` and `total_metric` fields on the schema with the ValidateBigTent function? or do we expect that usage of Terraform and Big Tent will always be via the `query` field?

internal/resources/slo/resource_slo.go
```diff
@@ -661,3 +666,79 @@ func apiError(action string, err error) diag.Diagnostics {
 		},
 	}
 }
+
+func ValidateBigTent() schema.SchemaValidateDiagFunc {
+	return func(i interface{}, path cty.Path) diag.Diagnostics {
```
elainevuong, line 712: what's this `cty` package?

internal/resources/slo/resource_slo.go
```diff
@@ -661,3 +666,79 @@ func apiError(action string, err error) diag.Diagnostics {
...
+		}
+
+		var gmrQuery []map[string]any
+		err := json.Unmarshal([]byte(v), &gmrQuery)
+		if err != nil {
+			diags = append(diags, diag.Diagnostic{
+				Severity:      diag.Warning,
+				Summary:       "Bad JSON format",
+				Detail:        "If this is a big tent query, this should be valid JSON. If this is a prometheus query, ignore this.",
+				AttributePath: path,
```
elainevuong, line 743: disclaimer - I'm not sure if this possible, and I don't know too much about it.

Basically - we always perform this `ValidateBigTent` function, which validates the `query` field.
We first:
1 - check if it's a string
2 - check if it's valid JSON, if it isn't, we **return a warning, even if it's a valid Prometheus query.** Since it just returns a warning, it's non-blocking for the provider and it'll continue to process. However, this might be a bit off-putting for existing users. They've never gotten these warnings before, but now they get them but it's not really actionable? Might not be a great experience. 
3 - we then check for the presence of a `refId`, `datasource`, a `type`, and a `uid`. 

--
Is there a way we can get around with **not returning the warning for valid Prometheus queries**? I was looking around, and maybe we could incorporate a [DiffSuppressFunc](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#diffsuppressfunc)?

_SchemaDiffSuppressFunc is a function which can be used to determine whether a detected diff on a schema element is "valid" or not, and suppress it from the plan if necessary._

https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#SchemaDiffSuppressFunc

Pretty much - if we identify that it ISN'T JSON; we assume it's a Prom query, and we allow it go pass through validation.

There are some examples if these DiffSuppressFunc within the terraform provider already that you could reference.

What are your thoughts?

elainevuong:
I'd like to see if we can remove the warnings for valid Prometheus queries when we validate the `query` field. It might not be a fantastic customer experience for existing users, to suddenly get warnings from the Provider for queries that they haven't changed. 

Perhaps we can investigate if this `DiffSuppressFunc` is an option to pursue?

internal/resources/slo/resource_slo.go
ellisda: @elainevuong , we decide a few weeks ago to not support the ratio type, so we don't need to be checking for big-tent JSON in the `success_metric` or `total_metric` ([generated/models/slo/v0_0/slo_gen.go#L139](https://github.com/grafana/slo/blob/0688db1280ab82b3d22fd6786d06a7c8c240496e/pkg/generated/models/slo/v0_0/slo_gen.go#L139))

We should assume that all ratio SLOs are still promQL. Is there a place this assumption gets tricky? I know it's kind of weird to treat the freeform as "json or promQL" - but in the MTO code the weirdness didn't bubble up to the level of us modifying the schema to clarify

Leo-DiCara: I did add the Validate function to ALL metrics but only tested it on the query field, since functionally these three string fields should be exactly the same in terms of what we want to verify I didn't think there would be much value to rewriting the same tests checking that field.

internal/resources/slo/resource_slo.go
Leo-DiCara: This cty package is a [hasicorp provided package](https://pkg.go.dev/github.com/hashicorp/go-cty/cty) for managing config values. cty.Path handles managing paths. It is required to write tests for terraform going forward. 

Functionally here it just tells us at what depth our error occurred in, which is helpful since we are diving through JSON.

internal/resources/slo/resource_slo.go
Leo-DiCara: We could return with no warnings. My thoughts below:

If you are writing a Big tent query you need to provide valid json. We can't distinguish between is this bad JSON and you are writing a big tent query and is this just a prom query. 

Big tent users who see this message have queries that will always fail to validate, which could be confusing if you think you've written proper JSON.

I guess it boils down to: Who is it better to confuse? 
Big Tent users who will silently pass a tf plan to fail on tf apply?
-OR-
Existing Prom users who see a new warning and may get spooked?

IMO setting up new SLOs for Big Tent is more painful because of the potentially very complex JSON blobs, so I opted for a warning, but it could be that many more users will never even touch big tent SLOs and therefore more users will be annoyed by unactionable warnings when they edit/create SLOs.

My last thought is I'm not even sure validate runs if a tf plan doesnt generate a diff for that field. This would mean unchanged SLOs wouldn't generate warnings. tf validate would generate these warnings everytime.

ellisda: I think we should assume that big tent SLOs are the minority and we shouldn't raise warnings about totally valid promQL SLOs.

This does complicate our ability to raise useful tf-plan warnings for invalid json. One idea would be to use a simple `strings.Contains("datasource"...)` as a way to test whether a given SLO was "trying to include a source-datasource that is required for big-tent and not part of promql...

internal/resources/slo/resource_slo.go
elainevuong: If we're not supporting Ratio type for big-tent json, I think we ought to remove the `ValidateBigTent()` call from the `success_metric` and `total_metric` and only keep it on the `query`

elainevuong:
Just want to circle back - I verified that if we use the Terraform Provider as is, it DOES give warnings that show up, even if there isn't a diff in the field. 

> My last thought is I'm not even sure validate runs if a tf plan doesnt generate a diff for that field. This would mean unchanged SLOs wouldn't generate warnings. tf validate would generate these warnings everytime.

With the way this PR is set up currently, we've see this warning in both Ratio SLOs (due to the ValidateBigTent on the success and total metrics) AND in the Freeform Prom QL SLOs. These warnings appear on both "CREATE" of a new SLO, as well as on "UPDATE" of an existing SLO. 

I think we need two changes:
- drop the ValidateBigTent() on both the `success_metric` and `total_metric`
- remove the warning if we can't parse as JSON

screenshots if anyone is interested in viewing what the user experience is like. 

**CREATE - Ratio SLO**

Create - Ratio SLO
--

**CREATE - Adv Freeform Prom SLO**

Create - Adv Freeform Prom SLO
--

**UPDATE - Adv Freeform Prom SLO**
  
Update - Existing Adv Freeform SLO

internal/resources/slo/resource_slo.go
Leo-DiCara: Done.

internal/resources/slo/resource_slo.go
Leo-DiCara: I readded the error for not being JSON now that we have a new grafanaQueries type. We can be sure it should be json format

examples/resources/grafana_slo/resource_nonprometheus.tf
```diff
@@ -0,0 +1,96 @@
...
+          },
+          expression : "($total - $errors) / $total",
+          intervalMs : 1000,
+          maxDataPoints : 43200,
+          refId : "C",
+          type : "math"
+        }
+      ])
+    }
+    type = "grafana_queries"
```
elainevuong, line 60: I think this example type might be incorrect - should it be `grafanaQueries` or `grafana_queries`?

if I use `grafana_queries` I get an validation error


```
╷
│ Error: expected query.0.type to be one of ["freeform" "query" "ratio" "threshold" "grafanaQueries"], got grafana_queries
│ 
│   with grafana_slo.one,
│   on ratio.tf line 18, in resource "grafana_slo" "onelabel":
│   18:     type          = "grafana_queries"
```

templates/resources/slo.md.tmpl
```diff
@@ -20,6 +20,15 @@ description: |-
 
 {{ tffile "examples/resources/grafana_slo/resource_complex.tf" }}
 
+## Enterprise Datasources
+
+Currently supported datasources: AppDynamics, Splunk, Graphite
+
+Enterprise Datasource queries use the freeform query field. It expects a JSON string list of valid grafana query JSON objects.
```
elainevuong, line 27: should this say the `grafana_queries` field?

elainevuong:
@Leo-DiCara - there's an issue with the Terraform Provider. I believe it's in the `packQuery` portion, but we're also missing support for the `unpackQuery`. `packQuery` - this is where we translate what's in the HCL to what is sent to our API. Using the sample Terraform, I'm getting a 400 from the plugin, which is indicating that there isn't a query set on the SLO.

see my video attached

https://github.com/user-attachments/assets/e4692aca-c2ed-4f46-bdcb-872a52a6e8ae

internal/resources/slo/resource_slo.go
```diff
@@ -79,8 +82,8 @@ Resource manages Grafana SLOs.
 					Schema: map[string]*schema.Schema{
 						"type": {
 							Type:         schema.TypeString,
-							Description:  `Query type must be one of: "freeform", "query", "ratio", or "threshold"`,
-							ValidateFunc: validation.StringInSlice([]string{"freeform", "query", "ratio", "threshold"}, false),
+							Description:  `Query type must be one of: "freeform", "query", "ratio", "grafanaQueries" or "threshold"`,
+							ValidateFunc: validation.StringInSlice([]string{"freeform", "query", "ratio", "threshold", "grafanaQueries"}, false),
```
elainevuong: this should say `grafana_queries` not `grafanaQueries`

internal/resources/slo/resource_slo.go
```diff
@@ -511,6 +529,26 @@ func packQuery(query map[string]interface{}) (slo.SloV00Query, error) {
 		return sloQuery, nil
 	}
 
+	if query["type"] == "grafana_queries" {
+		freeformquery := query["grafana_queries"].([]interface{})[0].(map[string]interface{})
```
ellisda: issue: it appears that you're accessing the `[0]` position in the slice without first checking the length, which could panic.

suggestion: remove `freeform` from the variable name and just use `queryObject` for the outer object here.

question: is this unchecked type assertion for the `["grafana_queries"].(string)` consuming the terraform resource where the value is actually a string? I ask because the API type if structured and I was surprised so see the `.(string)`

Maybe I could could get a live walkthrough of this later this week if I don't find the time to play with the terraform-provider locally?

docs/resources/slo.md
```diff
@@ -125,6 +125,112 @@ resource "grafana_slo" "test" {
...
+Enterprise Datasource queries use the freeform query field. It expects a JSON string list of valid grafana query JSON objects.
+For additional help, view our [documentation](https://grafana.com/docs/grafana-cloud/alerting-and-irm/slo/).
+
+```terraform
+resource "grafana_slo" "test" {
+  name        = "Terraform Testing"
+  description = "Terraform Description"
+  query {
+    grafana_queries {
+      grafana_queries = jsonencode([
```
elainevuong, line 130: issue: can we use a Big Tent SLO example that matches the format and RefID of how the frontend generates it? i.e. using the `Success` and `Total` RefIDs, so it's easier for users to follow along? the example currently used looks to be a non-standard SLO that we had used for initial testing

maybe something like this

```
[
  {
    "datasource": {
      "type": "graphite",
      "uid": "datasource-uid"
    },
    "refId": "Success",
    "target": "groupByNode(perSecond(web.*.http.2xx_success.*.*), 3, 'avg'')"
  },
  {
    "datasource": {
      "type": "graphite",
      "uid": "datasource-uid"
    },
    "refId": "Total",
    "target": "groupByNode(perSecond(web.*.http.5xx_errors.*.*), 3, 'avg')"
  },
  {
    "datasource": {
      "type": "__expr__",
      "uid": "__expr__"
    },
    "expression": "$Success / $Total",
    "refId": "Expression",
    "type": "math"
  }
]
```

examples/resources/grafana_slo/resource_nonprometheus.tf
```diff
@@ -0,0 +1,96 @@
+resource "grafana_slo" "test" {
+  name        = "Terraform Testing"
+  description = "Terraform Description"
+  query {
+    grafana_queries {
+      grafana_queries = jsonencode([
```
elainevuong: issue: can we use a Big Tent SLO example that matches the format and RefID of how the frontend generates it? i.e. using the `Success` and `Total` RefIDs, so it's easier for users to follow along? the example currently used looks to be a non-standard SLO that we had used for initial testing

maybe something like this

```
[
  {
    "datasource": {
      "type": "graphite",
      "uid": "datasource-uid"
    },
    "refId": "Success",
    "target": "groupByNode(perSecond(web.*.http.2xx_success.*.*), 3, 'avg'')"
  },
  {
    "datasource": {
      "type": "graphite",
      "uid": "datasource-uid"
    },
    "refId": "Total",
    "target": "groupByNode(perSecond(web.*.http.5xx_errors.*.*), 3, 'avg')"
  },
  {
    "datasource": {
      "type": "__expr__",
      "uid": "__expr__"
    },
    "expression": "$Success / $Total",
    "refId": "Expression",
    "type": "math"
  }
]
```

internal/resources/slo/resource_slo.go
```diff
@@ -661,3 +699,80 @@ func apiError(action string, err error) diag.Diagnostics {
...
+					Detail:        fmt.Sprintf("expected Big Tent Query (refId:%v) to have a type", refID),
+					AttributePath: append(currentPath.Copy(), cty.IndexStep{Key: cty.StringVal("type")}),
+				})
+			}
+			_, ok = s["uid"]
+			if !ok {
+				diags = append(diags, diag.Diagnostic{
+					Severity:      diag.Error,
+					Summary:       "Missing Required Field",
+					Detail:        fmt.Sprintf("expected Big Tent Query (refId:%v) to have a uid", refID),
```
elainevuong: issue: do we want to use "Big Tent" language in our error messages? how about "supported datasource must specify a datasource: uid" and or "datasource: type" 

nit: did you want to specify `datasource: uid` rather than just `uid`? same comment for `datasource: type` above. since the `uid` and `type` fields are part of the `datasource`  struct. this is what it looks like in the Terraform terminal (I removed the `type` field and then I removed the `uid` field - wanted to see how the error message appeared to a user)

Screenshot 2025-02-26 at 12 08 33 PM

templates/resources/slo.md.tmpl
```diff
@@ -20,6 +20,15 @@ description: |-
...
+For additional help, view our [documentation](https://grafana.com/docs/grafana-cloud/alerting-and-irm/slo/).
+
```
elainevuong, line 30: > Enterprise Datasource queries use the freeform query field.

Issue: I believe this should say they use the `grafana_queries` field, not the freeform query field

elainevuong:
Sample Unhappy grafanaquery response from the API

notice how we have three apostrophes in the Graphite query

this returns a:

```
{
    "code": 400,
    "error": "SLO failed validation: invalid parameters on query please check your input values"
}
```

---

```
[
  {
    "datasource": {
      "type": "graphite",
      "uid": "becy9yvjmuz9ca"
    },
    "refId": "Success",
    "target": "groupByNode(perSecond(web.*.http.2xx_success.*.*), 1, 'avg''')"
  },
  {
    "datasource": {
      "type": "graphite",
      "uid": "becy9yvjmuz9ca"
    },
    "refId": "Total",
    "target": "groupByNode(perSecond(web.*.http.*.*.*), 1, 'avg')"
  },
  {
    "datasource": {
      "type": "__expr__",
      "uid": "__expr__"
    },
    "expression": "$Success / $Total",
    "refId": "Expression",
    "type": "math"
  }
]
```

docs/resources/slo.md
```diff
@@ -17,7 +17,58 @@ Resource manages Grafana SLOs.
 
 ## Example Usage
 
-### Basic
+### Ratio
+
+```terraform
```
ellisda, line 22: praise: we've been meaning to swap / fix this for awhile. Not part of big-tent, but a good improvement

docs/resources/slo.md
```diff
@@ -66,27 +117,55 @@ resource "grafana_slo" "test" {
...
-      group_by_labels = ["job", "instance"]
+    grafana_queries {
+      grafana_queries = jsonencode([
+        {
+          datasource : {
+            "type" : "graphite",
+            "uid" : "datasource-uid"
+          },
+          refId : "Success",
+          target : "groupByNode(perSecond(web.*.http.2xx_success.*.*), 3, 'avg'')"
```
ellisda: issue: the extra ' char was from testing, is invalid in current form

```suggestion
          target : "groupByNode(perSecond(web.*.http.2xx_success.*.*), 3, 'avg')"
```

docs/resources/slo.md
```diff
@@ -66,27 +117,55 @@ resource "grafana_slo" "test" {
...
+        },
+        {
+          datasource : {
+            "type" : "graphite",
+            "uid" : "datasource-uid"
+          },
+          refId : "Total",
+          target : "groupByNode(perSecond(web.*.http.5xx_errors.*.*), 3, 'avg')"
```
ellisda, line 145: suggestion - 5xx errors isn't really a great "total" query, we should give a better example ... but I think we may have used this one on website too, so need to fix both

docs/resources/slo.md
```diff
@@ -116,15 +191,13 @@ resource "grafana_slo" "test" {
...
-      label {
-        key   = "type"
-        value = "slo"
-      }
     }
   }
 }
 ```
 
+For a list of currently supported datasources review the [documentation](https://grafana.com/docs/grafana-cloud/alerting-and-irm/slo/set-up/additionaldatasources/#supported-data-sources).
```
ellisda: nit / suggestion: most of our docs seem to try to have more informative link titles than "docs" (though I say "docs" internally all the time)

```suggestion
For a complete list, see [supported data sources](https://grafana.com/docs/grafana-cloud/alerting-and-irm/slo/set-up/additionaldatasources/#supported-data-sources).
```

internal/resources/slo/resource_slo.go
```diff
@@ -92,7 +95,23 @@ Resource manages Grafana SLOs.
...
+										Description: "Freeform Query Field - valid promQl",
+									},
+								},
+							},
+						},
+						"grafana_queries": {
+							Type:        schema.TypeList,
+							MaxItems:    1,
+							Optional:    true,
+							Description: "Array for holding a set of grafana queries",
```
ellisda, line 107:
```suggestion
```
the other query types seem to skip this outer `description`

templates/resources/slo.md.tmpl
```diff
@@ -12,13 +12,22 @@ description: |-
...
+{{ tffile "examples/resources/grafana_slo/resource_ratio.tf" }}
 
 ### Advanced
 
-{{ tffile "examples/resources/grafana_slo/resource_complex.tf" }}
+{{ tffile "examples/resources/grafana_slo/resource.tf" }}
+
+### Grafana Queries - Any supported datasource
+
+Grafana Queries use the grafana_queries field. It expects a JSON string list of valid grafana query JSON objects.
```
ellisda:
```suggestion
Grafana Queries use the grafana_queries field. It expects a JSON string list of valid grafana query JSON objects, the same as you'll find assigned to a Grafana Dashboard panel `targets` field.
```

ellisda (approved):
LGTM - we'll have some minor doc additions to add here with instructions to prototype queries in a dashboard panel and copy/paste here, but we can add that later as well

docs/resources/slo.md
Leo-DiCara: Yeah this is the exact same query.

PR #2021, opened 2025-02-04
ML: Example job using a power transformation

csmarchbanks:
Add an example of using a power transformation on an ML job.


sd2k (approved):
LGTM. I guess we should wait until we've finished documenting this in our Cloud docs before merging though!

csmarchbanks:
Eh, if someone finds this before our other docs are up I don't think it is too big of a deal. I'll try to get those up today as well though.

PR #2027, opened 2025-02-11
fix: fix listing oncall users when more than 1 page

brojd:
Related to https://github.com/grafana/support-escalations/issues/14634 

When public API of OnCall is hit from the TF provider, we always pass page=1. In the datasource code of the provider there's a loop that depends on the presence of `next` in paginated response. And it seems we don't bump the page so the `next` is always there causing infinite chain of requests


vstpme (approved):

Duologic (approved):
Approved but know that @grafana/platform-monitoring does not do gatekeeping, teams should feel empowered to get things merged on their own.

PR #2044, opened 2025-02-25
feat(feo11y): Add `collector_endpoint` attribute on `frontend_o11y_app` data source

kpelelis:
This pull request adds a computed attribute for the `collector_endpoint` to simplify setups that require this URL. Also fix a wording in resource docs

### DIFF
--- a/internal/resources/frontendo11y/models.go
+++ b/internal/resources/frontendo11y/models.go
@@ -2,6 +2,7 @@ package frontendo11y
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/hashicorp/terraform-plugin-framework/attr"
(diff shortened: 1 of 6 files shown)


spinillos (approved):

internal/resources/frontendo11y/models.go
```diff
@@ -63,6 +65,7 @@ func (tfData FrontendO11yAppTFModel) toClientModel(ctx context.Context) (fronten
 		CORSAllowedOrigins: allowedOrigins,
 		ExtraLogLabels:     extraLogLabels,
 		Settings:           actualSettings,
+		CollectEndpointURL: tfData.CollectorEndpoint.ValueString(),
```
Alex3k, line 68: Should this be Collector rather than Collect?

Issue #1984, opened 2025-01-17
Support grafana_user for cloud stack


jaisharma639:
### Feature Request

The resource (and data source) [grafana_user](https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/user) is currently only supported for self hosted Grafana installation and doesn't work with cloud stack. The challenge being that calling the underlying api requires basic auth which isn't supported for cloud enterprise stack. With service account api token having Admin privileges, I get:
`You'll need additional permissions to perform this action. Permissions needed: users:read`

**Use Case:**
* Trying to provide RBAC at datasource level using [grafana_data_source_permission](https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/data_source_permission) requires referencing the `user.id` attribute. In our case, we need to lookup the user id in Grafana using users' email id
```
data "grafana_user" "user" {
  email    = "example@email.com"
}

data "grafana_data_source" "foo" {
  name     = "my-datasource"
}

resource "grafana_data_source_permission" "fooPermissions" {
  datasource_uid = grafana_data_source.foo.uid
  permissions {
    user_id    = grafana_user.user.id
    permission = "Edit"
  }
}
```
* Doing similar RBAC for dashboard folders requires looking up user id

**Solution/Feature request**
The api that `grafana_user` uses the api `/api/users/lookup?loginOrEmail=example@email.com` which restricts access to basic auth. 

Proposal is to create another terraform resource (and data source) which uses the api `/api/org/users?query=example@email.com`. The resource can be named `grafana_user_org` or `grafana_user_cloud`.

**Workaround:**
We have used terraform [http](https://registry.terraform.io/providers/hashicorp/http/latest/docs/data-sources/http) as a workaround that calls the api mentioned in proposal section. The api looks up user accepting email id and yields user id as response to be further consumed for providing RBAC to datasource.

Please let me know if this feature request is an acceptable addition to enhancing Grafana terraform. If yes, I'd like to contribute to the repository.

vtorosyan:
Thank you for contribution and creating the issue. While we would like to prevent provisioning of Cloud users through terraform, using already provisioned users totally makes sense.

We are looking into adding an enhancement which will allow to import/use users provisioned in Grafana Cloud so that they can be used in other resource provisioning, like shared in the example.

volcanonoodle:
Hi! Thank you for your patience. We are working on adding a data source to import users.

[...]

PR #2051, opened 2025-03-03
[IAM] Add data source to import org user using token auth

volcanonoodle:
Related to https://github.com/grafana/terraform-provider-grafana/issues/1984.

Adds a data source to import Grafana users by email, so that their user_id can be referenced in other resources, using an auth token instead of just basic auth. For this, we need to use the `/org/users` endpoints. Please see https://github.com/grafana/terraform-provider-grafana/issues/1984 for a detailed explanation about why this is needed. 

I've used the `/org/users/lookup?query=...` [endpoint](https://grafana.com/docs/grafana/latest/developers/http_api/org/#get-all-users-within-the-current-organization-lookup) because the `/org/users?query=...` [endpoint](https://grafana.com/docs/grafana/latest/developers/http_api/org/#get-all-users-within-the-current-organization)'s OpenAPI spec was outdated, which caused the OpenAPI client to not include the `query` param (see explanation in https://github.com/grafana/terraform-provider-grafana/issues/1984#issuecomment-2690248928). In the future, we might want to update this data source to use the `/org/users?query=...` endpoint, which returns much more data about the user (but needs a different set of permissions, so that might be something to pay attention to).


linoman (approved):

mgyongyosi (commented):
Could you please add this file to the CODEOWNERS? Thanks!

mgyongyosi (approved):

PR #2074, opened 2025-03-14
Fix role assignment when using Cloud provider for provisioning a service account

vtorosyan:
**Description**

When provisioning a Service Account with [cloud_stack_service_account](https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/cloud_stack_service_account
), the returned resource ID has the format `slug:<sa_id>`. At the moment the role assignment item resource is relying on `SplitOrgResourceID` function to resolve the service account ID, which causes an issue as it can't parse the SA ID from it.

**Fix**
The fix uses `SplitServiceAccountID` which is slug aware and is designated for exactly this case.

Fixes the following issue: https://github.com/grafana/identity-access-team/issues/1186

### DIFF
--- a/internal/resources/grafana/resource_role_assignment_item.go
+++ b/internal/resources/grafana/resource_role_assignment_item.go
@@ -176,7 +176,7 @@ func (r *resourceRoleAssignmentItem) Create(ctx context.Context, req resource.Cr
 		assignmentType = "user"
 		resourceID = data.UserID.ValueString()
 	case !data.ServiceAccountID.IsNull():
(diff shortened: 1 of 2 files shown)


vtorosyan:
cloud acceptance tests

vtorosyan:
Ok, I was trying out different ways to test this, as it's tricky to mimic the realistic scenario with a mix of cloud resource and an oss resource, but I think I got something which does the job of verifying that the fix works.

colin-stuart (approved):

mgyongyosi (approved):

PR #2078, opened 2025-03-14
Alerting: Handle dependency when deleting contact point in notification policy

### LINKED ISSUE #1806: [Bug]: Unable to remove contact point if used in notification policy
### Terraform Version

1.5.7

2.19.4

### Grafana Version

11.3.0-75420.patch2-75797 (cloud)

### Affected Resource(s)

* grafana_contact_point
* grafana_notification_policy

```hcl
locals {
  teams = {
    "team1" = {
      id            = "team1"
      alert_channel = "#team1-alerts"
    },
    "team2" = {
      id            = "team2"
      alert_channel = "#team2-alerts"
    },
  }
}

# Create contact points for each team.
resource "grafana_contact_point" "team" {
  for_each = {
    for team in local.teams :
    team.id => team
  }

  name = "team-${each.key}"

  slack {
    token     = var.grafana_slack_token
    recipient = trimprefix(each.value.alert_channel, "#")
  }
}

resource "grafana_notification_policy" "default" {
  group_by = ["grafana_folder", "alertname"]

  contact_point = grafana_contact_point.team1.name

  dynamic "policy" {
    for_each = {
      for team in local.teams :
      team.id => team
    }
    content {
      matcher {
        label = "team"
        match = "="
        value = policy.value.id
      }

      contact_point = grafana_contact_point.team[policy.value.id].name
    }
  }
}

```

### Expected Behavior

If I remove a team from the `local.teams` map, I should be able to apply a plan that successfully removes:

a) The contact point from the notification policy
b) The contact point entirely

Currently the dependency tree, and resource creation, works well for _creating_ the resources, but not so much for _destroying_ them.

### Actual Behavior

An error like this:

```
grafana_contact_point.team["team1"]: Destroying... [id=0:team1]

│ Error: [DELETE /v1/provisioning/contact-points/{UID}] DeleteContactpoints (status 409): {}
```

If I try to manually remove that Contact Point via the API first, then I get:

```
{"message":"Contact point is currently referenced by a notification policy.","messageId":"alerting.notifications.contact-points.referenced","statusCode":409,"traceID":"fdd47c6ffcd5bc66f47a2453367e96c9"}
```

Which is understandable...

Because we've got the `provisioned = true` attribute set on the notification policy, we can't manually delete the reference to the contact point.

### Steps to Reproduce

1. `terraform apply` the config as supplied
2. Remove `team2` from the map.
3. `terraform plan`
   - get a legitimate-looking plan (delete 1 contact point, update notification policy).
4. `terraform apply`    - fails to delete contact point
   - doesn't attempt to update notification policy.

### Important Factoids

It's perhaps debatable whether this is a Terraform issue or a provider issue? I suppose if the provider called the DeleteContactPoint API, with a flag that represented "cascade to all references", then this could be fixed.

I could break the dependency on notification_policy -> contact_point, by not using Terraform implict dependencies, but I expect that if I did that then I'd get an error on creation, rather than destruction, due to trying to add a contact point which doesn't yet exist into a notification policy?

### References

_No response_

titolins:
## Context
- Fixes https://github.com/grafana/terraform-provider-grafana/issues/1806
- I'm not sure this is something we would like to handle at all, given that using the lifecycle argument in terraform should solve it the same
- Leaving it as a draft for now as it hasn't been thoroughly tested

## Proposed solution
- Check if the contact point is referenced in any notification policy
- If found, create a sanitized policy tree that removes references to the contact point
- Update the notification policy before deleting the contact point

### DIFF
--- a/internal/resources/grafana/resource_alerting_contact_point.go
+++ b/internal/resources/grafana/resource_alerting_contact_point.go
@@ -233,6 +233,23 @@ func updateContactPoint(ctx context.Context, data *schema.ResourceData, meta int
 func deleteContactPoint(ctx context.Context, data *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client, _, name := OAPIClientFromExistingOrgResource(meta, data.Id())
 
+	// Check if the contact point is being used by a notification policy
+	policyTreeResp, err := client.Provisioning.GetPolicyTree()
+	if err == nil && policyTreeResp.Payload != nil {
+		// Check if the contact point is referenced in the policy tree
+		if isContactPointInUse(policyTreeResp.Payload, name) {
+			// Create a temporary policy tree without references to this contact point
+			updatedPolicyTree := sanitizePolicyTree(policyTreeResp.Payload, name)
+
+			// Update the policy tree to remove references to this contact point
+			putParams := provisioning.NewPutPolicyTreeParams().WithBody(updatedPolicyTree)
+			_, err := client.Provisioning.PutPolicyTree(putParams)
+			if err != nil {
+				return diag.Errorf("failed to update notification policy to remove references to contact point '%s': %v", name, err)
+			}
+		}
+	}
+
 	resp, err := client.Provisioning.GetContactpoints(provisioning.NewGetContactpointsParams().WithName(&name))
 	if err, shouldReturn := common.CheckReadError("contact point", data, err); shouldReturn {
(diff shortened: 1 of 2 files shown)


titolins:
Closing this as not accepted. Terraform is designed to handle dependencies and this case specifically can be worked around with the `lifecycle` option, as mentioned in the issue

PR #2085, opened 2025-03-18
LBAC: Set rules as map in resource

spinillos:
It set rules directly as a map instead of use `jsonencode`


spinillos:
It wasn't the main issue of the broken test 😞

Issue #1507, opened 2024-04-23
[Feature Request]: Separate `grafana_rule_group` and `grafana_alert_rule` resources


bnjns:
### Feature Request

At the moment, you create multiple alert rules using `rule` blocks in the `grafana_rule_group` resource. This works well for simple configurations where you can configure all the alerts in a single place, but it breaks down for complex configurations.

For example, we use various Terraform modules to build up our dashboard JSON - each of these needs to conditionally create any relevant alerts within its own module, which means that these alerts cannot share the same evaluation/rule group, even though that's how they'd appear if you created them manually. At the moment, we work around this issue by making each alert have its own evaluation group, but that's not very pleasant in the UI and makes navigating the alerts frustrating.

Separating the `rule` blocks into their own `grafana_alert_rule` resource would make this configuration much easier to achieve, and I'd argue that this is much more aligned to how Terraform expects to manage resources (the UI shows the evaluation group and alert rules as separate entities, which should be managed separately by Terraform).

I appreciate that given how the alerting API appears to be designed this might be a challenge, but I feel this would be a better design/UX for the provider.

jsmrcaga:
Hello, just to add to the doc what I believe to be a very common pattern:

A small-to-medium sized tech company uses Grafana for Dashboarding and Alerting. They manage their infrastructure using Terraform, and each team has their own projects/systems.
Every team wants to manage their architecture on a by-project basis, every project having its own infrastructure needs, but also its own alerting needs.

In this case, teams that don't even know each other want to be able to add custom "Project" Rules to their "Team" Rule Group, maybe even to cross-team Rule Groups.

WaqasAliAbbasi:
Hi @julienduchesne, I attempted a first draft in the referenced commits. Happy to add tests/polish it if this is something the grafana provider team is interested in 😊 (would also love comments if you prefer a different implementation).

Our team just started using terraform for Grafana alerts and the files get too long (we have around ~15 rules per alert group) making lookups/edits difficult. We also can't modularize the rules as that would create a different alert group for each rule instance. Agree with @bnjns that a separate alert rule resource would improve the developer experience. Eager to hear the team's thoughts on this 💪

eschulma:
I am struggling with this issue right now and have spent several hours trying to figure out a good way to modularize this. 90% of our alert rules have similar formatting, it's silly to have such ridiculously long files.

paulojmdias:
[...]

PR #2099, opened 2025-03-25
Add resource grafana rule

dongho-jung:
- [unnecessary diff with grafana_rule_group](https://github.com/grafana/terraform-provider-grafana/issues/2098)
- [[Feature Request]: Separate grafana_rule_group and grafana_alert_rule resources](https://github.com/grafana/terraform-provider-grafana/issues/1507)

To resolve the issues mentioned earlier, I made some changes to the code.

I didn’t encounter any errors when running the tests locally, but I haven’t created tests for the newly added grafana_rule resource yet.

I also need to check whether it’s possible to create an alert rule group without any alert rules and implement support for that. (Currently, [an alert rule is required in order to create an alert rule group.](https://github.com/grafana/terraform-provider-grafana/issues/2098#issuecomment-2751163613))

PR #2100, opened 2025-03-25
Add a grafana alert rule resource

WaqasAliAbbasi:
As per https://github.com/grafana/terraform-provider-grafana/issues/1507, this MR adds a separate `grafana_alert_rule` resource.

For now, it's a WIP MR to allow maintainers to give feedback on the approach.

TODO:
- [ ] investigate issues that might arise using `grafana_rule_group` and `grafana_alert_rule` together for the same rule group name
- [x] investigate what happens when a `grafana_alert_rule` is created with a rule group name that does not exist -> `new rule group is created`
- [x] investigate what happens when the last `grafana_alert_rule` of a rule group is deleted (will the rule group be deleted automatically?) -> `rule group is deleted`


WaqasAliAbbasi:
replaces https://github.com/grafana/terraform-provider-grafana/pull/2099

PR #2111, opened 2025-04-01
Fleet Management collector data sources

johannaojeling:
Adds two new Fleet Management collector data sources:
- `grafana_fleet_management_collector` for getting a single collector by ID
- `grafana_fleet_management_collectors` for listing all collectors in a stack


internal/resources/fleetmanagement/data_source_collectors.go
johannaojeling, line 1: I have skipped adding acceptance tests for the data source `grafana_fleet_management_collectors` due to a known issue in the acceptance testing framework: [No id found in attributes](https://developer.hashicorp.com/terraform/plugin/framework/acctests#no-id-found-in-attributes). To work around the issue, the documentation suggests to add a root level ID attribute to the data source schema. However this data source represents a list of collectors and does not logically have an ID (the closest would be the stack ID). I want to avoid cluttering the schema with an ID field unnecessarily, so I have tested it manually instead.

internal/resources/fleetmanagement/model_collector.go
```diff
@@ -9,26 +9,56 @@ import (
...
-	return &collectorModel{
+	return &collectorResourceModel{
 		ID:               types.StringValue(msg.Id),
 		RemoteAttributes: remoteAttributes,
 		Enabled:          types.BoolPointerValue(msg.Enabled),
 	}, nil
 }
 
-func collectorModelToMessage(ctx context.Context, model *collectorModel) (*collectorv1.Collector, diag.Diagnostics) {
+func collectorResourceModelToMessage(ctx context.Context, model *collectorResourceModel) (*collectorv1.Collector, diag.Diagnostics) {
```
tpaschalis, line 61: Should this also handle LocalAttributes?

internal/resources/fleetmanagement/model_collector.go
```diff
@@ -9,26 +9,56 @@ import (
...
-func collectorMessageToModel(ctx context.Context, msg *collectorv1.Collector) (*collectorModel, diag.Diagnostics) {
+type collectorDataSourcesModel struct {
+	Collectors []collectorDataSourceModel `tfsdk:"collectors"`
+}
+
+type collectorResourceModel struct {
+	ID               types.String `tfsdk:"id"`
+	RemoteAttributes types.Map    `tfsdk:"remote_attributes"`
+	Enabled          types.Bool   `tfsdk:"enabled"`
+}
```
tpaschalis, line 27: Why does collectorResourceModel not include LocalAttributes while collectorDataSourceModel does?

Is this like the 'input' to CreateCollector calls while the other one is the output of ListCollectors?

tpaschalis (approved):
Did not test this manually but LGTM while reading through % the question below.

internal/resources/fleetmanagement/model_collector.go
johannaojeling: Yeah the resource is primarily used to create, update and delete a collector. When a collector is created (pre-registered) via `CreateCollector` it did not have any local attributes (up until recently - now it will have the `collector.ID`), so the field returned was empty. The next time the TF resource is read/refreshed, if `RegisterCollector` has been called by the Alloy instance in between and configured local attributes, there would be changes in the state since the local attributes would have changed value.

If there is a demand for making the local attributes available also for the collector resource, we could consider adding it.

PR #2113, opened 2025-04-03
Add proper error handling for generic app platform resource

radiohead:
### What

This change adds proper handling of API errors into Terraform diagnostics for all app platform resources which use the generic `Resource` as their implementation.

### Why

This makes sure that the API errors are clearly communicated to the users, as opposed to just vague "request failed on the server" messages.


radiohead:
Depends on https://github.com/grafana/grafana-app-sdk/pull/730

spinillos (approved):

PR #1727, opened 2024-07-29
Cloud: Fix stack wake-up function

julienduchesne:
Closes https://github.com/grafana/terraform-provider-grafana/issues/1725

In https://github.com/grafana/terraform-provider-grafana/pull/1545, the stack readiness check was improved by calling the `/api/health` endpoint instead of just the stack URL. While this does provide a better indication that the instance is up and running, the `/api/health` subpath does not trigger Grafana Cloud's stack wake-up 

With this PR, we'll call both the stack's base URL and its health endpoint

PR #2114, opened 2025-04-03
Use instance login page for waking up cloud stack

patrickeasters:
The Terraform provider currently tries to confirm a Grafana stack is ready hitting the base URL of the instance since #1727. For stacks with autologin enabled, this wake-up check will ultimately follow an HTTP redirect to the configured identity provider.

In cases where the IdP is unreachable or the SSO is misconfigured (invalid client ID, URL, etc), this can lead to a situation where Terraform can no longer manage the stack resources until the SSO misconfiguration is manually resolved.

This updates the wake-up check to use the stack login URL with the `noAutoLogin` query parameter set to prevent the client from being redirected away from the stack.

### DIFF
--- a/internal/resources/cloud/resource_cloud_stack.go
+++ b/internal/resources/cloud/resource_cloud_stack.go
@@ -574,10 +574,16 @@ func waitForStackReadiness(ctx context.Context, timeout time.Duration, stackURL 
 	if joinErr != nil {
 		return diag.FromErr(joinErr)
 	}
+	wakePath, joinErr := url.JoinPath(stackURL, "login")
+	if joinErr != nil {
+		return diag.FromErr(joinErr)
+	}
+	wakeURL := wakePath + "?disableAutoLogin=true"
+
 	err := retry.RetryContext(ctx, timeout, func() *retry.RetryError {
-		// Query the instance URL directly. This makes the stack wake-up if it has been paused.
+		// Query the instance login page directly. This makes the stack wake-up if it has been paused.
 		// The health endpoint is helpful to check that the stack is ready, but it doesn't wake up the stack.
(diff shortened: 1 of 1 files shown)


internal/resources/cloud/resource_cloud_stack.go
```diff
@@ -574,10 +574,16 @@ func waitForStackReadiness(ctx context.Context, timeout time.Duration, stackURL
 	if joinErr != nil {
 		return diag.FromErr(joinErr)
 	}
+	wakePath, joinErr := url.JoinPath(stackURL, "login")
+	if joinErr != nil {
+		return diag.FromErr(joinErr)
+	}
+	wakeURL := wakePath + "?noAutoLogin=true"
```
nachogiljaldo: I never used this setting, so please correct me if I'm wrong, but looking at the docs: https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/#avoid-automatic-login seems it should be `disableAutoLogin` ?

nachogiljaldo (commented):
Probably just my ignorance, but better safe than sorry :)

internal/resources/cloud/resource_cloud_stack.go
patrickeasters: Ah thanks. Typo on my part.


Confirmed via a quick curl:
```
# Confirm HTTP 200 with disableAutoLogin
❯ https --headers --follow "amer.cloud.demokit.grafana.com/login?disableAutoLogin=true"
HTTP/1.1 200 OK
...

# Confirm HTTP 302 redirect from root URL using old behavior
❯ https --headers amer.cloud.demokit.grafana.com/
HTTP/1.1 302 Found
Location: /login
...
```

nachogiljaldo (approved):

PR #2120, opened 2025-04-09
Feat: add new alerting rule fields

moustafab:
New fields for `keep_firing_for` and `missing_series_evals_to_resolve` have been added to the alerting rule group resource.

Added in https://github.com/grafana/grafana/pull/102150


moustafab:
This depends on fixes for grafana https://github.com/grafana/grafana/pull/103736 and maybe https://github.com/grafana/grafana/pull/103735

docs/resources/rule_group.md
```diff
@@ -142,7 +142,9 @@ Optional:
 - `exec_err_state` (String) Describes what state to enter when the rule's query is invalid and the rule cannot be executed. Options are OK, Error, KeepLast, and Alerting.  Defaults to Alerting if not set.
 - `for` (String) The amount of time for which the rule must be breached for the rule to be considered to be Firing. Before this time has elapsed, the rule is only considered to be Pending. Defaults to `0`.
 - `is_paused` (Boolean) Sets whether the alert should be paused or not. Defaults to `false`.
+- `keep_firing_for` (String) The amount of time for which the rule will considered to be Recovering after initially Firing. Before this time has elapsed, the rule will continue to fire once it's been triggered.
```
alexander-akhmetov, line 145: > Before this time has elapsed, the rule will continue to fire once it's been triggered

It'll be in `Recovering` state rather than firing:

```
Firing -(query returns normal results)> Recovering --(after keep_firing_for)--> Normal
```

Maybe something like `Before this time has elapsed, the rule is only considered to be Recovering.`?

moustafab: That's the internal state in the ruler though? But effectively it's treated externally as if it's still firing and no resolution alert would go out right?

alexander-akhmetov: No resolution, but in the UI it's shown as `Recovering`

alexander-akhmetov (dismissed):
LGTM

moustafab:
This became a bit more involved, so converting back to draft while I get the solution for the mute timings.

internal/resources/grafana/resource_alerting_rule_group.go
```diff
@@ -519,6 +542,15 @@ func packAlertRule(r *models.ProvisionedAlertRule) (interface{}, error) {
 		json["record"] = record
 	}
 
+	// FIXME: open api needs to be a reference to the duration
```
moustafab: @alexander-akhmetov I forgot to about this discovery. The `For` in the spec is a *Duration and the KeepFiringFor is a Duration. I think effectively they're the same thing but not 100% sure, can you have another look at this?

alexander-akhmetov: I'm not sure why it's not a `*Duration` as well, they are [similarly defined](https://github.com/grafana/grafana/blob/5a10e3b43e1936fa20552bb2772cac4ceb83ac6f/pkg/services/ngalert/api/tooling/definitions/provisioning_alert_rules.go#L158-L163). I'll check.

alexander-akhmetov: Probably because the For field is required, it's a pointer: https://github.com/go-swagger/go-swagger/issues/1386#issuecomment-360341120, and KeepFiringFor is not required.

PR #2125, opened 2025-04-11
feat: Oncall integration labels

willgallego-grafana:
Creating a label datasource, we can pass a label key/value pair to the `labels` attribute for a `grafana_oncall_integration` to enable labels through terraform.

Sample terraform:

```terraform
terraform {
  required_providers {
    grafana = { 
      source  = "grafana/grafana"
      version = ">= 1.22.0"
    }   
  }
}

provider "grafana" {
  alias      = "oncall"
  url        = "http://grafana:3000"
  oncall_url = "http://oncall-engine:8080"
  auth       = "<token here>"
}

// Step 2: Interact with Grafana OnCall
resource "grafana_oncall_integration" "test-acc-integration" {
  provider = grafana.oncall
  name     = "my integration"
  type     = "webhook"
  default_route {}
  labels = [data.grafana_oncall_label.test-label]
}

data "grafana_oncall_label" "test-label" {
  provider = grafana.oncall
  key      = "KeyInTerraform"
  value    = "ValueInTerraform"
}

output "example" {
  value = data.grafana_oncall_label.test-label
}

output "key_value_list" {
  value = [data.grafana_oncall_label.test-label]
}

output "integration" {
  value = resource.grafana_oncall_integration.test-acc-integration
}
```


internal/resources/oncall/data_source_label.go
```diff
@@ -0,0 +1,66 @@
...
+		},
+	}
+}
+
+func (d *labelDataSource) Read(ctx context.Context, req datasource.ReadRequest, resp *datasource.ReadResponse) {
+	// Read Terraform state data into the model
+	var data labelDataSourceModel
+	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
+
+	data.ID = basetypes.NewStringValue(data.Key.ValueString())
```
willgallego-grafana, line 57: TODO - I probably need a better way of setting the id here. Our modeling in the app has the uniqueness derived from the key/value pairing, not just the key itself (though I don't know if we care about that in the terraform code)

internal/resources/oncall/resource_integration.go
```diff
@@ -301,13 +318,15 @@ func resourceIntegrationCreate(ctx context.Context, d *schema.ResourceData, clie
...
 	defaultRouteData := d.Get("default_route").([]interface{})
+	labelsData := d.Get("labels").([]interface{})
 
 	createOptions := &onCallAPI.CreateIntegrationOptions{
 		TeamId:       teamIDData,
 		Name:         nameData,
 		Type:         typeData,
 		Templates:    expandTemplates(templatesData),
 		DefaultRoute: expandDefaultRoute(defaultRouteData),
+		Labels:       expandLabels(labelsData),
```
willgallego-grafana, line 323: Until the `amixr-api-go-client` changes are pushed, these will continue to fail.

internal/resources/oncall/resource_integration.go
```diff
@@ -792,3 +814,36 @@ func expandDefaultRoute(input []interface{}) *onCallAPI.DefaultRoute {
 	}
 	return &defaultRoute
 }
+
+func expandLabels(input []interface{}) []*onCallAPI.Label {
+	labelsData := make([]*onCallAPI.Label, 0, 1)
+
+	for _, r := range input {
+		inputMap := r.(map[string]interface{})
+		key, keyExists := inputMap["key"]
```
willgallego-grafana, line 817: This is a work around, of sorts.

We're using TypeMap here and the preferred way for users to interact with labels as part of oncall integrations is to create a data source object which enforces requiring `key` and `value` as attributes, then passing that object into the list.

A user could, instead, create the object directly within the list such as:
```
...
labels = [
    {
      key = "SomeKey"
      value = "SomeValue"
    }
]
...
```

which would work. That said, `schema.TypeMap` can't enforce required values (see: https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-types#typemap) and if the user leaves one out, it would normally panic.

This is at least a check so it doesn't totally fall over, but it isn't great UX as it doesn't indicate to the user they've left anything out. We could rework the returns on the `flattenLabels` and other checks such that we can return an error in the `diag` object.

mderynck (approved):

PR #2128, opened 2025-04-11
feat: add support for per check alerts

VikaCep:
Closes https://github.com/grafana/synthetic-monitoring/issues/228

Adds support for Synthetic Monitoring - Per Check Alerts ([docs](https://grafana.com/docs/grafana-cloud/testing/synthetic-monitoring/configure-alerts/configure-per-check-alerts/))

Requires https://github.com/grafana/synthetic-monitoring-api-go-client/pull/272

Example:

- defining the resources

```
resource "grafana_synthetic_monitoring_check" "test" {
  job     = "Test Check with Alerts"
  target  = "https://grafana.com"
  enabled = true
  probes  = [1]
  labels  = {}
  settings {
    http {
      ip_version = "V4"
      method     = "GET"
    }
  }
}

resource "grafana_synthetic_monitoring_check_alerts" "test" {
  check_id = grafana_synthetic_monitoring_check.test.id
  alerts = [{
    name      = "ProbeFailedExecutionsTooHigh"
    threshold = 1
    period    = "15m"
  },
  {
    name      = "TLSTargetCertificateCloseToExpiring"
    threshold = 14
    period    = ""
  }]
} 
```

- running `terraform apply`

image
image

- seeing the generated resources in the app 

image
image


docs/resources/synthetic_monitoring_check_alerts.md
```diff
@@ -0,0 +1,65 @@
...
+  Manages alerts for a check in Grafana Synthetic Monitoring.
+  These alerts will trigger when your check fails from the specified probes.
+---
+
+# grafana_synthetic_monitoring_check_alerts
+
+Manages alerts for a check in Grafana Synthetic Monitoring.
+
+* [Official documentation](https://grafana.com/docs/grafana-cloud/testing/synthetic-monitoring/configure-alerts/configure-per-check-alerts/)
+* [API documentation](https://github.com/grafana/synthetic-monitoring-api-go-client/blob/main/docs/API.md#alerts)
```
ka3de: I made a [comment](https://github.com/grafana/synthetic-monitoring-api-go-client/pull/272#pullrequestreview-2775499104) on this in the PR for the API client that this docs currently don't exist.

docs/resources/synthetic_monitoring_check_alerts.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `check_id` - (Required) The ID of the check to manage alerts for.
+* `alerts` - (Required) List of alerts for the check. Each alert has the following fields:
+  * `name` - (Required) Name of the alert. Must be one of: `ProbeFailedExecutionsTooHigh`, `TLSTargetCertificateCloseToExpiring`.
+  * `threshold` - (Required) Threshold value for the alert.
+  * `period` - (Required) Period for the alert threshold. Required and must be one of: `1m`, `2m`, `5m`, `10m`, `15m`, `20m`, `30m`, `1h` for `ProbeFailedExecutionsTooHigh` alerts. For other alerts, set to `""` (empty string).
```
ka3de: I think the _period_ is more related with the alert type, more than "with the threshold". Also, the min period is by now `5m`. So should this be:
```suggestion
  * `period` - (Required) Period for the alert. Required and must be one of: `5m`, `10m`, `15m`, `20m`, `30m`, `1h` for `ProbeFailedExecutionsTooHigh` alerts. For other alerts, set to `""` (empty string).
```
?

I'm also wondering if we should mention specific alert names or considerations for these, as that means that we'll have to update these docs everytime we add a new alert, or we can just rely on the link to the Grafana Cloud docs and expect to provide the full documentation there?

internal/resources/syntheticmonitoring/resource_check_alert.go
```diff
@@ -0,0 +1,226 @@
...
+func resourceCheckAlerts() *common.Resource {
+	return common.NewLegacySDKResource(
+		common.CategorySyntheticMonitoring,
+		"grafana_synthetic_monitoring_check_alerts",
+		resourceCheckAlertID,
+		&schema.Resource{
+			Description: `
+Manages alerts for a check in Grafana Synthetic Monitoring.
+
+* [Official documentation](https://grafana.com/docs/grafana-cloud/synthetic-monitoring/configure-alerts/)
```
ka3de: I think this link is incorrect:
```suggestion
* [Official documentation](https://grafana.com/docs/grafana-cloud/testing/synthetic-monitoring/configure-alerts/configure-per-check-alerts/)
```

internal/resources/syntheticmonitoring/resource_check_alert.go
```diff
@@ -0,0 +1,226 @@
...
+	)
+}
+
+func resourceCheckAlertCreate(ctx context.Context, d *schema.ResourceData, c *smapi.Client) diag.Diagnostics {
+	checkID := int64(d.Get("check_id").(int))
+
+	// Validate that the check exists
+	if err := validateCheckExists(ctx, c, checkID); err != nil {
+		return diag.FromErr(err)
+	}
```
ka3de: Do we have to do this from the client considering that the backend will already perform this validation as part of the "update alerts" request?

internal/resources/syntheticmonitoring/resource_check_alert.go
```diff
@@ -0,0 +1,226 @@
...
+	alerts := make([]model.CheckAlert, len(alertsList))
+
+	for i, alertMap := range alertsList {
+		alertData := alertMap.(map[string]interface{})
+		name := alertData["name"].(string)
+		period, hasPeriod := alertData["period"].(string)
+
+		if name == AlertNameProbeFailedExecutionsTooHigh && !hasPeriod {
+			return nil, fmt.Errorf("period is required when name is %s", AlertNameProbeFailedExecutionsTooHigh)
+		}
```
ka3de: This is OK, but this validation is already done in the backend, so I wonder if we should just pass the fields set by the user and delegate the validation. Otherwise with this approach we'll have to modify the Terraform provider every time we add new alert types. What's your take on this?

internal/resources/syntheticmonitoring/resource_check_alert.go
```diff
@@ -0,0 +1,226 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	_, err = c.UpdateCheckAlerts(ctx, checkID, alerts)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId(fmt.Sprintf("%d", checkID))
```
ka3de:
```suggestion
	d.SetId(strconv.FormatInt(checkID, 10))
```

Beyond that suggestion, I'm a little bit confused on this, probably because I don't know much about the Terraform provider implementation or how it actually works. In this function we receive a `ResourceData` that I understand matches the definition written by the user in the `.tf` file. So we expect that definition to contain a `check_id` as defined in the docs. Is this correct?

But then here we set the ID to the same check_id. I guess we have to do that due to how Terraform works? Reading the doc for the `SetId` method:
```
// SetId sets the ID of the resource. If the value is blank, then the
// resource is destroyed.
```

internal/resources/syntheticmonitoring/resource_check_alert.go
```diff
@@ -0,0 +1,226 @@
...
+func resourceCheckAlertRead(ctx context.Context, d *schema.ResourceData, c *smapi.Client) diag.Diagnostics {
+	checkID, err := strconv.ParseInt(d.Id(), 10, 64)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	// Validate that the check exists
+	if err := validateCheckExists(ctx, c, checkID); err != nil {
+		return diag.FromErr(err)
+	}
```
ka3de: Same comment as above. Since this is already done by the API, should rely on that and remove this from the client?

internal/resources/syntheticmonitoring/resource_check_alert.go
```diff
@@ -0,0 +1,226 @@
...
+func resourceCheckAlertUpdate(ctx context.Context, d *schema.ResourceData, c *smapi.Client) diag.Diagnostics {
+	checkID, err := strconv.ParseInt(d.Id(), 10, 64)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	// Validate that the check exists
+	if err := validateCheckExists(ctx, c, checkID); err != nil {
+		return diag.FromErr(err)
+	}
```
ka3de: Same comment as above regarding duplicated validation in client vs API backend.

internal/resources/syntheticmonitoring/resource_check_alert.go
```diff
@@ -0,0 +1,226 @@
...
+func resourceCheckAlertDelete(ctx context.Context, d *schema.ResourceData, c *smapi.Client) diag.Diagnostics {
+	checkID, err := strconv.ParseInt(d.Id(), 10, 64)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	// Validate that the check exists
+	if err := validateCheckExists(ctx, c, checkID); err != nil {
+		return diag.FromErr(err)
+	}
```
ka3de: Same comment as above regarding duplicated validation in client vs API backend.

internal/resources/syntheticmonitoring/resource_check_alert.go
```diff
@@ -0,0 +1,226 @@
...
+		alerts[i] = alert
+	}
+
+	return alerts, nil
+}
+
+func validateCheckExists(ctx context.Context, c *smapi.Client, checkID int64) error {
+	_, err := c.GetCheck(ctx, checkID)
+	return err
+}
```
ka3de: Related with the other comments, I'm not sure we need this explicit method in the client for alerts.

ka3de (commented):
I'm not an expert on Terraform or providers implementation, so someone else can probably provide better feedback on that.

Reviewing this mostly from the Synthetic Monitoring point of view I think overall looks OK, but I made a few comments, particularly on validation and on whether we should make this more generic and avoid referencing particular alert types so we don't have to make changes to the provider implementation or the docs whenever we add new alerts. Let me know what do you think.

docs/resources/synthetic_monitoring_check_alerts.md
VikaCep: Good call. Removed alert names and adjusted the min period.

internal/resources/syntheticmonitoring/resource_check_alert.go
VikaCep: You make a good point. Removed all validations here as they seem redundant.

internal/resources/syntheticmonitoring/resource_check_alert.go
VikaCep: > Beyond that suggestion, I'm a little bit confused on this, probably because I don't know much about the Terraform provider implementation or how it actually works. In this function we receive a ResourceData that I understand matches the definition written by the user in the .tf file. So we expect that definition to contain a check_id as defined in the docs. Is this correct?
But then here we set the ID to the same check_id. I guess we have to do that due to how Terraform works? Reading the doc for the SetId method

Yes, that's correct. The `ResourceData` object represents the resource configuration as defined in the `.tf` file. When we create a resource, we need to set it an ID which Terraform uses to track the resource in the state file. I'm setting the check_id as its ID as it's unique, and those IDs are scoped to their resource type (so different resource types can use the same ID value without conflict).

As an example, this is the state entry from `terraform.tfstate` that was generated for the the alerts resource I created when testing these changes:
image

And this is the resource definition
image

You can notice how the generated state entry contains a `check_id` plus an `id` property.

ka3de (approved):

PR #2129, opened 2025-04-14
Support to manage basic k6 resources

### LINKED ISSUE #2086: Creation of k6 resources ( project + role assignments)
### Feature Request

Hey there,

we would like to create K6 projects and assign teams permissions to these projects.

I was not able to find the matching terraform resources yet, is that something on the roadmap?

The REST API is here: https://grafana.com/docs/grafana-cloud/testing/k6/reference/cloud-rest-api/v6/projects/#create-a-project

joanlopez:
It adds support for the following:

- Resources:
  - Project
  - Project limits
  - Load test (w/ script)
  - Installation (k6 App)

- Data sources:
  - Project
  - Project limits
  - Load test
  - Load tests (to query multiple load tests by project id)

Closes #2086


.github/CODEOWNERS
```diff
@@ -4,6 +4,7 @@
 /internal/resources/cloud/*                 @grafana/platform-monitoring @grafana/grafana-com-maintainers
 /internal/resources/cloudprovider/*         @grafana/platform-monitoring @grafana/middleware-apps
 /internal/resources/connections/*           @grafana/platform-monitoring @grafana/middleware-apps
+/internal/resources/k6/*                    @grafana/platform-monitoring @grafana/k6-core
```
joanlopez: Looks like I don't have permission to add this team to the repository, can you help, please? 🙏🏻  cc/ @spinillos @Duologic

joanlopez: Could you also add: `grafana/k6-backend`, please?
This will likely be also co-owned by https://github.com/orgs/grafana/teams/k6-backend

.github/workflows/acc-tests.yml
```diff
@@ -41,6 +41,7 @@ jobs:
             GRAFANA_AUTH=cloud-instance-tests:auth
             GRAFANA_ONCALL_ACCESS_TOKEN=cloud-instance-tests:oncall-token
             GRAFANA_SM_ACCESS_TOKEN=cloud-instance-tests:sm-token
+            GRAFANA_K6_ACCESS_TOKEN=cloud-instance-tests:k6-token
```
joanlopez: Do you know where/how should I define/create this? Thanks! 🙏🏻 

cc/ @spinillos @Duologic

joanlopez: Also, note that k6 resources rely on `stack_id`, is that set for "cloud instance tests"? 🤔

internal/resources/k6/resource_project.go
```diff
@@ -0,0 +1,289 @@
...
+// Metadata returns the resource type name.
+func (r *projectResource) Metadata(_ context.Context, _ resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = resourceProjectName
+}
+
+// Schema defines the schema for the resource.
+func (r *projectResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Manages a k6 project.",
+		Attributes: map[string]schema.Attribute{
```
patrickeasters, line 63: Would it be possible to expose the Grafana folder ID here too? That'd enable using `grafana_folder_permission` to set RBAC. New projects only grant the Admin role access, so I could see this being a common requirement.

internal/resources/cloud/resource_k6_installation.go
```diff
@@ -0,0 +1,175 @@
...
+
+	"github.com/grafana/terraform-provider-grafana/v3/internal/common"
+)
+
+func resourceK6Installation() *common.Resource {
+	schema := &schema.Resource{
+
+		Description: `
+Sets up the k6 App on a Grafana Cloud instance and generates a token. 
+Once a Grafana Cloud stack is created, a user can either use this resource or go into the UI to install k6.
```
Duologic, line 23: Out of curiosity, why isn't this service part of the rergular Cloud Stack (via Stack State Service)? This pattern also exists for Synthetic Monitoring and we've notice it causes friction.

For example, someone already clicked the button in the UI and now tries to create it with Terraform as well, what happens then? Does the API return a token as well?

.github/workflows/acc-tests.yml
Duologic: You can create a new secret in Vault at `ci/repo/grafana/terraform-provider-grafana/<secret name>` then referrence it here as `GRAFANA_K6_ACCESS_TOKEN=<secret name>:k6-token`.

I'm not sure to which 'stack' is referred with `stack_id`, in the docs on this PR I see that k6 also has a 'stack' concept... either way, if you know the id upfront you can also put it in the same secret and read it here.

internal/resources/cloud/resource_k6_installation.go
joanlopez: Let me invoke @d14c here, as I think he's the best candidate to answer to this one.

.github/workflows/acc-tests.yml
joanlopez: How I'm supposed to know in what stack do I need to create the k6 token, if I can't access/see the secret in use?
I guess, here, I need to know `GRAFANA_URL` (and likely `GRAFANA_AUTH`) in order to log in, and generate the token. Right?

Then, I hope to be able to create a new secret at `ci/repo/grafana/terraform-provider-grafana/...`.

joanlopez: It worked, thanks @Duologic for your assistance! 💟 🙇🏻

internal/resources/cloud/resource_k6_installation.go
d14c: > why isn't this service part of the rergular Cloud Stack (via Stack State Service)?

When the puguin was implemented SSS was not yet the service entirely responsible of provisioning the plugins, but neverthenless it seems good seperation to me that SSS is only responsible of the plugin provisioning /management (including jsonData nad secureJsonData) and the k6-cloud responsible of intitating the app state from k6-cloud perspective.

> For example, someone already clicked the button in the UI and now tries to create it with Terraform as well, what happens then? Does the API return a token as well?

yes the initialization api is idempotent, if the app is already intialized it will return the master token.

internal/resources/k6/resource_project.go
joanlopez: Yeah, that's something I had in mind since the very beginning, but I was waiting for the most recent release of the k6 API (v1.2.0) to re-generate the Go client and apply all the associated updates to this PR (see most recent commits). Now it's done, concretely in https://github.com/grafana/terraform-provider-grafana/pull/2129/commits/f9a4fa9ebc6fc343de7323089d641c24c1710248.

But anyway, thanks for the reminder @patrickeasters! :)

internal/resources/examples_test.go
```diff
@@ -103,6 +103,13 @@ func TestAccExamples(t *testing.T) {
 				testutils.CheckCloudInstanceTestsEnabled(t)
 			},
 		},
+		{
+			category: "k6",
+			testCheck: func(t *testing.T, filename string) {
+				t.Skip() // TODO: Make all examples work
```
joanlopez, line 109: All tests here are skipped, so I added the one for k6 examples but left it skipped as well, for now.

Alex3k:
@joanlopez - would this enable users to change the API used to `api.staging.k6.io` or does it hard code it to `api.k6.io`? I ask as we have a scenario internally where we use both

joanlopez:
> @joanlopez - would this enable users to change the API used to `api.staging.k6.io` or does it hard code it to `api.k6.io`? I ask as we have a scenario internally where we use both

The [k6 client](https://github.com/grafana/k6-cloud-openapi-client-go), by default connects to `api.k6.io`, as you can see [here](https://github.com/grafana/k6-cloud-openapi-client-go/blob/main/k6/configuration.go#L97). But, as you can also see [here](https://github.com/grafana/terraform-provider-grafana/blob/66c7ebe4e74020edf4fc273dcadec0eb894b554f/pkg/provider/configure_clients.go#L371-L375), the Terraform provider overwrites that configuration if either `k6_url` attribute or `GRAFANA_K6_URL` environment variable are provided.

Alex3k:
@joanlopez when this is released could we update https://docs.google.com/spreadsheets/d/1hBpYQpOq9MBlt1Fbf6A-JDsZmcf7y9BC-zk0hbJVCgo/edit?gid=0#gid=0 please? I wasn't sure if this is part of an official terraform provider release process or not but this sheet is proving incredibly useful in #terraform-provider

docs/data-sources/k6_load_test.md
```diff
@@ -0,0 +1,49 @@
...
+
+```terraform
+resource "grafana_k6_project" "load_test_project" {
+  name = "Terraform Load Test Project"
+}
+
+resource "grafana_k6_load_test" "test_load_test" {
+  project_id = grafana_k6_project.load_test_project.id
+  name       = "Terraform Test Load Test"
+  script     = <<-EOT
```
fornfrey, line 23: Can `script` accept a `.tar` archive? If not, will the provider support multi-file test scripts in any other way?

joanlopez: That's a very good question @fornfrey. The short answer is: no.

The challenge here is that we don't have an easy and reproducible way to check for "changes" in the script if they use an archive. I've seen that, in case of creating a new load test with an archive, then the "script" can be retrieved from the API, but it returns some form of a UTF-8 representation of the contents of the archive, which looks similar to multipart, but doesn't make it easy to check for updates.

So, in this case, we should find an alternative way for doing so. For instance, the provider could define a couple of attributes, like `script_file` and `script_file_hash`, so it uses the hash of the local file to detect whenever the file pointed by the `script_file` attribute has changed. AFAIK that's the pattern that some providers use when dealing with files within Terraform modules.

That said, I'd suggest that, for now, we ship this initial support without, and to add that support in the next iteration.
So we unblock this initial release, users can test it, and give us feedback for that next iteration, beyond this specific "feature".

Wdyt?

fornfrey: My initial concern was that we might need to change the `scrip` field schema to an object to support different types of scripts, but as you suggested, a separate field also solves this problem. I don't see a blocker here 👍 

> For instance, the provider could define a couple of attributes, like script_file and script_file_hash, so it uses the hash of the local file to detect whenever the file pointed by the script_file attribute has changed. 

To clarify, will they be calculated internally in the provider based on `script_file` input rather than being requested from the user?

fornfrey (approved):

docs/data-sources/k6_load_test.md
joanlopez: > To clarify, will they be calculated internally in the provider based on script_file input rather than being requested from the user?

I'm not 100% sure how that's usually handled, but yes, I guess that could be an option. Alternatively, we may suggest users to use [the `filesha256` function](https://developer.hashicorp.com/terraform/language/functions/filesha256).

joanlopez:
> @joanlopez when this is released could we update https://docs.google.com/spreadsheets/d/1hBpYQpOq9MBlt1Fbf6A-JDsZmcf7y9BC-zk0hbJVCgo/edit?gid=0#gid=0 please? I wasn't sure if this is part of an official terraform provider release process or not but this sheet is proving incredibly useful in #terraform-provider

I don't know how frequently is this being updated, nor if it is up to date, but I have added the rows that correspond to resources/data sources added in this PR 👍🏻

PR #2162, opened 2025-05-06
LBAC: Error out if Grafana provider is not correctly configured

IevaVasiljeva:
LBAC data source rule resource requires Grafana provider to be configured. Previously we weren't validating if the provider was correctly configured, which lead to a crash in case the provider was not configured as expected. Report of a crash: https://github.com/grafana/grafana/issues/104459


eleijonmarck (approved):
lgtm! tested out that we catch the change 👍

PR #2171, opened 2025-05-13
feat:fe-o11y: Add geolocation settings

kpelelis:
This commit adds support for geolocation settings in Frontend Observability App resource. The settings available are

* `geolocation.enabled`: Toggle geolocation feature
* `geolocation.level`: Control the level of specificity of geolcation
* `geolocation.country_denylist`: Disable the feature for a specific comma separated list of ISO country codes.


docs/resources/frontend_o11y_app.md
```diff
@@ -42,7 +45,7 @@ resource "grafana_frontend_o11y_app" "test-app" {
 - `allowed_origins` (List of String) A list of allowed origins for CORS.
 - `extra_log_attributes` (Map of String) The extra attributes to append in each signal.
 - `name` (String) The name of Frontend Observability App. Part of the Terraform Resource ID.
-- `settings` (Map of String) The key-value settings of the Frontend Observability app. Available Settings: `{combineLabData=(0|1)}`
+- `settings` (Map of String) The key-value settings of the Frontend Observability app. Available Settings: `{combineLabData=(0|1),geolocation.level=(0|1),geolocation.level=0-4,geolocation.country_denylist=<comma-separated-list-of-country-codes>}`
```
Alex3k, line 48: Are we able to point to docs which outline what the levels mean?

Alex3k (commented):
Thank you for doing this so quickly! I added a question around the docs

codecapitano:
Offtopic but related:

I think this is another good example why should get rid of the double activation logic for Geolocation tracking.
TF users will likely not notice that they need a second activation in the web-sdk

docs/resources/frontend_o11y_app.md
kpelelis: https://grafana.com/docs/grafana-cloud/monitor-applications/frontend-observability/geolocation/#enable-geolocation

codecapitano (approved):
Looks good thanks for the fix👍

PR #2173, opened 2025-05-14
Recreate deleted k6 resources

fornfrey:
* Ensure that k6 resources (Projects, Load Tests) create a Terraform drift instead of raising an error when deleted outside of Terraform.
* Add tests to verify the resources are removed from the state when not found on the backend.
* Update existing tests to generate unique k6 project names.

### DIFF
--- a/internal/resources/k6/resource_loadtest.go
+++ b/internal/resources/k6/resource_loadtest.go
@@ -154,12 +154,8 @@ func (r *loadTestResource) Read(ctx context.Context, req resource.ReadRequest, r
 	k6Req := r.client.LoadTestsAPI.LoadTestsRetrieve(ctx, state.ID.ValueInt32()).
 		XStackId(r.config.StackID)
 
-	lt, httpResp, err := k6Req.Execute()
-
-	if httpResp != nil && httpResp.StatusCode == http.StatusNotFound {
-		resp.State.RemoveResource(ctx)
-		return
-	} else if err != nil {
+	lt, _, err := k6Req.Execute()
+	if err != nil {
 		resp.Diagnostics.AddError(
 			"Error reading k6 load test",
 			"Could not read k6 load test with id "+strconv.Itoa(int(state.ID.ValueInt32()))+": "+err.Error(),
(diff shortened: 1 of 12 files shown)


joanlopez (approved):
Looks good, thanks @fornfrey!

I think it could be a very nice idea to try to write a test that:
- Loads a config directory, with a given resource (and so it creates it)
- Use the Go client/direct HTTP request to remove it.
- Try to reload the configuration, and see how it behaves (no error).

However, I'm not sure how easy that would be, and if there's any similar test for any other resource within the provider. If so, and it's easy to copy and adjust, I'd try it, if not, feel free to merge it as-is! We could look at it later.

Thanks! 🙌🏻

internal/resources/k6/resource_loadtest.go
```diff
@@ -166,8 +171,12 @@ func (r *loadTestResource) Read(ctx context.Context, req resource.ReadRequest, r
 	scriptReq := r.client.LoadTestsAPI.LoadTestsScriptRetrieve(ctx, state.ID.ValueInt32()).
 		XStackId(r.config.StackID)
 
-	script, _, err := scriptReq.Execute()
-	if err != nil {
+	script, httpResp, err := scriptReq.Execute()
+
+	if httpResp != nil && httpResp.StatusCode == http.StatusNotFound {
+		// 404 response from the script endpoin for an existing test means that the script is undefined
```
joanlopez:
```suggestion
		// 404 response from the script endpoint for an existing test means that the script is undefined
```

joanlopez (approved):

spinillos (approved):
Thanks for fix the cloud tests 😃

internal/resources/k6/resource_loadtest.go
```diff
@@ -153,8 +154,12 @@ func (r *loadTestResource) Read(ctx context.Context, req resource.ReadRequest, r
 	k6Req := r.client.LoadTestsAPI.LoadTestsRetrieve(ctx, state.ID.ValueInt32()).
 		XStackId(r.config.StackID)
 
-	lt, _, err := k6Req.Execute()
-	if err != nil {
+	lt, httpResp, err := k6Req.Execute()
+
+	if httpResp != nil && httpResp.StatusCode == http.StatusNotFound {
+		resp.State.RemoveResource(ctx)
```
vortegatorres, line 160: Do you think it make sense to add a warning here?
```
		resp.Diagnostics.AddWarning(
			"Resource not found",
			fmt.Sprintf("The k6 load test with ID %d was not found. It may have been deleted outside of Terraform. Removing from state.", state.ID.ValueInt32()),
		)
```

vortegatorres (approved):

internal/resources/k6/resource_loadtest.go
fornfrey: Sorry, I missed your message. Most resources in the provider don't log anything when deleted resources are removed from the state. I suggest keeping the k6 resources consistent with the others.

PR #2179, opened 2025-05-19
Add SCIM Config resource

colin-stuart:
This PR adds a new `grafana_scim_config` resource that manages SCIM configuration using the app platform APIs. The resource allows users to enable/disable user and group synchronization.

Changes:
- Added new SCIM config resource using app platform SDK
- Added acceptance tests
- Added documentation and example

Dependent on https://github.com/grafana/grafana-enterprise/pull/8651
Closes https://github.com/grafana/identity-access-team/issues/1415


colin-stuart:
Ah, I need to fix the import of the SCIM config API for this new resource, I added a `replace` to go.mod to import the SCIM config API when I was testing this locally.

linoman (approved):

internal/resources/grafana/resource_scim_config.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+	err = d.Set("enable_group_sync", scimConfig.Spec.EnableGroupSync)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	return nil
+}
+
+func UpdateSCIMConfig(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
dmihai: Is there any difference between the `UpdateSCIMConfig()` and `CreateSCIMConfig()` methods? Can we use one method for handling both create and update actions?

internal/resources/grafana/resource_scim_config.go
```diff
@@ -0,0 +1,279 @@
...
+			EnableGroupSync: d.Get("enable_group_sync").(bool),
+		},
+	}
+
+	jsonData, err := json.Marshal(scimConfig)
+	if err != nil {
+		return diag.FromErr(fmt.Errorf("failed to marshal SCIM config: %w", err))
+	}
+
+	url := fmt.Sprintf("%s://%s%s/apis/scim.grafana.app/v0alpha1/namespaces/stacks-%d/config/default",
```
dmihai: There is no slash (`/`) separating the host and basepath. Are we sure that the host ends with slash? Or the basepath starts with a slash?

Also this URL works only for cloud instances, on prem instances have the `default` namespace if I'm not mistaken. Is this intended?

colin-stuart: Good catches, both of these are fixed now.

internal/resources/grafana/resource_scim_config.go
colin-stuart: Good point, the differences are minor enough that I merged these two into a single method.

dmihai (approved):

Issue #1917, opened 2024-11-20
Add support for Grafana Cloud integration installation


lsibilla:
### Feature Request

Hello,

I would like to be able to install [Grafana Cloud integrations](https://grafana.com/docs/grafana-cloud/monitor-infrastructure/integrations/) using Terraform.

Thank you,

Laurent

PR #2191, opened 2025-05-23
Grafana Cloud Integrations

clementduveau:
Hello there 👋 

This is a "simple" implementation of Grafana Cloud Integrations in the Terraform provider.

Integrations don't have an OpenAPI specs (because they rely on the webclient to create the dashboards, it's a mix), nor they have a client (AFAIK). So the client relies on direct API calls for the "easy start" part.

What works and have been tested:

- Installing integrations
- Installing with/without alerts or logs
- Disabling integration
- import

What I didn't test or even tried:
- Adding a datasource (because why ?)

What could be improved

- Update integrations ? I guess it should be an option like `always-try-update` to consider than the resource needs a change, forcing the upgrade.
- Unit tests. I tested manually everything as I didn't know how to write tests for Cloud

EDIT: tested more things


clementduveau:
Solves #1917

clementduveau:
Since it's been a while, I just want to add some documentation around how it works, otherwise I will forget.

When someone wants to install the integration, they click on the "Install dashboard and alerts" button. The chain of action:

- An API call is made to an endpoint. This endpoint will install the recording rules and alert rules as well as changing the integration status to `enabled`
- Another API call is made to retrieve some metadata about the integrations (id, name, folder, a URL to find the dashboards)
- The web client call the URL to get the dashboards JSON
- The web client push the JSON into the folder

So the client is the one doing most of the installation. This terraform ressources does that the same way.

When uninstalling, it's pretty much the same thing, but instead of getting the JSON, it just removes the folder with the dashboards and call the endpoint to disable & remove alerts.

docs/resources/integration.md
```diff
@@ -0,0 +1,123 @@
...
+Manages Grafana Cloud integrations.
+
+* [Official documentation](https://grafana.com/docs/grafana-cloud/data-configuration/integrations/)
+
+Required access policy scopes:
+
+* folders:read
+* folders:write
+* dashboards:read
+* dashboards:write
```
jeschkies, line 24: It's also installing alerts and rules isn't it?

internal/resources/integrations/client.go
```diff
@@ -0,0 +1,371 @@
+package integrations
```
jeschkies, line 1: I see the other clients in `internal/common`. Could you move it?

internal/resources/integrations/client.go
```diff
@@ -0,0 +1,371 @@
...
+	err := c.doAPIRequest(ctx, http.MethodGet, path, nil, &response)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get integration %s: %w", slug, err)
+	}
+
+	return &response, nil
+}
+
+// PostDashboards posts dashboards for an integration with the given configuration
+func (c *Client) PostDashboards(ctx context.Context, slug string, config *InstallationConfig) (*GetDashboardsResponse, error) {
```
jeschkies, line 102: I thought the integrations API is doing that.

jeschkies (changes requested):

Dasomeone:
Thanks @jeschkies for reminding me this exists!

@clementduveau this is an interesting one, and it's related to work that we're picking up that may affect this.
You're spot on in that the integrations-api is more a resource/source of truth for grabbing the integration to install, and (currently) installing alerts/rules, while the frontend in Grafana Cloud handles the installation of the dashboards themselves.

With the rollout of [Grafana Managed Alerts](https://grafana.com/docs/grafana/latest/alerting/alerting-rules/create-grafana-managed-rule/) this will extend to alerts/rules as well, so we'll need to touch base on what all is needed and desired here in order to get all that added in :)

clementduveau:
Cleaning up the repo. The PR is relevant only for documentation purposes as the backend is evolving.

PR #2204, opened 2025-06-03
Convert k6 resource IDs to string

fornfrey:
* Convert k6 resource IDs from int32 to string to work correctly in the crossplane provider when generated with upjet.
* Upgrade resource schema versions and implement the `UpgradeState` methods to migrate the existing states. This makes the change backward compatible.
* Add a check for empty ID in the `Read` methods to prevent errors when called from crossplane `Observe()` method for the first time before the resource is created. 
* Add tests to verify empty plans when upgrading to the new schemas.

### DIFF
--- a/internal/resources/k6/resource_loadtest.go
+++ b/internal/resources/k6/resource_loadtest.go
@@ -21,8 +21,9 @@ import (
 
 // Ensure the implementation satisfies the expected interfaces.
 var (
-	_ resource.ResourceWithConfigure   = (*loadTestResource)(nil)
-	_ resource.ResourceWithImportState = (*loadTestResource)(nil)
+	_ resource.ResourceWithConfigure    = (*loadTestResource)(nil)
+	_ resource.ResourceWithImportState  = (*loadTestResource)(nil)
+	_ resource.ResourceWithUpgradeState = (*loadTestResource)(nil)
 )
 
 var (
@@ -40,12 +41,22 @@ func resourceLoadTest() *common.Resource {
 }
 
 // loadTestResourceModel maps the resource schema data.
-type loadTestResourceModel struct {
+type loadTestResourceModelV0 struct {
 	ID                types.Int32  `tfsdk:"id"`
 	ProjectID         types.Int32  `tfsdk:"project_id"`
 	Name              types.String `tfsdk:"name"`
 	Script            types.String `tfsdk:"script"`
 	BaselineTestRunID types.Int32  `tfsdk:"baseline_test_run_id"`
+	Created           types.String `tfsdk:"created"`
+	Updated           types.String `tfsdk:"updated"`
+}
+
+type loadTestResourceModelV1 struct {
+	ID                types.String `tfsdk:"id"`
+	ProjectID         types.String `tfsdk:"project_id"`
+	Name              types.String `tfsdk:"name"`
+	Script            types.String `tfsdk:"script"`
+	BaselineTestRunID types.String `tfsdk:"baseline_test_run_id"`
 	Created           types.String `tfsdk:"created"`
 	Updated           types.String `tfsdk:"updated"`
 }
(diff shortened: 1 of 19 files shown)


internal/resources/k6/resource_loadtest.go
```diff
@@ -40,7 +41,7 @@ func resourceLoadTest() *common.Resource {
 }
 
 // loadTestResourceModel maps the resource schema data.
-type loadTestResourceModel struct {
+type loadTestResourceModelV0 struct {
 	ID                types.Int32  `tfsdk:"id"`
 	ProjectID         types.Int32  `tfsdk:"project_id"`
```
yorugac, line 46: 2 models are basically needed for backwards compatibility, correct?

internal/resources/k6/resource_project_limits.go
```diff
@@ -151,15 +228,25 @@ func (r *projectLimitsResource) Read(ctx context.Context, req resource.ReadReque
 	// Set the ID to match the project_id
 	state.ID = state.ProjectID
 
+	intID, err := strconv.ParseInt(state.ID.ValueString(), 10, 32)
```
yorugac, line 231: There is no "crossplane check" for ID being empty in this resource, it seems. Why?

yorugac (changes requested):
@fornfrey, thanks for this :raised_hands: 
I don't see anything obviously off (within my limited knowledge of this code base), except for Read method in project limits resource: please see my comment.

internal/resources/k6/resource_loadtest.go
fornfrey: Yes, this approach is suggested in the Terraform docs https://developer.hashicorp.com/terraform/plugin/framework/resources/state-upgrade#stateupgrader-with-priorschema

internal/resources/k6/resource_project_limits.go
fornfrey: `project_limits` does not have its own ID. Instead, we use the supplied `ProjectID`, which is expected to be set by the practitioner explicitly in the resource.

yorugac: :thinking: so it's a characteristic of the API; I didn't realize that. OK then, thanks for clarifying.

yorugac (approved):

joanlopez (approved):

joanlopez:
Is there any chance to test these changes with Crossplane before merging and releasing it (maybe locally, pointing to the "dev" provider)? 🤔 

Regarding existing states, I guess the upgrader solution should be fine - thanks for adding the according tests 👌🏻 🙇🏻

yorugac:
@joanlopez, AFAIS it's possible to test Crossplane provider's _runtime_ with the branch (commit hash) of the Terraform provider. But _generation_ of Crossplane provider requires a Terraform release.

Looking [here](https://discuss.hashicorp.com/t/terraform-use-local-provider-plugin/26153/5), I think it might be possible to also test Crossplane provider's generation with the local "binary" of the Terraform provider but it must be built first, from the branch. Is it possible to quickly build that binary? I'll be happy to try that, if there's a known command for that. (just in case though, there's no guarantee that it'd work...)

yorugac:
> But generation of Crossplane provider requires a Terraform release.

Just to clarify: since we're changing types of fields in this PR, so changing structure of the objects, Crossplane's generation is also needed for a full test.

fornfrey:
@yoruga the binary can be built with just `go build`, and terraform can be configured to use it as a dev override as described here https://github.com/grafana/terraform-provider-grafana/blob/main/README.md#local-development-with-grafana

yorugac:
OK, running `go build` is simple enough to create a binary :smile: Thanks @fornfrey!

So simple testing with Crossplane provider was done; some notable observations:
1. CRUD for `project` is working in a rather straight-forward manner.
1. `loadtest` is creating a test but not running it. 
2. `loadtest` can be modified and subsequent test runs are picking the change up :+1: Though as we already know, some would want to pass an archive there instead of the script.
3. When `loadtest` is deleted, all the test runs associated with it are deleted.
4. `projectlimits` is changing existing values when created or edited.
5. When `projectlimits` is deleted, the values don't return to the default ones but remain as the last ones configured.
6. AFAIS, `projectlimits` was calling `PATCH` even on creation and deletion, which matches up with observations from point 5 and 6.

I hope all of the above are expected at the moment.

I'll run some more tests. And there is one more thing I'd like to look into for crossplane-provider (the way to reference projects without an ID) but that one shouldn't require changes in Terraform provider. So preliminary I believe this PR did the main fix: generated Crossplane objects can be created and manipulated at a basic level :raised_hands:

fornfrey:
@yorugac  Thank you for testing this! All your observations are the expected behavior of the API. 
Considering we fixed the main problem and no other issues were found, I'll merge the PR.

PR #2206, opened 2025-06-04
cloudprovider: update docs with defaults

tristanburgess:
fixes https://github.com/grafana/grafana-csp-app/issues/586

This adds default values to relevant field descriptions in CloudProvider resource schemas. 

Default values aren't being automatically added to docs by tfplugindocs.  This appears to be because we use the newer [terraform-plugin-framework](https://developer.hashicorp.com/terraform/plugin/framework), which requires fields with Default values to be Computed, but tfplugindocs won't list Computed fields as having default values (since Computed means the providing service (Cloud Provider API) can provide its own value).


HCloward (approved):
Looks good to me! Thanks @tristanburgess

PR #2230, opened 2025-06-26
oncall: Allow oncall schedules of type web

wettick:
Add support for managing oncall schedules of type `web`. This supports a use case where one wants to enforce the creation of a schedule (and name, settings) via terraform, but manage the shifts themselves manually in the web application.


matiasb (approved):
LGTM

olegbespalov (approved):

PR #2231, opened 2025-06-26
feat(tools): add terraform resource issue template generator

simaarfania:
This PR follows up on https://github.com/grafana/terraform-provider-grafana/pull/2228 as a part of Hackathon 13 (see https://github.com/grafana/deployment_tools/issues/239391). It adds a Go script that generates a Github issue template with a searchable multi-select dropdown of terraform resources and data sources. This helps automate issue triaging by allowing users to precisely specify which terraform resources are affected by bugs, which can then be used downstream to assign issues to owning teams. 

Resource discovery is driven through the provider schema.

See a successful run of the Github action here: https://github.com/grafana/terraform-provider-grafana/actions/runs/15981317587/job/45076105244?pr=2231


tools/auto-triage/main.go
```diff
@@ -0,0 +1,361 @@
...
+	"os"
+	"path/filepath"
+	"regexp"
+	"sort"
+)
+
+func main() {
+	// Try methods in priority order:
+	// 1. Catalog files from PR #2228 (when available)
+	// 2. Reliable fallback list
```
Duologic: I've been generate the provider schema as a source of truth for which resources exist. See https://github.com/grafana/terraform-provider-grafana/blob/duologic/backstage/resource_ownership/scripts/generate_schema.sh 

I'm currently generating this schema every time but perhaps we can require it to be updated on changes to the code and commit the schema directly? Then have a CI job to keep the comitted schema in sync with the code.

spinillos:
I'll set all the scripts under `scripts` folder as its the way that we use in general at Grafana.

tools/auto-triage/main.go
simaarfania: Makes sense! Is the idea here to use the schema to generate the available resources to create an issue for?

Duologic: I think that's a solid plan, the schema is definitive.

simaarfania: I _think_ I've gone ahead and done that. The Go script is correctly sourcing from the provider schema. I tried modeling the GHA off of other Grafana owned repos but not sure how good that's looking.

cindy (approved):
I think the current CI job failure is unrelated to your changes 🤔 Otherwise, the changes look good to me!

tools/auto-triage/main.go
spinillos: Maybe we can add this generator inside the Makefile and use the CI step only to verify if its updated. As the same way we generate the [docs](https://github.com/grafana/terraform-provider-grafana/blob/main/GNUmakefile#L72) and this [check](https://github.com/grafana/terraform-provider-grafana/blob/main/.github/workflows/unit-tests.yml#L33).

scripts/generate_issue_template.go
```diff
@@ -0,0 +1,149 @@
...
+	os.WriteFile(schemaFile, data, 0600)
+	fmt.Println("Schema updated")
+	return schema
+}
+
+func generateTemplate(resources []string) {
+	os.MkdirAll(".github/ISSUE_TEMPLATE", 0755)
+
+	template := `name: Bug Report (Enhanced)
+description: File a bug report with resource dropdown
```
Duologic:
```suggestion
	template := `# NOTE: this template is automatically generated
name: Bug Report (Enhanced)
description: File a bug report with resource dropdown
```

Duologic (approved):
LGTM, one nitty comment.

PR #2240, opened 2025-07-03
feat: add auto-triage via Backstage

simaarfania:
Related to Hackathon project which is built off of https://github.com/grafana/deployment_tools/issues/239391.

- **Triggers**: When issues are opened with "Affected Resource(s)" 
- **Queries**: Backstage API to find resource ownership -- *needs to be tested**
- **Assigns**: Issues to team GitHub projects automatically **needs to be tested**
- **Labels**: Adds `team/[name]` labels  
- **Notifies**: Mentions owning teams in comments


scripts/backstage-lookup/main.go
```diff
@@ -0,0 +1,244 @@
...
+	"net/http"
+	"os"
+	"regexp"
+	"strings"
+	"time"
+)
+
+// Constants for API and configuration
+const (
+	defaultURL     = "https://enghub.grafana-ops.net"
```
spinillos: Should we get the value from the vault? I don't know if we should expose this...

simaarfania: I think I've done that now! Let me know if that's the best way to do it.

Duologic:
The logic from this PR has been swallowed by https://github.com/grafana/enghub-github-issue-assigner

PR #2241, opened 2025-07-04
oncall: trim whitespace for route routing_regex

mderynck:
Trim whitespace for oncall resource route `routing_regex` so it stops showing as changed when it has not changed.

### DIFF
--- a/internal/resources/oncall/resource_route.go
+++ b/internal/resources/oncall/resource_route.go
@@ -63,9 +63,7 @@ func resourceRoute() *common.Resource {
 				Required:    true,
 				Description: "Python Regex query. Route is chosen for an alert if there is a match inside the alert payload.",
 				StateFunc: func(v interface{}) string {
-					input := v.(string)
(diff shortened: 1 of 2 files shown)


mderynck:
Looking into the failing tests

willgallego-grafana (approved):
nice catch! :shipit:

internal/resources/oncall/resource_route.go
```diff
@@ -62,6 +62,11 @@ func resourceRoute() *common.Resource {
 				Type:        schema.TypeString,
 				Required:    true,
 				Description: "Python Regex query. Route is chosen for an alert if there is a match inside the alert payload.",
+				StateFunc: func(v interface{}) string {
+					input := v.(string)
+					trimmed := strings.TrimSpace(input)
+					return trimmed
```
olegbespalov:
```suggestion
					return strings.TrimSpace(v.(string))
```

olegbespalov (approved):
LGTM :+1: suggested non blocking minor simplification

PR #2248, opened 2025-07-09
fix: update schema workflow to use comment-based approach for PRs

simaarfania:
This PR fixes the `update schema` Github workflow (introduced in https://github.com/grafana/terraform-provider-grafana/pull/2231) to use a comment-based approach when the schema needs to be updated. Previously the workflow attempted to update the schema and commit it, but due to permission issues, this was failing. 

See an example of the comment by the `github-actions` bot below.


Duologic:
Can the CI job also fail if running the command is required, effectively blocking the merge?

simaarfania:
> Can the CI job also fail if running the command is required, effectively blocking the merge?

Successfully does this now. See [here](https://github.com/grafana/terraform-provider-grafana/actions/runs/16224305511/job/45812413559?pr=2248)

.github/workflows/update-schema.yml
```diff
@@ -1,40 +1,95 @@
...
+            if ! diff -q provider_schema.json current_schema.json > /dev/null; then
+              echo "schema_changed=true" >> $GITHUB_OUTPUT
+              echo "Schema changes detected!"
+            else
+              echo "schema_changed=false" >> $GITHUB_OUTPUT
+              echo "No schema changes detected."
+            fi
+          else
+            echo "schema_changed=true" >> $GITHUB_OUTPUT
+            echo "No existing schema file found."
```
jjo: nit: maybe being more explicit on the expectation, here? (which is a bit different from the above `diff` case)
```
echo "Required provider_schema.json file not found"
```

jjo (approved):
LGTM w/nit

cindy (approved):

PR #2254, opened 2025-07-14
fix(roles): Handle issue with duplicate IDs cause `400` despite using existing resources

twu:
Related-To: https://github.com/grafana/support-escalations/issues/17189, https://github.com/grafana/grafana-enterprise/pull/9055

## What?

Adds duplicate checking to `grafana_role_assignment_item` resource to prevent sending duplicate team/user/service account IDs in role assignment API requests.

- Modified `Create()` method in `resource_role_assignment_item.go` to check for existing assignments before appending new ones
- Added `TestAccRoleAssignmentItem_NoDuplicates` test case to verify the fix
- Applied the same logic to teams, users, and service accounts

## Why?

When multiple `grafana_role_assignment_item` resources target the same role, the provider was  appending assignments to existing ones without checking for duplicates. This caused API requests with duplicate IDs, resulting in 400 errors from Grafana. I opened a PR to gracefully handle duplicates but I'm not 100% whether this is how it should be done (See https://github.com/grafana/grafana-enterprise/pull/9055)

**Reproduce**
```hcl
data "grafana_role" "role" {
  name     = "plugins:grafana-oncall-app:admin"
}

data "grafana_team" "team_a" {
  name     = "A"
}

resource "grafana_user" "alice" {
  email    = "alice@localhost"
  login    = "alice"
  password = "alice"
}

resource "grafana_role_assignment_item" "alice" {
  role_uid = data.grafana_role.role.uid
  user_id  = grafana_user.alice.id
}

resource "grafana_role_assignment_item" "team_a" {
  role_uid = data.grafana_role.role.uid  // Same role as alice assignment
  team_id  = data.grafana_team.team_a.id
}
```

This fails with `Failed to set role assignments: [PUT /access-control/roles/{roleUID}/assignments] setRoleAssignments (status 400): {}`

The second assignment item would fetch existing assignments `[users: [15], teams: []]`, append team 2, but due to state inconsistencies could result in duplicate entries like `[users: [15], teams: [2, 2]]`.

We now check if assignment already exists before adding, preventing duplicates.


.github/ISSUE_TEMPLATE/3-bug-report-enhanced.yml
```diff
@@ -125,6 +126,7 @@ body:
         - grafana_role_assignment (resource)
         - grafana_role_assignment_item (resource)
         - grafana_rule_group (resource)
+        - grafana_scim_config (resource)
```
twu: **C**: This looks like it came from somewhere else. SCIM is most likely also the thing that changed the schema.

spinillos (approved):

PR #2265, opened 2025-07-21
4565: Setup for TF for Alert Config for App Platform CRDS

matt-metivier:
# Add AlertConfig Resource Support for Asserts via Grafana App Platform API

## 📋 **Summary**

This PR adds support for managing Asserts AlertConfig resources through the Grafana App Platform API. The implementation provides a new Terraform resource `grafana_apps_asserts_alertconfig_v2alpha1` that allows users to configure alert policies for Asserts monitoring.

### **New Terraform Resource**
- **Resource Name**: `grafana_apps_asserts_alertconfig_v2alpha1`
- **Category**: Grafana Apps
- **API Group**: `asserts.ext.grafana.com/v2alpha1`

### **Resource Schema**
```hcl
resource "grafana_apps_asserts_alertconfig_v2alpha1" "example" {
  metadata {
    uid        = "my-alert-config"
    folder_uid = "folder-123"
  }

  spec {
    match_labels = {
      alertname = "HighCPUUsage"  # OR asserts_slo_name = "api-latency"
    }
    alert_labels = {
      severity = "warning"
      team     = "platform"
    }
    duration = "5m"
    silenced = false
  }

  options {
    overwrite = true
  }
}
```

### **Core Implementation**
- `internal/resources/appplatform/alert_config_resource.go` - Main resource implementation
- `pkg/provider/resources.go` - Resource registration

### **Test Infrastructure**
- `testdata/appplatform/alertconfigs.tf` - Example Terraform configuration
- `testdata/appplatform/README.md` - Updated documentation

### **Documentation**
- `docs/resources/apps_asserts_alertconfig_v2alpha1.md` - Resource documentation
- `internal/resources/appplatform/catalog-resource.yaml` - Backstage catalog entry

### **Required Fields**
- `spec.match_labels` - Must contain either `alertname` OR `asserts_slo_name`

### **Optional Fields**
- `spec.alert_labels` - Additional labels to add to alerts
- `spec.duration` - Alert evaluation duration (Prometheus format)
- `spec.silenced` - Whether the alert config is silenced
- `metadata.folder_uid` - Folder to save the resource in
- `options.overwrite` - Whether to overwrite existing resources

### **Computed Fields**
- `metadata.uuid` - Globally unique identifier
- `metadata.url` - Full URL of the resource
- `metadata.version` - Resource version


internal/resources/appplatform/alert_config_resource.go
radiohead, line 1: @matt-metivier I see you've started by implementing the resource from scratch. Could you instead use the generic resource instead? You can use Playlists as a reference implementation – https://github.com/grafana/terraform-provider-grafana/blob/main/internal/resources/appplatform/playlist_resource.go#L40-L43.

This way you don't need to implement client logic and the like, only the conversions from / to HCL schema.

Let me know if you need help with that!

matt-metivier: Updated the pr to be similar to that example,
 but I wanted to ask how do we setup this 
 
 ```
 	v2alpha1 "github.com/grafana/grafana/apps/asserts/alertconfig/pkg/apis/alertconfig/v2alpha1"
```

radiohead: Good question.

The expectation is that you'd use app SDK to generate Go files from your schemas and make them available in some sort of public Go module (e.g. a public Github repo in case of Grafana).

I know that in your case it's slightly more complicated – if you don't have a public Go module, one other option would be to generate the files and manually copy them over. It's a hack but it doesn't require setting up an extra repo.

Though I would recommend having a public repo with generated code for different languages anyway, in case you want to share schemas / client code for external API consumers.

matt-metivier: I see what you mean, I guess we can generate the GO files and have them in a public repo for Asserts. 
Thank You

PR #2271, opened 2025-07-22
Add grafana_k6_project_allowed_load_zones resource / data source

pmdz:
Implement allowed load zones management for k6 projects.
Relevant k6 public API docs:  [link](https://grafana.com/docs/grafana-cloud/testing/k6/reference/cloud-rest-api/load-tests/)

1. Add `grafana_k6_project_allowed_load_zones` resource
2. Add `grafana_k6_project_allowed_load_zones` data source

Issue: https://github.com/grafana/k6-cloud/issues/3512


internal/resources/k6/common.go
```diff
@@ -80,3 +82,72 @@ func (d *basePluginFrameworkDataSource) Configure(_ context.Context, req datasou
...
+	for _, k6LoadZoneID := range k6LoadZoneIds {
+		resp, _, err := client.LoadZonesAPI.LoadZonesList(ctx).
+			K6LoadZoneId(k6LoadZoneID).
+			XStackId(config.StackID).
+			Execute()
+		if err != nil {
+			return err
+		}
+
+		// If k6_load_zone_id is correct, response should contain exactly one element
```
fornfrey: Why not list all load zones at once and search for IDs in the returned list?

pmdz: Very good point, thank you! Done in 10b9a57b97cf94f2860873dac6a45fb589574472

internal/resources/k6/resource_project.go
```diff
@@ -96,6 +98,11 @@ func (r *projectResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 				Description: "The date when the project was last updated.",
 				Computed:    true,
 			},
+			"allowed_load_zones": schema.ListAttribute{
```
fornfrey: Why add it to the project resource? Managing permissions requires calling a different endpoint and makes the project resource more complex. It will also make it harder to evolve the permission management as new fields would also have to be added to the project resource. I suggest keeping project and load zone permissions management separate. 

The `grafana_k6_project_allowed_load_zones` is simpler and does not require the related project to be defined in terraform, which might be bummer for the default project.

internal/resources/k6/data_source_k6_project.go
```diff
@@ -80,6 +82,11 @@ func (d *projectDataSource) Schema(_ context.Context, _ datasource.SchemaRequest
 				Description: "The date when the project was last updated.",
 				Computed:    true,
 			},
+			"allowed_load_zones": schema.ListAttribute{
```
fornfrey: Related to my previous comment: consider adding a separate data source for LZ permissions.

internal/resources/k6/resource_project_allowed_load_zones.go
```diff
@@ -0,0 +1,233 @@
...
+	ctx = context.WithValue(ctx, k6.ContextAccessToken, r.config.Token)
+	_, httpResp, err := r.client.ProjectsAPI.ProjectsRetrieve(ctx, projectID).
+		XStackId(r.config.StackID).
+		Execute()
+
+	if httpResp != nil && httpResp.StatusCode == http.StatusNotFound {
+		resp.State.RemoveResource(ctx)
+		return
+	}
+
```
fornfrey: The 404 error in this case means the related project does not exist, which is a configuration error rather than a state diff. It shouldn't be handled gracefully.

```suggestion
```

internal/resources/k6/resource_project_allowed_load_zones.go
```diff
@@ -0,0 +1,233 @@
...
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	projectID := state.ProjectID.ValueInt32()
+
+	// Verify project exists
+	ctx = context.WithValue(ctx, k6.ContextAccessToken, r.config.Token)
+	_, httpResp, err := r.client.ProjectsAPI.ProjectsRetrieve(ctx, projectID).
```
fornfrey: IIRC, the permission list endpoints already returns a 404 if the project does not exist, so this call can be omitted.

internal/resources/k6/resource_project.go
pmdz: > Why add it to the project resource? Managing permissions requires calling a different endpoint and makes the project resource more complex. It will also make it harder to evolve the permission management as new fields would also have to be added to the project resource. I suggest keeping project and load zone permissions management separate.

I agree, I was just under the impression that this is a convention that I should follow.

> The grafana_k6_project_allowed_load_zones is simpler and does not require the related project to be defined in terraform, which might be bummer for the default project.

I'm not sure if I get "might be bummer for the default project" part correctly. Are you stating the fact or this is a suggestion that we should find a solution for that?

internal/resources/k6/resource_project_allowed_load_zones.go
```diff
@@ -0,0 +1,233 @@
...
+	diags := req.State.Get(ctx, &state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	projectID := state.ProjectID.ValueInt32()
+
+	// Clear allowed load zones (set to empty list)
+	err := setProjectAllowedLoadZones(ctx, r.client, r.config, projectID, []string{})
```
fornfrey: We should probably handle the 404 response here in case the related project is already removed.

fornfrey (changes requested):
Thanks a lot for the PR! I left a couple of suggestions regarding state management. Let me know if you have any questions or need help.

internal/resources/k6/resource_project_allowed_load_zones.go
pmdz: That is correct. Removed in 255e40d4ec9809e8546a84b5cdd9c71ec0a261f7

internal/resources/k6/resource_project.go
fornfrey: Not at all. I meant that `grafana_k6_project_allowed_load_zones` can set permissions for the default project, unlike the `allowed_load_zones` property of `grafana_k6_project`. The latter requires the project to be managed in terraform as a resource, which is impossible for the default project.

Leaving only the `grafana_k6_project_allowed_load_zones` resource is the way to go :+1:

pmdz: Change made to project resource reverted: https://github.com/grafana/terraform-provider-grafana/pull/2271/commits/6dbe12a20afe0cd4e5780c57e8d92646ed237217

internal/resources/k6/resource_project_allowed_load_zones.go
pmdz: done in 9a273e06420d75ec7fa1f53572f67d5ec00da92c

internal/resources/k6/data_source_k6_project.go
pmdz: data source added: d76b0160bd4cb3d9b8dab8a87888d9b0d115fcb2

internal/resources/k6/resource_project.go
yorugac: Perhaps, such things should be documented? Like "this is the approach we adopt for the design of k6 Terraform resources: ..." Something small, in this repo, but which can be referenced by anyone working on this code.

internal/resources/k6/data_source_k6_project_allowed_load_zones.go
```diff
@@ -0,0 +1,99 @@
...
+func (d *projectAllowedLoadZonesDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Retrieves allowed load zones for a k6 project.",
+		Attributes: map[string]schema.Attribute{
+			"project_id": schema.Int32Attribute{
+				Description: "The identifier of the project to retrieve allowed load zones for.",
+				Required:    true,
+			},
+			"allowed_load_zones": schema.ListAttribute{
+				Description: "List of allowed k6 load zone IDs for this project.",
```
yorugac:
```suggestion
			"allowed_load_zones": schema.ListAttribute{
				Description: "List of allowed k6 private load zones (names) for this project.",
```

I've wondered about the naming here: this object is supposed to be only for configuring private LZs, in reality. AFAIK, this is not going to change in the future either.
Changing the `allowed_load_zones` to `allowed_private_load_zones` seems a bit too verbose ATM, but perhaps, at least the description could contain clarification?
cc @pmdz @fornfrey

pmdz: I've tweaked the description a little bit in b2396268ceb79670aa9887fb1b3a5a3ead5c3369

This are not names btw., we're referencing the IDs here 
image

internal/resources/k6/resource_project.go
fornfrey: In this case, it one of the provider design principles: [Resources should represent a single API object](https://developer.hashicorp.com/terraform/plugin/best-practices/hashicorp-provider-design-principles#resources-should-represent-a-single-api-object). 

Having a k6-specific guide might be useful, but we don't have any k6-specific practices ATM.

fornfrey (approved):
Well done! LGTM

internal/resources/k6/data_source_k6_project_allowed_load_zones.go
yorugac: Thanks, Pawel!
From user's perspective, this can be confusing actually, because it's a `name` in Kubernetes resource. But in UI, we call it "ID" :smile: But you're right that this is a frontend level problem: I've brought it up before in the Slack too. 
cc @dgzlopes for visibility: just in case, we should change this everywhere.

PR #2273, opened 2025-07-24
Add Grafana k6 schedules to the provider

DefCon-007:


internal/resources/k6/data_source_k6_schedule.go
```diff
@@ -0,0 +1,187 @@
...
+	resp.Schema = schema.Schema{
+		Description: "Retrieves a k6 schedule.",
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Description: "Numeric identifier of the schedule.",
+				Required:    true,
+			},
+			"load_test_id": schema.StringAttribute{
+				Description: "The identifier of the load test to schedule.",
+				Computed:    true,
```
fornfrey: For most scenarios, using `load_test_id` to search for a schedule will likely be more convenient than using the schedule ID. Let's make the `id` `Computed` and `load_test_id` `Required`.

internal/resources/k6/resource_schedule.go
```diff
@@ -0,0 +1,450 @@
...
+				Description: "Numeric identifier of the schedule.",
+				Computed:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"load_test_id": schema.StringAttribute{
+				Description: "The identifier of the load test to schedule.",
+				Required:    true,
+			},
```
fornfrey: `load_test_id` should be immutable. Otherwise, changing it will create a schedule for another test but not remove the schedule for the old one.

```suggestion
			"load_test_id": schema.StringAttribute{
				Description: "The identifier of the load test to schedule.",
				Required:    true,
				PlanModifiers: []planmodifier.String{
					stringplanmodifier.RequiresReplace(),
				},
			},
```

internal/resources/k6/resource_schedule.go
```diff
@@ -0,0 +1,450 @@
...
+	loadTestID, err := strconv.ParseInt(plan.LoadTestID.ValueString(), 10, 32)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Error parsing load test ID",
+			"Could not parse load test ID '"+plan.LoadTestID.ValueString()+"': "+err.Error(),
+		)
+		return
+	}
+
+	// Check if a schedule already exists for this load test
```
fornfrey: The parameter parsing code can be moved before this check to avoid excessive API calls.

internal/resources/k6/resource_schedule.go
```diff
@@ -0,0 +1,450 @@
...
+	resp.Schema = schema.Schema{
+		Description: "Manages a k6 schedule for automated test execution.",
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Description: "Numeric identifier of the schedule.",
+				Computed:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
```
fornfrey, line 90: Relying on schedule IDs in the provider might lead to inconsistent behaviours 🤔 For example, a user might configure a schedule in the provider but then modify, delete, and recreate it in the UI with a different configuration. This scenario will result in Terraform suggesting to create a new resource (because the schedule with the saved ID no longer exists) and failing to create it because the load test already has a schedule. This situation can be resolved only by deleting/importing the existing schedule.

From the UX perspective, using `load_test_id` as resource `id` would be better, as the scenario described above will simply create a diff instead of requiring deleting/importing the remote resource. This is similar to how project limits work in the provider. WDYT?

internal/resources/k6/resource_schedule.go
```diff
@@ -0,0 +1,450 @@
...
+func (r *scheduleResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
+}
+
+// populateModelFromAPI populates the terraform model from the k6 API response
+func (r *scheduleResource) populateModelFromAPI(schedule *k6.ScheduleApiModel, model *scheduleResourceModel) {
+	model.ID = types.StringValue(strconv.Itoa(int(schedule.GetId())))
+	model.LoadTestID = types.StringValue(strconv.Itoa(int(schedule.GetLoadTestId())))
+	model.Starts = types.StringValue(schedule.GetStarts().Format(time.RFC3339))
+	model.Deactivated = types.BoolValue(schedule.GetDeactivated())
```
fornfrey, line 497: How can we prevent deactivated/expired schedules causing noise in the diff or when applying the changes? 

Finite schedules are transient resources that don't fit the Terraform "declarative configuration" model well. If a schedule is configured to expire, than its lifecycle is managed by GCk6 and it shouldn't be managed by Terraform at all.

I suggest we remove the expiration options from the resource and focus only on non-expiring schedules.

fornfrey: The second part of this is handling schedules that were deactivated due to errors. They should show up in the diff so that a user can reactivate them by applying the config.

internal/resources/k6/data_source_k6_schedules.go
fornfrey, line 1: What would be the use case for this data source? Can we add only `grafana_k6_schedule` to reduce the maintenance burden considering it satisfies most scenarios (also using `for_each` if necessary)?

fornfrey (changes requested):
Thanks a lot for the PR! I left a bunch of suggestions. 

I'm also not sure if expiring schedules is a good fit for the Terraform model 🤔

internal/resources/k6/data_source_k6_schedules.go
DefCon-007: I thought it might be helpful to fetch all available schedules from the cloud

fornfrey: Let's keep it for consistency with other resources, then.

internal/resources/k6/resource_schedule.go
DefCon-007: That makes sense

internal/resources/k6/data_source_k6_schedule.go
DefCon-007: Makes sense, updated it

internal/resources/k6/resource_schedule.go
DefCon-007: Yes, updated it

internal/resources/k6/resource_schedule.go
DefCon-007: A schedule that is deactivated on the cloud will show up like this now. 
Screenshot 2025-07-28 at 18 19 34

fornfrey: Thanks 👍

internal/resources/k6/resource_schedule.go
```diff
@@ -0,0 +1,545 @@
...
+						Description: "The frequency of the schedule (HOURLY, DAILY, WEEKLY, MONTHLY, YEARLY).",
+						Required:    true,
+						Validators: []validator.String{
+							stringvalidator.OneOf("HOURLY", "DAILY", "WEEKLY", "MONTHLY", "YEARLY"),
+						},
+					},
+					"interval": schema.Int32Attribute{
+						Description: "The interval between each frequency iteration (e.g., 2 = every 2 hours for HOURLY). Defaults to 1.",
+						Optional:    true,
+					},
```
fornfrey: This value defaults to `null` in the provider and to `1` in the API, which leads to the following error:
```
│ 
│ When applying changes to grafana_k6_schedule.k6_tf_demo_k6_schedule, provider "provider[\"registry.terraform.io/grafana/grafana\"].k6" produced an unexpected new value:
│ .recurrence_rule.interval: was null, but now cty.NumberIntVal(1).
```

To fix this, we should set the correct default in the provider.
```suggestion
						Optional:    true,
						Computed:    true,
						Default:      int32default.StaticInt32(1),
					},
```

internal/resources/k6/resource_schedule.go
```diff
@@ -0,0 +1,545 @@
...
+				Computed:    true,
+			},
+		},
+		Blocks: map[string]schema.Block{
+			"recurrence_rule": schema.SingleNestedBlock{
+				Description: "The schedule recurrence settings. If not specified, the test will run only once on the 'starts' date.",
+				Attributes: map[string]schema.Attribute{
+					"frequency": schema.StringAttribute{
+						Description: "The frequency of the schedule (HOURLY, DAILY, WEEKLY, MONTHLY, YEARLY).",
+						Required:    true,
```
fornfrey: With this attribute marked as required, Terraform returns the following error when creating a one-off schedule:
```
│ Error: Missing Configuration for Required Attribute
│ 
│   with grafana_k6_schedule.k6_tf_demo_k6_schedule,
│   on terraform.tf line 74, in resource "grafana_k6_schedule" "k6_tf_demo_k6_schedule":
│   74: resource "grafana_k6_schedule" "k6_tf_demo_k6_schedule" {
│ 
│ Must set a configuration value for the recurrence_rule.frequency attribute as the provider has marked it as required.
│ 
│ Refer to the provider documentation or contact the provider developers for additional information about configurable attributes that are required.
```

We must mark it as optional and move the validation into a block validator:
```go
		Blocks: map[string]schema.Block{
			"recurrence_rule": schema.SingleNestedBlock{
                                ...
				Validators: []validator.Object{
					objectvalidator.AlsoRequires(
						path.MatchRelative().AtName("frequency"),
					),
				},
                           }
```

fornfrey (changes requested):
Thanks for addressing the comments! I found 2 more issues after testing locally, but they are easy to fix.

internal/resources/k6/resource_schedule.go
DefCon-007: The frequency is marked as required in the API documentation

Screenshot 2025-08-08 at 15 09 13

fornfrey (approved):
Well done! LGTM

PR #2283, opened 2025-07-30
SCIM: Use correct attribute to dllow/deny non-provisioned users

patrickeasters:
The [app platform CRD](https://github.com/grafana/grafana-enterprise/blob/dca28996b71bf5bd5f0e726319ba1a13f2e7394f/src/pkg/extensions/apis/scim/v0alpha1/types.go#L34) for `ScIMConfig` defines an attribute of `rejectNonProvisionedUsers`, while the Terraform provider attempts to set `allowNonProvisionedUsers` instead, resulting in the setting having no effect.

This PR updates the provider to use the correct attribute.


colin-stuart (approved):
Thanks for this! I went ahead and removed all references to `allow_non_provisioned_users` since we changed this SCIM setting to `reject_non_provisioned_users` to avoid any confusion

PR #2285, opened 2025-07-30
When multiple organization users found, filter results for an exact match on email or login

colin-stuart:
Part of https://github.com/grafana/support-escalations/issues/17635

### Why?
Querying the grafana_organization_user resource from the terraform provider gets all users within the current organization in Grafana. The `SearchOrgUsers` sql query uses the `like` operator, so the query is `WHERE email LIKE '%[test.name@org.com](mailto:test.name@org.com)%'` and the result can include multiple users. Currently, if multiple users are returned in the query, the TF provider returns an error.

### Fix

The fix relies on switching from `GetOrgUsersForCurrentOrgLookup` over to `GetOrgUsersForCurrentOrg` which returns more information, including the `email`. That allows us to then tweak the logic to perform an exact match on `email` or `login`.

- If `login` is provided: the TF provider iterates through the results of the query to return a single user if an exact match on `user.login` is found.
- if `email` provided: The TF provider now iterates through the results of the query to return a single user if an exact match on `user.email` is found.


internal/resources/grafana/data_source_organization_user_test.go
```diff
@@ -38,3 +38,111 @@ func TestAccDatasourceOrganizationUser_basic(t *testing.T) {
...
+			{
+				Config: testAccDatasourceOrganizationUserExactMatch,
+				Check:  resource.ComposeTestCheckFunc(checks...),
+			},
+		},
+	})
+}
+
+// TestDataSourceOrganizationUserExactMatchLogic tests the exact matching logic without requiring a Grafana instance
+func TestDataSourceOrganizationUserExactMatchLogic(t *testing.T) {
```
volcanonoodle: If we want to test the exact-matching logic, would it make sense to extract the logic from `internal/resources/grafana/data_source_organization_user.go` into a function that we can use there as well as in this test? 

Just wondering if otherwise this could be a bit brittle, having to keep the exact-match login from this test in-sync with the one in `internal/resources/grafana/data_source_organization_user.go` manually.

volcanonoodle (approved):
LGTM! 😄

cinaglia:
FYI Looks like the acceptance tests are failing for an unrelated issue. I've spot checked other recent PRs that have the same issue.

internal/resources/grafana/data_source_organization_user.go
```diff
@@ -45,34 +45,46 @@ func dataSourceOrganizationUserRead(ctx context.Context, d *schema.ResourceData,
...
+		if matchBy(user, emailOrLogin) {
+			d.Set("user_id", user.UserID)
+			d.Set("login", user.Login)
+			d.Set("email", user.Email)
+			d.SetId(MakeOrgResourceID(orgID, user.UserID))
+			return nil
+		}
+	}
+
+	return diag.Errorf("ambiguous query when reading organization user, multiple users returned by query: %q", emailOrLogin)
```
volcanonoodle: We would only reach this point if the API returned 1 or more users but they didn't match the email or login right? If so, would it make sense changing this error message to something that expresses that there aren't any users matching the provided `login` or `email`?

volcanonoodle (approved):
Great 💯

internal/resources/grafana/data_source_organization_user.go
cinaglia: That's a good idea. This endpoint is paginated, so it will also hit this scenario when there are multiple pages returned (even if a subsequent page might include an exact match.

cinaglia: Pushed [08570fd](https://github.com/grafana/terraform-provider-grafana/pull/2285/commits/08570fd4622afb81b7469352dfe07097de13e8f4). WDYT?

colin-stuart (commented):
I can't approve @cinaglia's changes since this is my own PR, but the changes look good.

internal/resources/grafana/data_source_organization_user.go
volcanonoodle: Oh I forgot about pagination. https://github.com/grafana/terraform-provider-grafana/pull/2285/commits/08570fd4622afb81b7469352dfe07097de13e8f4 handles it very nicely I think. Thanks!

volcanonoodle (approved):

cinaglia (approved):

PR #2289, opened 2025-08-05
feat: Report Terraform client and version to SM API

### LINKED ISSUE #2282: Start reporting `X-Client-ID`/`X-Client-Version` for Synthetics
Start identifying the Terraform provider/version when issuing requests to the Synthetic Monitoring API to help evaluate usage and support requirements.

Depends on https://github.com/grafana/synthetic-monitoring-api-go-client/issues/317 to close first so the `smapi.Client` header values can be overridden.

The-9880:
Closes https://github.com/grafana/terraform-provider-grafana/issues/2282

Uses the  `X-Client-ID` and `X-Client-Version` HTTP headers to report the Terraform client and version to the SM API.

* Updates synthetic-monitoring-agent to v0.39.2
* Updates synthetic-monitoring-api-go-client to v0.15.1
* Updates github.com/prometheus/common to v0.65.0
* Updates github.com/urface/cli/v2 to v2.27.7
* Updates golang.org/x/exp to v0.0.0-20250718183923-645b1fa84792
* Updates golang.org/x/text to v0.27.0


The-9880:
I can see the logs getting reported - our API is registering `client_type` as `terraform`: [logs](https://ops.grafana-ops.net/explore?schemaVersion=1&panes=%7B%22bd9%22:%7B%22datasource%22:%22000000193%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bjob%3D~%5C%22synthetic-monitoring.%2A%2Fsynthetic-monitoring-api%5C%22%7D%20%7C%20json%20%7C%20client_type%21%3D%5C%22unknown%5C%22%20%7C%20client_type%20%21%3D%20%5C%22%5C%22%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22000000193%22%7D,%22editorMode%22:%22code%22,%22direction%22:%22backward%22%7D%5D,%22range%22:%7B%22from%22:%22now-2d%22,%22to%22:%22now%22%7D,%22compact%22:false%7D%7D&orgId=1).

The `client_version` in these logs is `testacc`, but that is because this is being [hardcoded](https://github.com/grafana/terraform-provider-grafana/blob/07783cfeb8d56a7be01be698698b359ec9ddbd24/internal/testutils/provider.go#L29) for the CI tests. We expect this to be set to a semver string in release versions (see [here](https://github.com/grafana/terraform-provider-grafana/blob/07783cfeb8d56a7be01be698698b359ec9ddbd24/main.go#L20)).

The-9880:
^ Had to rebase because something merged in

internal/resources/cloud/resource_synthetic_monitoring_installation.go
```diff
@@ -20,6 +20,14 @@ var smAPIURLsExceptions = map[string]string{
 	"us-azure":        "https://synthetic-monitoring-api-us-central2.grafana.net",
 }
 
+// createSMClient creates a new SMAPI client with proper client-id and client-version settings
+func createSMClient(apiURL, accessToken string) *SMAPI.Client {
+	client := SMAPI.NewClient(apiURL, accessToken, nil)
+	client.SetCustomClientID("terraform")
+	client.SetCustomClientVersion("unknown") // TODO: see if we can get the provider version here
```
Pokom, line 27: That is a bit unfortunate. I was trying to trace through the code and see if we could rely upon the common client created in the provider. 

What's the implications of this SMClient having an unknown version, but the base provider having it set? I see some logs where version _does_ come through 🤔

Pokom (approved):
One comment, I was hoping to see if there was an easy path to getting the version, but I think we'll need to rework a bit of the code.

internal/resources/cloud/resource_synthetic_monitoring_installation.go
The-9880: This one is just for installation, so at least it doesn't have any bearing on tracking which specific SM resources (e.g. checks) are being managed by TF. It'll still let us know which tenants were initialized via TF, though.

For any (org, stack) combination, this endpoint should only be called when installing SM for the first time and perhaps again in the future if the tenant was disabled and they want to re-enable it.

Pokom: Oh okay, yeah that's such a limited case that I don't think it's worth the effort required. Thanks!

PR #2284, opened 2025-07-30
feat: support oncall webhook presets

matiasb:
Related to https://github.com/grafana/irm/issues/6185

Requires oncall engine public API release including this [PR](https://github.com/grafana/irm/pull/6580) and also missing the corresponding amixr-api-go-client [update](https://github.com/grafana/amixr-api-go-client/pull/34).

e.g.
```
resource "grafana_oncall_outgoing_webhook" "simple" {
  provider = grafana.oncall
  name = "Testing simple"
  preset= "simple_webhook"
  url = "https://example.com"
}
```
or
```
resource "grafana_oncall_outgoing_webhook" "sift" {
  provider = grafana.oncall
  name = "Testing Sift"
  preset = "grafana_sift"
  trigger_type = "acknowledge"
}
```
or
```
resource "grafana_oncall_outgoing_webhook" "testing_incident" {
  provider = grafana.oncall
  name = "Testing Incident from TF"
  preset = "incident_webhook"
  trigger_type = "incident declared"
  url = "https://example.com"
  http_method = "POST"
  data = "{\"title\": \"{{ incident.title }}\"}"
}
```

PR #2291, opened 2025-08-06
docs: add simple example for IRM incident webhooks

matiasb:
Related to https://github.com/grafana/terraform-provider-grafana/pull/2284

Show an additional example in the webhooks docs for an incident webhook.


willgallego-grafana (approved):
👍

PR #2315, opened 2025-08-20
feat: make plugin version optional - installs latest version

### LINKED ISSUE #2312: Make plugin version optional for cloud_plugin_install
### Feature Request

https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/cloud_plugin_installation

Given running the latest version is recommended and in cloud we are increasingly automating this, it would be beneficial to support this better within the terraform provider. 

To start with, we should make the version optional

s4kh:
Fixes https://github.com/grafana/terraform-provider-grafana/issues/2312
Tested with dev instance on local using locally built provider.
**resource.tf**
```
terraform {
  required_providers {
    grafana = {
      source = "registry.terraform.io/grafana/grafana"
    }
  }
}
provider "grafana" {
  cloud_api_url             = "https://grafana-dev.com/api"
  cloud_access_policy_token = "dev token"
}
resource "grafana_cloud_plugin_installation" "test" {
  stack_slug = "skh3"
  slug       = "axiomhq-axiom-datasource"
}
```

### DIFF
--- a/internal/resources/cloud/resource_cloud_plugin_installation.go
+++ b/internal/resources/cloud/resource_cloud_plugin_installation.go
@@ -46,9 +46,9 @@ Required access policy scopes:
 				ForceNew:    true,
 			},
 			"version": {
-				Description: "Version of the plugin to be installed.",
+				Description: "Version of the plugin to be installed, latest version is installed when omitted.",
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
 				ForceNew:    true,
 			},
 		},
@@ -92,10 +92,14 @@ func listStackPlugins(ctx context.Context, client *gcom.APIClient, data *ListerD
 func resourcePluginInstallationCreate(ctx context.Context, d *schema.ResourceData, client *gcom.APIClient) diag.Diagnostics {
 	stackSlug := d.Get("stack_slug").(string)
 	pluginSlug := d.Get("slug").(string)
+	version := "latest"
+	if v, ok := d.GetOk("version"); ok {
+		version = v.(string)
+	}
 
 	req := gcom.PostInstancePluginsRequest{
 		Plugin:  pluginSlug,
-		Version: common.Ref(d.Get("version").(string)),
+		Version: common.Ref(version),
 	}
 
 	err := retry.RetryContext(ctx, 2*time.Minute, func() *retry.RetryError {
@@ -136,7 +140,9 @@ func resourcePluginInstallationRead(ctx context.Context, d *schema.ResourceData,
 
 	d.Set("stack_slug", installation.InstanceSlug)
 	d.Set("slug", installation.PluginSlug)
-	d.Set("version", installation.Version)
+	if _, ok := d.GetOk("version"); ok {
+		d.Set("version", installation.Version)
+	}
 	d.SetId(resourcePluginInstallationID.Make(stackSlug, pluginSlug))
 
 	return nil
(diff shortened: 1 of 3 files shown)


s4kh:
cloud acceptance tests

DanCech:
I kind of wonder whether it would make sense to support `latest` as an explicit version vs making the attribute optional, then we could potentially check against the plugin catalog in Read whether the installed version is the latest, if so report `latest` so there's no drift but if not then report the actual installed version to trigger an update.

s4kh:
@DanCech yeah.
But do we need to auto update the plugin?
There is a following scenario:
1. Plugin is installed without the version specified.
    a. We don't return installed version in the read context(current state) function - if we do, terraform will try to delete and create again due to installed version != "" (declared in the resource.tf) **loop happens**
3. New plugin version is released
4. We run `terraform plan`  
    a. It will show nothing to apply - because we don't show version in the readContext function because of **1a.**
6. What do we want in this scenario?

If we prefer stability and don't want to auto update, current implementation is fine. But in order to install the latest after the release, user either has to define the version number in the resource.tf

DanCech:
yeah, I'm thinking that if the user asks for the latest we should keep it up to date, and if they want a specific version they should specify it. Hence the idea of using an explicit `version: latest` in the spec and having read return `version: latest` if the installed version is the latest, but `version: <installed version>` if it's not the latest, which would trigger it to update to `latest`.

DanCech:
vs asking for latest and getting whatever it happens to be at install time but then never updating.

internal/resources/cloud/resource_cloud_plugin_installation.go
```diff
@@ -89,10 +92,14 @@ func listStackPlugins(ctx context.Context, client *gcom.APIClient, data *ListerD
 func resourcePluginInstallationCreate(ctx context.Context, d *schema.ResourceData, client *gcom.APIClient) diag.Diagnostics {
 	stackSlug := d.Get("stack_slug").(string)
 	pluginSlug := d.Get("slug").(string)
+	version := "latest"
```
DanCech: should use `LatestVersion` here

internal/resources/cloud/resource_cloud_plugin_installation.go
```diff
@@ -117,10 +124,19 @@ func resourcePluginInstallationRead(ctx context.Context, d *schema.ResourceData,
 	if err, shouldReturn := common.CheckReadError("plugin", d, err); shouldReturn {
 		return err
 	}
+	catalogPlugin, _, err := client.PluginsAPI.GetPlugin(ctx, pluginSlug.(string)).Execute()
+	if err, shouldReturn := common.CheckReadError("plugin", d, err); shouldReturn {
+		return err
+	}
```
DanCech: Do we want to restructure slightly so this only runs when the desired version is latest?

s4kh: yup

internal/resources/cloud/resource_cloud_plugin_installation.go
```diff
@@ -92,10 +95,11 @@ func listStackPlugins(ctx context.Context, client *gcom.APIClient, data *ListerD
 func resourcePluginInstallationCreate(ctx context.Context, d *schema.ResourceData, client *gcom.APIClient) diag.Diagnostics {
 	stackSlug := d.Get("stack_slug").(string)
 	pluginSlug := d.Get("slug").(string)
+	version := d.Get("version").(string)
```
s4kh, line 98: defaults to latest

DanCech (approved):
LGTM, I don't have particular religion about whether the version param is optional with latest as the default or required.

PR #2317, opened 2025-08-20
SLO: Support Custom UUIDs in Terraform Provider

### LINKED ISSUE #2316: SLO: Support custom UUIDs in the Terraform Provider
The SLO API allows a user to specify a custom vanity UUID on POST
Enable this functionality in the Terraform Provider, which would allow users to create SLOs with vanity UUIDs

elainevuong:
Adds support for setting custom UUIDs on the Terraform resources for SLO
This behaviour is already enabled in the API

Fixes #2316


internal/resources/slo/resource_slo.go
```diff
@@ -59,6 +59,13 @@ Resource manages Grafana SLOs.
 				Optional:    true,
 				Description: `UID for the SLO folder`,
 			},
+			"uuid": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
```
ellisda, line 65: question: are we using this combination of "computed=true, optional=true" elsewhere in SLO or grafana?

From a GPT search it seems this combination does what we want

> Key Characteristics of Computed fields:
> 
> Provider-Generated: The value is not known until the resource is created or updated by the provider.
> 
> Read-Only in Configuration: Users typically do not set values for computed attributes in their .tf files. If they do, the provider will overwrite it with the computed value.
> 
> Can be Optional: An attribute can be both Optional and Computed. This allows a user to optionally provide a value. If a value is provided, it will be used; if not, the provider will compute a value.

internal/resources/slo/resource_slo.go
```diff
@@ -59,6 +59,13 @@ Resource manages Grafana SLOs.
...
+				ForceNew:    true,
```
ellisda: question: does this ForceNew help the user experience or hurt it?

We don't intend to allow users to change the UUID of a SLO once it's created. Without this flag, a user who tried to change the UUID would get a 5xx error from our PUT endpoint saying changing the UUID isn't supported. How does the terraform user's experience change with this flag? Making the create/destroy cycle too easy isn't ideal either, since they lose history with the delete that the create will not inherit.

internal/resources/slo/resource_slo.go
```diff
@@ -414,8 +422,14 @@ func packSloResource(d *schema.ResourceData) (slo.SloV00Slo, error) {
 		tflabels = packLabels(labels)
 	}
 
+	// Use vanity UUID if set, otherwise use an empty string - will be replaced by the API with a random UUID
+	tfUUID := ""
+	if vanityUUID, ok := d.GetOk("uuid"); ok && vanityUUID.(string) != "" {
+		tfUUID = vanityUUID.(string)
+	}
```
ellisda: question: is "custom UUID" a better phrase to enshrine in the open source code here?

I know we call them vanity UUIDs, but maybe custom is simpler/better for people outside our team

ellisda (commented):
LGTM (but I'm not qualified here to approve proper) - I left a couple of questions

internal/resources/slo/resource_slo.go
elainevuong: yep - `computed: true, optional: true` is used in Grafana, and you can see the same use case - the UID is automatically generated unless it is specified during the panel creation. You can see it in here in the `resourceLibraryPanel` in our Grafana Terraform Provider

```
func resourceLibraryPanel() *common.Resource {
	schema := &schema.Resource{

		Description: Manages Grafana library panels.
		...
			"org_id": orgIDAttribute(),
			"uid": {
				Type:     schema.TypeString,
				Computed: true,
				Optional: true,
				Description: "The unique identifier (UID) of a library panel uniquely identifies library panels between multiple Grafana installs. " +
					"It’s automatically generated unless you specify it during library panel creation." +
					"The UID provides consistent URLs for accessing library panels and when syncing library panels between multiple Grafana installs.",
			},
```

For your comment below - I think the [ForceNew](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#forcenew) could potentially hurt the customer. `ForceNew` means that if the field is modified, the resource will be destroyed and re-created. Our API already has a preventative mechanism for this - we don't allow PUTs to change the SLO UUIDs, so I agree with your thoughts and the ForceNew isn't needed here. Will remove, thanks! 👍

matt-metivier (approved):

PR #2330, opened 2025-09-02
Alerting: Contact point schema based packing/unpacking

JacobsonMT:
### Changes
- Improves test coverage for packing and unpacking of alerting contact point resources.  
- Refactors notifier-specific `pack` and `unpack` methods to use shared `packNotifier` and `unpackNotifier` functions.  
- These functions use the existing Terraform schema to automatically handle all fields, ~including nested hierarchies~ and sensitive fields, removing the need for redundant `secureFields` and manual field enumeration.  

### Why
Previously, packing and unpacking was done manually for both sensitive and non-sensitive fields. Sensitive fields had to be listed in `secureFields`, while non-sensitive fields were individually mapped in `pack` and `unpack` methods.  

This created duplication with the schema definition and made every schema change require edits in multiple places:  
1. Update the `schema` method.  
2. Update `pack` to store the field in Terraform state.  
3. Update `unpack` to send the field to the Grafana API.  
4. Update `secureFields` if the field was sensitive.  

Because of this, the code was harder to maintain and the actual logic that mattered for each notifier, such as key or value transformations, was buried in boilerplate.  

For example:  
- In the `pushover` notifier, the TF field `ok_priority` maps to `okPriority` in the Grafana API. The value is stored as an integer in Terraform but sent as a string to the API. This important detail was easy to miss among all the repetitive field mappings.  

~The old approach also did not support deeply nested fields/secrets, which we need for upcoming work (ex. `http_config`).~ **NOTE**: Extracted support for packing/unpacking nested schemas from this PR to make the noop refactor easier to review, will add it back with the PR for `http_config` and `payload` webhook fields.

The new approach:
- Uses the schema as the single source of truth for packing and unpacking.  
- Automatically handles sensitive fields without needing a separate `secureFields` list.  
- Makes notifier-specific code smaller and clearer by keeping only the field transformations that actually differ.  
- Reduces duplication and makes schema changes easier and less error-prone.
- Supports both sensitive and non-sensitive deeply nested fields.

### Reviewer notes
Tested to ensure TF state and Grafana state remains consistent before and after the refactor:  
- Existing Terraform state remains unchanged after update and destroy/apply.
- Existing Grafana state remains unchanged after update and destroy/apply.

This was done with the existing "fully defined contact points" as well as the newly added "minimal contact points" to ensure defaults/omissions were unaffected.

### DIFF
--- a/internal/resources/grafana/resource_alerting_contact_point.go
+++ b/internal/resources/grafana/resource_alerting_contact_point.go
@@ -410,6 +410,246 @@ func commonNotifierResource() *schema.Resource {
 	}
 }
 
+// fieldMapper is a helper struct to map fields that differ between Terraform and Grafana schema. Such as field keys or type conversions.
+type fieldMapper struct {
+	newKey        string
+	packValFunc   func(any) any
+	unpackValFunc func(any) any
+}
+
+func newFieldMapper(newKey string, packValFunc, unpackValFunc func(any) any) fieldMapper {
+	return fieldMapper{
+		newKey:        newKey,
+		packValFunc:   packValFunc,
+		unpackValFunc: unpackValFunc,
+	}
+}
+
+// newKeyMapper is a fieldMapper that only changes the key name in the schema.
+func newKeyMapper(newKey string) fieldMapper {
+	return fieldMapper{
+		newKey: newKey,
+	}
+}
+
+// valueAsInt is a fieldMapper function that converts a value to an integer.
+func valueAsInt(value any) any {
+	switch typ := value.(type) {
+	case int:
+		return typ
+	case float64:
+		return int(typ)
+	case string:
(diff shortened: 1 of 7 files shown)


internal/resources/grafana/resource_alerting_contact_point.go
```diff
@@ -282,7 +282,7 @@ func unpackContactPoints(data *schema.ResourceData) []statePair {
 			// If it's not deleted, it will either be created or updated
 			result = append(result, statePair{
 				tfState: pointMap,
-				gfState: unpackPointConfig(n, p, name),
+				gfState: unpackNotifier(p.(map[string]any), name, n),
```
yuri-tceretian: nit: why don't we use `pointMap` here instead?

yuri-tceretian (approved):
LGTM

PR #2333, opened 2025-09-05
Add ID field to app platform resources

alexander-akhmetov:
Adding a computed ID field to the app platform base Resource model. This field is [required](https://developer.hashicorp.com/terraform/plugin/framework/acctests#no-id-found-in-attributes) with by the current testing framework version for acceptance tests.

The field is computed and should not result in a plan change. Tested locally:

1. Created a new `grafana_apps_playlist_playlist_v0alpha1` resource
2. Applied it with the old provider version and checked the state: no id field

```
terraform state show 'grafana_apps_playlist_playlist_v0alpha1.test_playlist'

# grafana_apps_playlist_playlist_v0alpha1.test_playlist:
resource "grafana_apps_playlist_playlist_v0alpha1" "test_playlist" {
    metadata {
        uid     = "tf_test_playlist"
        uuid    = "D2tQoZxmFNLHZ55sHDzaooxGV97VxtexXNl9nDE2E8sX"
        version = "1757057965719"
    }
    options {
        overwrite = true
    }
    spec {
        interval = "1h"
        items    = [
            {
                type  = "dashboard_by_uid"
                value = "ed155665"
            },
        ]
        title    = "[tf] Test Playlist"
    }
}
```

3. Updated the provider version and ran the plan again:

```
terraform plan

...
No changes. Your infrastructure matches the configuration.

Terraform has compared your real infrastructure against your configuration and found no differences, so no changes are needed.
```

### DIFF
--- a/internal/resources/appplatform/resource.go
+++ b/internal/resources/appplatform/resource.go
@@ -24,6 +24,7 @@ import (
 
 // ResourceModel is a Terraform model for a Grafana resource.
 type ResourceModel struct {
+	ID       types.String `tfsdk:"id"`
 	Metadata types.Object `tfsdk:"metadata"`
 	Spec     types.Object `tfsdk:"spec"`
 	Options  types.Object `tfsdk:"options"`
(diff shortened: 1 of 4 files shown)


internal/resources/appplatform/resource.go
```diff
@@ -510,6 +520,8 @@ func SaveResourceToModel[T sdkresource.Object](
 		}
 	}
 
+	dst.ID = meta.UID
```
radiohead: Since `meta.uid` is not unique across time (e.g. an object X with `uid = "foo"` can be created and then deleted, after which object Y can be again created with the same `uid), I think it's best to use `UUID`, which is unique across space & time:
```suggestion
	dst.ID = meta.UUID
```

radiohead (commented):
Looks great! Just a small note about using `UUID` instead of `UID`.

internal/resources/appplatform/resource.go
alexander-akhmetov: Updated!

radiohead (approved):

PR #2338, opened 2025-09-05
4658:add custom model rules resource

matt-metivier:
This PR introduces a new Terraform resource for managing Asserts custom model rules, enabling users to define custom entity types and their detection rules using YAML configuration.

### Changes

**New Resource: `grafana_asserts_custom_model_rules`**
- Manages custom entity definitions through the Asserts CustomModelRulesControllerAPI
- Supports flexible entity type configuration with Prometheus query-based detection
- Implements full CRUD operations with proper retry logic for eventual consistency
- Uses name-based resource identification for individual rule set management

**Key Features**
- Entity type definition with custom names and types (Service, Pod, Namespace, etc.)
- Prometheus query-based entity detection rules
- Nested schema structure supporting multiple entities per rule set
- Error handling and retry mechanisms

### Technical Details

The resource implements proper retry logic using shared utilities to handle eventual consistency issues common with distributed systems. The implementation includes conversion functions between Terraform schema and API DTOs, ensuring proper data transformation and validation.

### Testing

- Unit tests validate all CRUD operations and data conversion
- Acceptance tests verify resource lifecycle and state management
- Stress tests confirm retry logic handles concurrent operations
- Import functionality verified with proper state handling

### Breaking Changes

None. This is a new resource addition with no impact on existing functionality.


docs/resources/asserts_custom_model_rules.md
```diff
@@ -0,0 +1,46 @@
...
+
+```terraform
+resource "grafana_asserts_custom_model_rules" "test" {
+  name  = "test-anything"
+  rules = <<-EOT
+    entities:
+    - type: Whatever
+      name: Nothing
+      definedBy:
+      - query: "up{job=\\"nothing\\"}"
```
Alex3k: Would it make sense to make this more Terraform native by creating it like 
```
resource "grafana_asserts_custom_model_rules" "test" {
  name  = "test-anything"
  rules {
    entity {
      type = "whatever"
      name = "foo"
      definedBy {
        query = "up{job=\\"nothing\\"}"
      }

      definedBy {
        query = "up{job=\\"nothing\\"}"
      }
    }
  }
}
```

I feel this follows other conventions like in the k8s provider when defining aspects like environment variables as an example. 

Yaml provider a lot of opportunity for error

matt-metivier: Thanks for the suggestion, I updated the string

docs/resources/asserts_custom_model_rules.md
```diff
@@ -0,0 +1,108 @@
...
+<a id="nestedblock--rules--entity--defined_by"></a>
+### Nested Schema for `rules.entity.defined_by`
+
+Required:
+
+- `query` (String) The Prometheus query that defines this entity.
+
+Optional:
+
+- `disabled` (Boolean) Whether this query is disabled.
```
egovova11: I think it should say something like `whether is this rule is disabled`. Specifically, when `disabled` flag is set to `true`, then only the `query` field is used to match a rule to disable. 

This may save some time for people sifting though the docs.

egovova11 (approved):

docs/resources/asserts_custom_model_rules.md
matt-metivier: added a better description on it

docs/resources/asserts_custom_model_rules.md
```diff
@@ -0,0 +1,108 @@
...
+Required:
+
+- `defined_by` (Block List, Min: 1) List of queries that define this entity. (see [below for nested schema](#nestedblock--rules--entity--defined_by))
+- `name` (String) The name of the entity.
+- `type` (String) The type of the entity (e.g., Service, Pod, Namespace).
+
+Optional:
+
+- `disabled` (Boolean) Whether this entity is disabled.
+- `enriched_by` (List of String) List of enrichment sources for the entity.
```
jgehrett, line 84: `enriched_by` has the same schema as `defined_by`.
It's a [`PropertyRule`](https://github.com/grafana/asserts-adi/blob/master/model/rules/src/main/java/ai/asserts/model/rules/PropertyRule.java)

internal/resources/asserts/resource_custom_model_rules.go
```diff
@@ -0,0 +1,478 @@
...
+									"lookup": {
+										Type:        schema.TypeMap,
+										Optional:    true,
+										Description: "Lookup mappings for the entity.",
+										Elem:        &schema.Schema{Type: schema.TypeString},
+									},
+									"enriched_by": {
+										Type:        schema.TypeList,
+										Optional:    true,
+										Description: "List of enrichment sources for the entity.",
```
jgehrett, line 297: I think this should be similar to `defined_by`.
Something like:
`List of queries that enrich entities of this type.`

jgehrett (commented):
Looking good! Sorry I'm late to the party. Just a small comment on `enriched_by` doc

Issue #2176, opened 2025-05-16
Failure to realize provider succeeded in creating a folder


jimmy-outschool:
### Terraform Version

1.5.4

### Terraform Grafana Provider Version

3.25.0

### Grafana Version

11.4.0

### Affected Resource(s)

* grafana_folder

### Terraform Configuration Files

```hcl
resource "grafana_folder" "the_one" {
  count = 100
  title = "The One: ${count.index}"
}
```

### Expected Behavior

100 folders created

### Actual Behavior

Between 85-90 folders created successfully with the rest resulting in:

```
Error: failed to create folder: [POST /folders][409] createFolderConflict {"message":"a folder with the same name already exists in the current location"}
```

While they were indeed created and do exist.

### Steps to Reproduce

1. `terraform apply`

### References

Potentially related #1910 .

jimmy-outschool:
This happens intermittently with dynamic stacks that create 1-3 folders.

jimmy-outschool:
Setting:

```hcl
uid = "the_one_${count.index}"
```

Changes the error to:

```
Error: failed to create folder: [POST /folders] createFolder (status 412): {}
```

but same ratio.

mustafasencer:
I had a hard time replicating the issue locally and can confirm that this is happening in an intermittent fashion. The `412` error codes point to an already existing folder resource or that the target folder resource has been updated by a non-automation actor. When those cases are not present, the issue happen on a very rarely fashion. Anyway, the root cause behind its occurrence intermittently while creating folders in rapid succession is so far not clear to me. 

In order to mitigate the issue, we have introduced `412` as a default retry status on the client side. After introducing a retry logic, I haven't been able to replicate/reproduce the issue when creating folders in rapid succession.

jimmy-outschool:
Thanks for digging into this. So after a `412` Grafana will produce a success? We always have to go and delete the folder in order to get a successful deploy after this happens. As such either the provider would need to treat the response as a success if a follow-up indicates the folder exists or import it of sorts. Otherwise, I am not clear on why retrying in the provider would be different than rerunning the apply.

jimmy-outschool:
I am also curious if your issues with reproducing indicate there are factors on the grafana side that can be adjusted.

mustafasencer:
[...]

PR #2341, opened 2025-09-09
fix: extend default retry status codes with 412

mustafasencer:
The issue surfaces intermittently when creating folders in rapid succession. The Grafana server returns `412` response code for some of the folders. In this PR, we extend the default retry HTTP status codes with the new status code encountered. This approach permanently alleviates the problem on the terraform provider side.

**Fixes:** https://github.com/grafana/support-escalations/issues/18187 & https://github.com/grafana/terraform-provider-grafana/issues/2176

**Note:** I haven't been able to pinpoint the issue on the grafana side. I am still looking into it. Most importantly checking where this weird race condition is being triggered.

### DIFF
--- a/internal/resources/grafana/resource_folder_test.go
+++ b/internal/resources/grafana/resource_folder_test.go
@@ -367,6 +367,28 @@ func TestAccFolder_PreventDeletionNested(t *testing.T) {
 	})
 }
 
+func TestAccFolder_RapidCreation(t *testing.T) {
+	testutils.CheckOSSTestsEnabled(t)
+
+	name := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)
+	folderCount := 100
+
+	var checks []resource.TestCheckFunc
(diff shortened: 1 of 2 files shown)


mustafasencer:
cloud acceptance tests ran here: https://github.com/grafana/terraform-provider-grafana/actions/runs/17577317978

internal/resources/grafana/resource_folder_test.go
```diff
@@ -475,3 +497,13 @@ func testAccFolderExample_PreventDeletion(name string, preventDeletion bool) str
 		}
 	`, name, preventDeletionStr)
 }
+
+func testAccFolderExample_RapidCreation(count int) string {
```
pstibrany: nit: given the single usage of the function, perhaps this could be moved into the `TestAccFolder_RapidCreation` function?

pstibrany (approved):
lgtm

PR #2353, opened 2025-09-15
4665 get log config v2 asserts

matt-metivier:


matt-metivier:
closed since it is better to just combine the Crud into 2 prs for easy of testing

PR #2358, opened 2025-09-16
Update grafana_k6_installation docs

fornfrey:
Add missing cloud access policy scope in grafana_k6_installation docs


docs/resources/k6_installation.md
```diff
@@ -28,6 +28,7 @@ Required access policy scopes:
 * stacks:write
 * subscriptions:read
 * orgs:read
+* stack-service-accounts:write
```
d14c, line 31: what is this scope needed for?

d14c (approved):

docs/resources/k6_installation.md
fornfrey: It is required for creating a serivce account token that installs the GCk6 app in the stack (step 2 from the example): https://github.com/grafana/terraform-provider-grafana/blob/a8cd7114fa2763fc22a984a5a200ba013e311e97/docs/resources/k6_installation.md?plain=1#L58

PR #2359, opened 2025-09-18
fix fe-o11y: use regionSlug instead of clusterSlug

t00mas:


internal/resources/frontendo11y/resources.go
```diff
@@ -19,8 +19,9 @@ var Resources = []*common.Resource{
 	makeResourceFrontendO11yApp(),
 }
 
-func apiURLForCluster(cluster string, host string) string {
-	return fmt.Sprintf("https://faro-api-%s.%s/faro", cluster, host)
+// apiURLForRegion constructs the API URL for a given region slug and a host.
+func apiURLForRegion(regionSlug string, host string) string {
+	return fmt.Sprintf("https://faro-api-%s.%s/faro", regionSlug, host)
```
mar4uk: I'm not sure we can use regionSlug instead of clusterSlug. They are not always the same. For example,
cluster us-centra2 has regionSlug us-azure, but clusterSlug is us-central2, the api url will change from `https://faro-api-us-central2.grafana.net/faro` to `https://faro-api-us-azure.grafana.net/faro`

t00mas: You're correct - I'll try something else
We cannot use clusterSlug anyway because some clusters' faroEndpointUrl doesn't match to its clusterSlug hence this change. We'll have to find the faroEndpointUrl attr directly.

rlankfo: Here we need to make another call to gcom /stack-regions endpoint with the region slug to get the right faro endpoint.

mar4uk (approved):

rlankfo (approved):

PR #2360, opened 2025-09-18
Alerting: Add provenance support to alert enrichment resource

alexander-akhmetov:
Add the `disable_provenance` attribute to the `grafana_apps_alertenrichment_alertenrichment_v1beta1` resource, similar to the `grafana_rule_group`:
- `disable_provenance = False` (default): applies the `grafana.com/provenance` = `api` annotation to the resource, marking it as read-only in the UI.
- `disable_provenance = True`: sets an empty `grafana.com/provenance` annotation.

### DIFF
--- a/internal/resources/appplatform/alertenrichment_resource.go
+++ b/internal/resources/appplatform/alertenrichment_resource.go
@@ -1229,7 +1229,7 @@ func parseAlertEnrichmentSpec(ctx context.Context, src types.Object, dst *v1beta
 
 	meta, err := utils.MetaAccessor(dst)
 	if err == nil {
-		if !data.DisableProvenance.IsNull() && data.DisableProvenance.ValueBool() {
+		if data.DisableProvenance.ValueBool() {
 			if meta.GetAnnotations() == nil {
 				meta.SetAnnotations(map[string]string{})
 			}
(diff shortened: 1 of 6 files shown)


internal/resources/appplatform/alertenrichment_resource.go
```diff
@@ -1214,6 +1227,20 @@ func parseAlertEnrichmentSpec(ctx context.Context, src types.Object, dst *v1beta
 		}
 	}
 
+	meta, err := utils.MetaAccessor(dst)
+	if err == nil {
+		if !data.DisableProvenance.IsNull() && data.DisableProvenance.ValueBool() {
```
laurenashleigh: I'm trying to understand the logic here, is `!data.DisableProvenance.IsNull()` a double negative or it's Go's way of checking DisableProvenance is not null?

alexander-akhmetov: It's the `types.Bool` from the terraform-plugin-framework. I think it can probably be replaced with just `data.DisableProvenance.ValueBool()`, I'll check

alexander-akhmetov: `ValueBool` returns false when the value is null, so we can just replace that. Good catch! https://github.com/grafana/terraform-provider-grafana/pull/2360/commits/0d64aa4fa159097c343d5b360b26d7016b6d840b

fayzal-g (approved):

PR #2363, opened 2025-09-22
Add k6bundle provider function

dgzlopes:
GCk6 and SM resources require users to leverage an external bundler to configure multi-file k6 tests.

Sometimes, users end up configuring it like this:
```hcl
resource "null_resource" "bundle_backend_test" {
  provisioner "local-exec" {
    command = "npx esbuild ../src/backend/get_recommendation.ts --bundle --outfile=../dist/backend/get_recommendation.js --platform=node --external:k6 --external:https*"
  }
}

resource "grafana_synthetic_monitoring_check" "scripted" {
  ...
  settings {
    scripted {
      script = file("${path.module}/../dist/backend/get_recommendation.js" )
    }
  }
}
```

As you can imagine, this setup is far from ideal and requires users to install even more tools (e.g., node).

I thought it would be great to expose a Provider Function that does... exactly that. 

That's why this PR exists. Users will be able to do this instead:
```hcl
resource "grafana_synthetic_monitoring_check" "scripted" {
  ...
  settings {
    scripted {
      script = provider::grafana::k6bundle("${path.module}/../dist/backend/get_recommendation.js")
    }
  }
}
```


peterschretlen (commented):
I won't comment on the code review, but I 👍 the need for this and it's utility.  

This does come up from time-to-time on synthetics, and we see more and more usage of the terraform provider to manage large scale synthetics installations that are likely to use bundling so they can modularize their scripts.

fornfrey (approved):
A very much needed improvement! Great job 👍

PR #2365, opened 2025-09-24
App Platform Resource: prefer stack_id over org_id in the provider config

alexander-akhmetov:
org_id is 1 [by default](https://github.com/grafana/terraform-provider-grafana/blob/9e64b58779b8a974a85aae2c41549c6a8d4a7217/pkg/provider/framework_provider.go#L85), so to use app platform resources `org_id` must be explicitly set to 0:

```
provider "grafana" {
  url  = "https://...grafana.net/"
  auth = "..."
  org_id = 0
  stack_id = 123
}
```

This PR changes the order of checks to prefer `stack_id` over `org_id` when both are present, ensuring that Grafana Cloud users don't need to explicitly set `org_id = 0` in the provider configuration:

```
provider "grafana" {
  url  = "https://...grafana.net/"
  auth = "..."
  stack_id = 123
}
```

stack_id is [0 by default](https://github.com/grafana/terraform-provider-grafana/blob/9e64b58779b8a974a85aae2c41549c6a8d4a7217/pkg/provider/framework_provider.go#L91), so this does not change the behaviour when both org_id and stack_id are not set and org_id=1 is used.

### DIFF
--- a/internal/resources/appplatform/resource.go
+++ b/internal/resources/appplatform/resource.go
@@ -22,6 +22,10 @@ import (
 	"github.com/grafana/terraform-provider-grafana/v4/internal/common"
 )
 
+const (
+	errNamespaceMissingIDs = "Expected either Grafana org ID (for local Grafana) or Grafana stack ID (for Grafana Cloud) to be set"
+)
+
 // ResourceModel is a Terraform model for a Grafana resource.
 type ResourceModel struct {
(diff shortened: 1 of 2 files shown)


stephan-rayner (approved):
Slick change @alexander-akhmetov and thanks for the tests 😄

PR #2366, opened 2025-09-25
chore(CODEOWNERS): remove platform-monitoring ownership where an individual team already owns it

Duologic:
Currently @grafana/platform-monitoring gets requested for review on practically every PR, as individual teams own their resources now, I do not think this is necessary any longer.


stephan-rayner (approved):
Solid work @Duologic

PR #2369, opened 2025-09-26
Fix Fleet Management Private Connectivity

SeamusGrafana:
Currently, the Private Connectivity Values for Fleet Management return `null`, it looks like the API is using `agent-management` instead of `fleet_management`.

This need verifying, but it should fix that issue and correctly return the Private Connectivity details.


paobulas:
Could someone please review this pull request at your earliest convenience?

tpaschalis (approved):
I've tested this for an AWS stack and it seems to work 🎉  Thank you @SeamusGrafana for locating the fix and apologies for not getting on this faster.

```diff
$ diff aws-after.txt aws-before.txt
10c10
< data.grafana_cloud_stack.stack: Read complete after 0s [id=1473880]
---
> data.grafana_cloud_stack.stack: Read complete after 1s [id=1473880]
23,24c23,24
<       + fleet_management_private_connectivity_info_private_dns  = "fleet-management-prod-024-cortex-gw.eu-central-2.vpce.grafana.net"
<       + fleet_management_private_connectivity_info_service_name = "com.amazonaws.vpce.eu-central-2.vpce-svc-09bf9f5fbd9a0f5da"
---
>       + fleet_management_private_connectivity_info_private_dns  = null
>       + fleet_management_private_connectivity_info_service_name = null
```

We'll still need to follow-up with the details for Azure and GCP stacks, but I think this can be merged after checking the cloud acceptance tests

PR #2374, opened 2025-10-05
Add `v1alpha1.productactivation`

ArturWierzbicki:


radiohead (approved):
FYI @jcreixell would be nice if the generated types were in a repo which we could import them from, rather than copy-pasting them around like that.

jcreixell:
do you mean in a public repo just containing the resources? do you have an example I can look at?

jcreixell:
The metadata `ProcessedTimestamp` field is no longer used, I wonder if we should clean it up before publishing the contract. Other than that, I think this should be fine

ArturWierzbicki:
> The metadata `ProcessedTimestamp` field is no longer used, I wonder if we should clean it up before publishing the contract. Other than that, I think this should be fine

the contract exposed by TF here is just the `spec.enabled` - or am I missing something :)

jcreixell:
oh, you are right, in that case, LGTM

jcreixell (approved):

radiohead:
> do you mean in a public repo just containing the resources? do you have an example I can look at?
Yup, correct!

Unfortunately I don't have an example, but I _can_ think of a single repo for all public o11y schemas.

Here's an example from another place I used to work at – https://github.com/speechly/api, I'm sure there's more of the same on GitHub.

Let me know if that makes sense!

ArturWierzbicki:
@jcreixell @radiohead i fixed failing tests/docs/catalog info, lmk if you have other objections - otherwise ill merge tomorrow!

radiohead (approved):

jcreixell:
LGTM, Also tagging @mar4uk @t00mas in case they have capacity to work on publishing the generated types

PR #2381, opened 2025-10-15
Set the correct User-Agent in the k6 API client

fornfrey:


pkg/provider/configure_clients.go
```diff
@@ -387,6 +387,7 @@ func createK6Client(client *common.Client, providerConfig ProviderConfig) error
 			{URL: providerConfig.K6URL.ValueString()},
 		}
 	}
+	k6Cfg.UserAgent = providerConfig.UserAgent.ValueString()
```
fornfrey, line 390: Here is how the provider user agent is initialized https://github.com/grafana/terraform-provider-grafana/blob/135f25ac72994e47bde2b5f2d48176ba8db44a27/pkg/provider/framework_provider.go#L300

d14c (approved):

PR #2382, opened 2025-10-15
fix(oncall): enable creation of web based schedules through tf

pedroespindula:
## Description

This PR fixes an issue that prevented the creation of web type schedules in Grafana OnCall using Terraform. Although the provider was intended to support all schedule types, a validation conflict caused the creation of web schedules to fail.

The provider was incorrectly returning the error `Error: time_zone can not be set with type: web`. 

image

This contradicted the Grafana OnCall API, which requires the time_zone field for this schedule type and would fail with the error `Error: POST <oncall_api_url>/api/v1/schedules/: 400 {time_zone: [This field is required.]}`.

image

This fix resolves the conflicting validation logic, allowing time_zone to be correctly passed to the API for web schedules.

## Key Changes

 - `time_zone` logic: The validation for the time_zone field has been adjusted to allow its use with web type schedules, resolving the API conflict.
 - Forced recreation for `web` schedules: Implemented the CustomizeDiff function, which forces the resource to be recreated (ForceNew) whenever any of its attributes are changed and the type is web. This ensures that the Terraform state remains consistent, as changes in the Grafana UI are not reflected back into the state.
 - `web` schedule example: Added a new grafana_oncall_schedule resource example with type = "web" to demonstrate its correct usage, including the lifecycle { ignore_changes = all } block.

 ## How to Test

   1. Create a new grafana_oncall_schedule resource with type = "web".
   2. Run terraform apply and verify that the schedule was created correctly in Grafana OnCall, whereas it would have failed before.
   3. Change an attribute in the .tf file (e.g., the name).
   4. Run terraform apply again and confirm that the old schedule was destroyed and a new one was created.
   5. To test the synchronization, add the lifecycle { ignore_changes = all } block to the resource, modify the schedule in the Grafana UI, and then run terraform plan. No changes should be detected.

## Additional Details

* API request that was made to On Call to create the Web Schedule manually:
```bash
curl --location '$GRAFANA_ONCALL_URL/api/v1/schedules/' \
--header 'Content-Type: application/json' \
--header 'Authorization: $GRAFANA_ONCALL_ACCESS_TOKEN' \
--data '{
    "name": "[TEST] Example Schedule",
    "type": "web",
    "time_zone": "UTC"
}'
```
* Apply working with the modified version:
```tf
resource "grafana_oncall_schedule" "test" {
  name      = "[TEST] Example Schedule"
  type      = "web"
  time_zone = "UTC"
}
``` 
image

* Replace working with the modified version: 
image

* If no attributes changed, nothing happens:
image

* If ignore changes used, nothing happens:
```tf
resource "grafana_oncall_schedule" "test" {
  name      = "[TEST] Example Schedule - Changed"
  type      = "web"
  time_zone = "UTC"

  lifecycle {
    ignore_changes = all
  }
}
```

image
image

### DIFF
--- a/examples/resources/grafana_oncall_schedule/resource.tf
+++ b/examples/resources/grafana_oncall_schedule/resource.tf
@@ -42,3 +42,19 @@ resource "grafana_oncall_schedule" "example_schedule" {
   ]
   ical_url_overrides = "https://example.com/example_overrides_ical.ics"
 }
+
+// Web based schedule
+resource "grafana_oncall_schedule" "example_schedule" {
+  name      = "Example Calendar Schadule"
+  type      = "web"
+  time_zone = "America/New_York"
+
+  // Optional: specify the team to which the schedule belongs
+  team_id = data.grafana_oncall_team.my_team.id
+
+  // Is highly recommended to set ignore changes on all properties to avoid
+  // recreating the schedule all the time due to changes on the Web UI.
+  lifecycle {
+    ignore_changes = all
+  }
+}
--- a/internal/resources/oncall/resource_schedule.go
+++ b/internal/resources/oncall/resource_schedule.go
@@ -32,6 +32,8 @@ func resourceSchedule() *common.Resource {
 			StateContext: schema.ImportStatePassthroughContext,
 		},
 
+		CustomizeDiff: resourceScheduleCustomizeDiff,
+
 		Schema: map[string]*schema.Schema{
 			"name": {
 				Type:         schema.TypeString,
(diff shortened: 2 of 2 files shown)


jwelch92:
@pedroespindula Apologies for missing this PR! #2401 added support for creating web schedules with a `time_zone` field. 

Does the linked PR fix the issue you're seeing, or is there another issue that is addressed by this PR? Thanks!

pedroespindula:
Hey, @jwelch92! No problem! Happy it got fixed! 😄 

My PR just fixes this issue. I will closed it here. Thanks!

Issue #1705, opened 2024-07-22
[Feature Request]: Automatic token rotations via taint


julienduchesne:
### Feature Request

Resources:
- `grafana_service_account_token`
- `grafana_cloud_access_policy_token`
- `grafana_cloud_stack_service_account_token`

On these resources, it's possible to set a token expiration but it's up to the user to taint the resource and reapply it. For users that have automatic drift detection (and resolution), a way to automatically taint and reapply these resources would allow for rotation

I propose a `rotation_time_before_expiration` which will allow users to set an expiration and also a duration that will trigger a rotation if (current time - expiration timer) is lesser than that rotation time. For example, an expiration time of 30d and a rotation time of 7d means that the token will be rotated via taint after 23d (7d from expiration)

mjseaman:
Hi @julienduchesne and @tleopold , I'm doing a little feature request triage. This issue is in the backlog, does that mean you intend to do it?

undefinedhuman:
Hi there, we are able to rotate the token in time before expiration with the Hashicorp/time provider, but the rotation would also delete/invalidate the old token and therefore an instant update of our entire Infra would be necessary (which is not possible) - would it be possible to implement a flag for the tokens "delete = false/true"

Our current example ():

```hcl
terraform {
  required_providers {
    grafana = {
      source = "grafana/grafana"
      version = "3.12.0"
    }
  }
}

provider "grafana" {
    alias = "cloud"
}

data "grafana_cloud_stack" "test" {
    provider = grafana.cloud
    slug = "test"
}

// Step 2: Install Synthetic Monitoring on the stack
resource "grafana_cloud_access_policy" "push_policy" {
  provider = grafana.cloud

  region = data.grafana_cloud_stack.test.region_slug
  name   = "metric-publisher-test"
  scopes = ["metrics:write", "logs:write", "traces:write"]
  realm {
    type       = "stack"
    identifier = data.grafana_cloud_stack.test.id
  }
}

resource "time_rotating" "push_token_expiration" {
    rotation_minutes = 4
}

resource "grafana_cloud_access_policy_token" "push_policy" {
  provider = grafana.cloud

  region           = data.grafana_cloud_stack.test.region_slug
  access_policy_id = grafana_cloud_access_policy.push_policy.policy_id
  name             = "metric-publisher-devops"

  expires_at = timeadd(time_rotating.push_token_expiration.rfc3339, "8m")

  lifecycle {
    create_before_destroy = true
  }
[...]

PR #2390, opened 2025-10-17
feat: Add `grafana_cloud_access_policy_rotating_token` resource

volcanonoodle:
Related to https://github.com/grafana/terraform-provider-grafana/issues/1705

#### Separate resource
Originally we planned on updating the `grafana_cloud_access_policy_token` resource with a couple of new fields to specify the rotation parameters and a couple of new fields to specify the `computed_name` and `computed_expires_at`, but since it created a bit of a split behaviour on the resource depending on whether rotation was enabled or not, we ended up going with a separate resource for rotating tokens.

We have extracted as much logic into common functions that can be shared by both the existing token resource and the new rotating token resource, such as most of each of the CRUD operations and the common schema fields between the 2 resources.

#### Computed `name`
Since Grafana Cloud tokens have the restriction of having unique names within an org, and we need to have both the old and new tokens co-exist when a rotation happens, we need to ensure that the each token's name is unique. We accomplish that by adding strings to each token's name that are unique for each rotation, like `rotate_after` and `post_rotation_lifetime`. By doing this, we also get to use the name as the "cloud" storage for those fields, since otherwise these would only live in Terraform because the API does not store them and we'd not be able to set them when importing a rotating token into Terraform, for example.

#### Computed `expires_at` from `rotate_after` and `post_rotation_lifetime` fields
For rotating tokens, `expires_at` is not user-defined. It's computed from adding the `rotate_after` timestamp and the `post_rotation_lifetime` duration. If a user wants their token to expire in 3 months and they want to rotate it at some point in the last month of that period, they should set `rotate_after = <timestamp of 2 months in the future>` and `post_rotation_lifetime = 30d`.

### Ideas considered that were not implemented but are not completely discarded either
* Adding a `-v1` suffix to the rotating tokens' `name`. This way if we need to evolve the resource, we have an easy way to distinguish between existing rotating tokens from the current provider version and rotating tokens from a newer version of the provider which might use a different pattern to compute the `name` field.
* Making `rotate_after` an RFC3339 string, as opposed to a unix timestamp, which was the original idea. The former seems better for humans but we need to add diff-suppression logic to take timezone differences into account, since 2 RFC3339 strings can represent the same time in different timezones but swapping between them would trigger a resource renewal by default in Terraform. However, TF has `static_time` and `rotation_time` resources that allow turning RFC3339 strings into unix timestamps, so it might not be a big deal to leave it as a unix timestamp.
* Making `expires_at` user-defined, as the hard expiration time, and just add a `rotate_before` field that would specify how early the token can be rotated before it expires. So essentially, any TF plan that happened after `expires_at - rotate_before` would trigger a rotation, while the token would continue to live until its `expires_at` time. It would replace the `rotate_after` and `post_lifetime_rotation` fields. It makes it easier to reason about the date at which the token will expire, but it might make it more difficult to think about the rotation timings.

---

Additionally it fixes a problem with Cloud Access Policy token tests using LBAC with scopes that aren't compatible with it in https://github.com/grafana/terraform-provider-grafana/pull/2390/commits/7363cbcbb56fbd7b7f35092c3a8cb88e2ea81e32


volcanonoodle:
Looks like https://github.com/grafana/terraform-provider-grafana/pull/2394 will fix the acceptance tests

internal/resources/cloud/resource_cloud_access_policy_rotating_token.go
```diff
@@ -0,0 +1,284 @@
...
+		},
+
+		CustomizeDiff: func(ctx context.Context, d *schema.ResourceDiff, meta any) error {
+			customDiffErr := errors.New("error while generating the customized diff")
+
+			name, err := computeRotatingTokenName(d)
+			if err != nil && !errors.Is(err, errEmptyTokenRotationValue) {
+				return fmt.Errorf("%w: %w", customDiffErr, err)
+			}
+			if name != "" {
```
dmihai: I don't think we need to check whether name != "" because `computeRotatingTokenName()` should never return an empty string. Worst case, it returns `--` if all `rotatingTokenNameAttributes` are empty.

internal/resources/cloud/resource_cloud_access_policy_rotating_token.go
```diff
@@ -0,0 +1,284 @@
...
+				ForceNew:     true,
+				Description:  "Prefix for the name of the access policy token. The actual name will be stored in the computed field `name`, which will be in the format '<name_prefix>-<rotate_after>-<post_rotation_lifetime>'",
+				ValidateFunc: validation.StringLenBetween(1, 200),
+			},
+			"rotate_after": {
+				Type:         schema.TypeInt,
+				Required:     true,
+				ForceNew:     true,
+				Description:  "The time after which the token will be rotated, as a unix timestamp (number of seconds elapsed since epoch time - January 1, 1970 UTC).",
+				ValidateFunc: validation.IntAtLeast(0),
```
dmihai: Do you think we should specify an upper limit for `rotate_after`? Maybe some timestamp of a future date such as 2080-01-01.

internal/resources/cloud/resource_cloud_access_policy_rotating_token.go
```diff
@@ -0,0 +1,284 @@
...
+}
+
+func rotatingTokenAttributesFromName(name string) (*rotatingTokenNameAttributes, error) {
+	parts := strings.Split(name, "-")
+	if len(parts) < 3 {
+		return nil, fmt.Errorf("rotating token name does not follow the expected pattern '<name_prefix>-<rotate_after>-<post_rotation_lifetime>': %s", name)
+	}
+
+	// rotateAfter
+	rotateAfterStr := parts[len(parts)-2]
```
dmihai: I think it's easier to write this as `parts[1]`.

internal/resources/cloud/resource_cloud_access_policy_rotating_token.go
```diff
@@ -0,0 +1,284 @@
...
+	rotateAfter, err := strconv.ParseInt(rotateAfterStr, 10, 64)
+	if err != nil || rotateAfter <= 0 {
+		return nil, fmt.Errorf("could not infer 'rotate_after' from rotating token name: %w", err)
+	}
+
+	// postRotationLifetime
+	postRotationLifetime := parts[len(parts)-1]
```
dmihai: This can be `parts[2]`.

internal/resources/cloud/resource_cloud_access_policy_rotating_token.go
```diff
@@ -0,0 +1,284 @@
...
+	if _, err = time.ParseDuration(postRotationLifetime); err != nil {
+		return nil, fmt.Errorf("could not infer 'post_rotation_lifetime' from rotating token name: %w", err)
+	}
+
+	// namePrefix
+	namePrefix := strings.Join(parts[:len(parts)-2], "-")
```
dmihai: `parts[:1]` here as well.

dmihai (approved):
LGTM, I just have some minor comments. I leave it up to you if you want to address them. Thank you!

cinaglia:
> Making expires_at user-defined, as the hard expiration time, and just add a rotate_before field that would specify how early the token can be rotated before it expires. So essentially, any TF plan that happened after expires_at - rotate_before would trigger a rotation, while the token would continue to live until its expires_at time. It would replace the rotate_after and post_lifetime_rotation fields. It makes it easier to reason about the date at which the token will expire, but it might make it more difficult to think about the rotation timings.

FYI @volcanonoodle discussed offline and decided to experiment with pursuing this. We agreed it makes for a cleaner UX. He's going to work on that over the next couple of days.

internal/resources/cloud/resource_cloud_access_policy_rotating_token.go
```diff
@@ -0,0 +1,284 @@
...
+		d.Set("name_prefix", attrs.namePrefix),
+	)
+	return diag.FromErr(err)
+}
+
+func deleteCloudAccessPolicyRotatingToken(ctx context.Context, d *schema.ResourceData, client *gcom.APIClient) diag.Diagnostics {
+	return diag.Diagnostics{
+		diag.Diagnostic{
+			Severity: diag.Warning,
+			Summary:  "Rotating tokens do not get deleted.",
```
cinaglia: What are we protecting against by preventing the deletion of rotating tokens? I'm concerned that this might end up being problematic. The API itself supports the deletion, so I'm not quite convinced it's a good idea to just let this be automatically expired. For instance, what if someone wants to delete a token in a jiffy as part of a security incident / leak?

internal/resources/cloud/resource_cloud_access_policy_rotating_token.go
cinaglia: I believe this is here to guard against name prefixes containing hyphens (e.g. `my-token-1761323605-2m`). If we blindly go with `parts[1]`, `rotateAfterStr` will end up being `token`.

volcanonoodle: @cinaglia exactly 👍  I'll add a comment to the code and maybe try to write a test to ensure that it protects against a `name_prefix` with hyphens, to make it more explicit. Thanks!

internal/resources/cloud/resource_cloud_access_policy_rotating_token.go
volcanonoodle: When a token is rotated, the underlying Terraform resource is destroyed and re-created. By not deleting the token in Grafana Cloud when the TF resource is destroyed, we allow both the old token and the new token to co-exist for the `post_rotation_lifetime` period.

If we forced the deletion of the token when the TF resource is destroyed, I believe it'd behave exactly as the regular (non-rotating) token resource with a [lifecycle policy](https://developer.hashicorp.com/terraform/language/meta-arguments#lifecycle) of `create_before_destroy`, where replacing the TF resource results in a new token being created and the old token being deleted right after.

We could add an optional field like `delete_rotated_token` (final name TBD) that defaults to `false` and triggers the deletion of a token as soon as it gets rotated, as we once discussed. I believe most users will want to combine the use of this field with the `create_before_destroy` lifecycle policy.

I implemented it in one of the WIP commits, but ended up thinking maybe it wasn't adding enough value and removed it haha. I'm happy to add it back, though, as it'd definitely be useful for the use case you mention.

WDYT about releasing this new resource without the deletion field and adding it in the next version of the terraform provider? Just to make sure we ship this new resource in time.

internal/resources/cloud/resource_cloud_access_policy_rotating_token.go
volcanonoodle: Sounds like a good idea to improve the UX, thanks!

internal/resources/cloud/resource_cloud_access_policy_rotating_token.go
volcanonoodle: `computeRotatingTokenName()` returns "" in case of error, and we ignore `errEmptyTokenRotationValue` errors, which means that `name` could be "".

I did this so long ago that I'm not sure I remember the reason correctly, but I think it was because in Terraform you can set a field to have a value that will depend on another resource that will be created when the plan is applied, meaning that during the diff the value will be "unkown" and we'll get an empty value. We need to ignore those, rather than throwing an error, which is why I added `errEmptyTokenRotationValue` and ignored it. I will double check that this is the case and add a comment explaining it, because I just realized that it's not obvious at all, thanks! 😆

internal/resources/cloud/resource_cloud_access_policy_rotating_token.go
cinaglia: > We could add an optional field like delete_rotated_token (final name TBD) that defaults to false and triggers the deletion of a token as soon as it gets rotated, as we once discussed. I believe most users will want to combine the use of this field with the create_before_destroy lifecycle policy.

> WDYT about releasing this new resource without the deletion field and adding it in the next version of the terraform provider? Just to make sure we ship this new resource in time.

I think that's a great idea. I'd be happy to take a stab at it as well (for a short change of scenery, 😄 ).

Duologic (approved):
General approve on the idea, didn't go in depth, folks on the team working on this are  probably better suited.

volcanonoodle:
In the end I think the approach in https://github.com/grafana/terraform-provider-grafana/pull/2405 is the one that looks the simplest and most elegant to me. Happy to hear your thoughts on it. I've based https://github.com/grafana/terraform-provider-grafana/pull/2405 on this PR so that the diff shows the differences from it.

docs/resources/cloud_access_policy_rotating_token.md
```diff
@@ -0,0 +1,93 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "grafana_cloud_access_policy_rotating_token Resource - terraform-provider-grafana"
+subcategory: "Cloud"
+description: |-
```
Rydez, line 5: This description looks like it has a couple typos: the `API documentation` in the first link and the scopes have no separation. Not sure if this matters or not. I'm guessing this is only used for SEO.

volcanonoodle: Oh that's a nice finding, thanks, I hadn't noticed it. I think it's only for seach engines yeah. It seems it happens in other resources too (example [cloud_stacks](https://github.com/grafana/terraform-provider-grafana/blob/c75349cbc5d9242a3a6a844b01f5708a3af0c3e4/docs/resources/cloud_stack.md)) 😕

Looks like it might be caused by the `plainmarkdown` directive from the [Terraform docs generator](https://github.com/hashicorp/terraform-plugin-docs#template-functions). We might be able to get around it by adding a custom template for our resources in https://github.com/grafana/terraform-provider-grafana/tree/main/templates/resources but I'm guessing we will want to get other higher-priority work done before tackling this.

Rydez (approved):
Spent some time looking through everything yesterday and, from my end, I think this looks good now that #2405 is merged in. Nice Work!

volcanonoodle:
I'm planning on merging this and creating a minor release of the Terraform provider on Monday

PR #2393, opened 2025-10-22
Update k6 test schedules to support cron expressions

fornfrey:
https://github.com/grafana/k6-cloud/issues/3652

Update k6 test schedule resource and data sources to support cron expressions.

### DIFF
--- a/examples/data-sources/grafana_k6_schedules/data-source.tf
+++ b/examples/data-sources/grafana_k6_schedules/data-source.tf
@@ -48,10 +48,9 @@ resource "grafana_k6_schedule" "test_schedule_1" {
 resource "grafana_k6_schedule" "test_schedule_2" {
   load_test_id = grafana_k6_load_test.schedules_load_test_2.id
   starts       = "2023-12-26T14:00:00Z"
-  recurrence_rule {
-    frequency = "WEEKLY"
-    interval  = 2
-    until     = "2047-01-31T23:59:59Z"
+  cron {
+    schedule = "0 10 1 12 6"
+    timezone = "UTC"
   }
 
   depends_on = [
--- a/examples/data-sources/grafana_k6_schedules/data-source.tf
+++ b/examples/data-sources/grafana_k6_schedules/data-source.tf
@@ -30,6 +30,21 @@ resource "grafana_k6_load_test" "schedules_load_test_2" {
   ]
 }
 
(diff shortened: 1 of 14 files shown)


examples/data-sources/grafana_k6_schedules/data-source.tf
```diff
@@ -48,10 +48,9 @@ resource "grafana_k6_schedule" "test_schedule_1" {
 resource "grafana_k6_schedule" "test_schedule_2" {
   load_test_id = grafana_k6_load_test.schedules_load_test_2.id
   starts       = "2023-12-26T14:00:00Z"
-  recurrence_rule {
-    frequency = "WEEKLY"
-    interval  = 2
-    until     = "2047-01-31T23:59:59Z"
+  cron {
```
DefCon-007, line 80: In place of replacing the recurrence_rule example, should we add a new example for cron expression and keep both?

internal/resources/k6/data_source_k6_schedule.go
```diff
@@ -108,6 +109,19 @@ func (d *scheduleDataSource) Schema(_ context.Context, _ datasource.SchemaReques
...
 			},
+			"cron": schema.SingleNestedBlock{
+				Description: "The cron schedule to trigger the test periodically. If null, the test will run only once on the 'starts' date.",
+				Attributes: map[string]schema.Attribute{
+					"schedule": schema.StringAttribute{
+						Description: "A cron expression with exactly 5 entries, or an alias. The allowed aliases are: @yearly, @annually, @monthly, @weekly, @daily, @hourly.",
+						Computed:    true,
+					},
+					"timezone": schema.StringAttribute{
+						Description: "The timezone of the cron expression.",
```
DefCon-007: Nit: Adding an example in the description might be helpful to illustrate the expected format for timezones.

examples/data-sources/grafana_k6_schedules/data-source.tf
fornfrey: Done ✅

internal/resources/k6/data_source_k6_schedule.go
fornfrey: Makes sense. Done 👍

DefCon-007 (approved):

PR #2411, opened 2025-11-03
Bug Fix: Add missing type to escalation policy

mcschatz:
# Background

Why is this change required?

Bug reported in this issue: https://github.com/grafana/irm/issues/7362 

UI supports creating escalation policies with type: notify_if_num_alerts_in_window, but it is not supported via terraform. This PR fixes that gap. 

Related PRs: 
1. https://github.com/grafana/irm/pull/7480
2. https://github.com/grafana/amixr-api-go-client/pull/39


matiasb (approved):
LGTM, wondering if we should check that both fields are required if setting up this step (not blocking, but it doesn't seem we enforce this in the API either, and the escalation step will fail and log the missing bit in the timeline)

mcschatz:
@matiasb I had this conversation with @willgallego-grafana yesterday. We don't currently enforce it, but it seems like a change we should make as a follow up. I was surprised you could create this type without the two parameters.

PR #2418, opened 2025-11-10
feat: Added resource for data source cache config

### LINKED ISSUE #765: Add Caching support to Datasources
Hi,

Currently datasource caching ([Documentation](https://grafana.com/docs/grafana/latest/administration/data-source-management/#query-caching)) is not supported by the Provider. 

We're moving to Enterprise instances and would like to enable Caching using Terraform. Asked Grafana about provisioning option and got this reply:

```
Hello, 
 
Unfortunately as of now, is not possible to enable/disable cache for a datasource via provisioning, but there is an API you could programmatically call to enable it as part of your deployment:
 
/api/datasources/:datasourceUId/cache/enable
/api/datasources/:datasourceUId/cache/disable
 
These APIs are yet to be documented but you should be able to use them.
 
In regards of having this feature enabled as part of provisioning, I can certainly pass this over to our Voice of the Customer team to let them know you're interested in this. 
 
Please let me know if you have any further questions or comments. 

```

So the API supports it, can this be added to the data_source resources?

bradleypettit:
This adds support for managing data source cache config via a new resource `grafana_data_source_cache_config`.

Leverages the [Query and Resourcing Caching API](https://grafana.com/docs/grafana/latest/developers/http_api/query_and_resource_caching) to toggle and config[ure Query and Resource Caching](https://grafana.com/docs/grafana/latest/administration/data-source-management/#query-and-resource-caching), an enterprise & cloud feature available to configure on a per data source basis.

Fixes https://github.com/grafana/terraform-provider-grafana/issues/765


bradleypettit:
Thanks in advance for the review. This is the first time I've worked with a Terraform provider from this side, so might have overlooked something.

Tests wise, I think the failing tests are unrelated? but I'm also not 100% sure that the tests for this new resource ran. It's marked as enterprise only which might explain why, but I would expect them to still run on a cloud instance. If you have a look and advise that would be great.

internal/resources/grafana/catalog-resource.yaml
```diff
@@ -92,6 +92,19 @@ spec:
...
 kind: Component
+metadata:
+  name: resource-grafana_data_source_cache_config
+  title: grafana_data_source_cache_config (resource)
+  description: |
+    resource `grafana_data_source_cache_config` in Grafana Labs' Terraform Provider
+spec:
+  subcomponentOf: component:default/terraform-provider-grafana
+  type: terraform-resource
+  owner: group:default/data-sources
```
dana-axinte:
```suggestion
  owner: group:default/grafana-operator-experience-squad
```
query caching is under operator experience for now, we'll change when we'll hand over it

internal/resources/grafana/resource_data_source_cache_config.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+func CreateOrUpdateDataSourceCacheConfig(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client, _ := OAPIClientFromNewOrgResource(meta, d)
+	dsUID := d.Get("datasource_uid").(string)
+
+	// If enabled is explicitly set to false, disable via dedicated endpoint to avoid omitempty issues.
+	if _, present := d.GetOk("enabled"); present || d.HasChange("enabled") {
+		if !d.Get("enabled").(bool) {
+			if _, err := client.Enterprise.DisableDataSourceCache(dsUID); err != nil {
```
dana-axinte: when disable endpoint is called, this will error if the config doesn't exist

you'd need to check if the config exist, then call disable if `enabled` is false

bradleypettit: OK I actually was struggling to simulate the error - if I created the resource from scratch with `enabled` being false, it seemed to all work OK. Though, I'm not sure if this was being handled by Terraform or somewhere else.

Regardless - it's a good idea to cover for this and make it explicit so I've added some logic and tested it OK.

bradleypettit:
I think the acceptance tests that are failing are unrelated to the changes here 🧐

Edit: confirmed :)

dana-axinte (approved):

PR #2424, opened 2025-11-12
Make k6 API URL configurable in the plugin installation

fornfrey:
This PR makes the previously hardcoded k6 URL configurable in the `grafana_k6_installation` resource.

The URL is already configurable on the provider level for other k6 resources.

### DIFF
--- a/docs/resources/k6_installation.md
+++ b/docs/resources/k6_installation.md
@@ -108,6 +108,10 @@ resource "grafana_k6_project" "my_k6_project" {
 - `grafana_user` (String) The user to use for the installation.
 - `stack_id` (String) The identifier of the stack to install k6 on.
 
+### Optional
+
+- `k6_api_url` (String) The URL of the k6 API.
+
 ### Read-Only
 
 - `id` (String) The ID of this resource.
--- a/docs/resources/k6_installation.md
+++ b/docs/resources/k6_installation.md
@@ -110,7 +110,7 @@ resource "grafana_k6_project" "my_k6_project" {
 
 ### Optional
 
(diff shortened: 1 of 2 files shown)


docs/resources/k6_installation.md
```diff
@@ -108,6 +108,10 @@ resource "grafana_k6_project" "my_k6_project" {
 - `grafana_user` (String) The user to use for the installation.
 - `stack_id` (String) The identifier of the stack to install k6 on.
 
+### Optional
+
+- `k6_api_url` (String) The URL of the k6 API.
```
federicotdn: Should we add some sort of comment saying that most users won't need to configure this?

federicotdn (approved):

docs/resources/k6_installation.md
fornfrey: The descriptions of the rest of the custom API URL [parameters](https://github.com/grafana/terraform-provider-grafana/blob/k6-installation-url/docs/index.md#optional) does not mention this, so I think it's ok to not mention that. That said, I updated the name in the description to "Grafana Cloud k6 API".

docs/resources/k6_installation.md
```diff
@@ -110,7 +110,7 @@ resource "grafana_k6_project" "my_k6_project" {
 
 ### Optional
 
-- `k6_api_url` (String) The URL of the k6 API.
+- `k6_api_url` (String) The Grafaba Cloud k6 API url.
```
federicotdn: typo! Grafaba

fornfrey: Sorry 🤦 Fixed, thanks

PR #2426, opened 2025-11-13
fix(rules): temp switch to map[string]string for compat

rwwiv:
Dynamic types currently cause compilation to fail in [grafana/crossplane-provider-grafana](https://github.com/grafana/crossplane-provider-grafana), so revert to a simple map until this is fixed in that project.


moustafab (approved):
LGTM

PR #2431, opened 2025-11-14
Prometheus type migration check

bossinc:
Closes https://github.com/grafana/data-sources/issues/649

This adds a check to see if users are creating Prom data sources that use deprecated Azure and Amazon authentication. If they are using deprecated auth, it warns them to update the data source type.


aangelisc (approved):
LGTM

PR #2436, opened 2025-11-18
feat(service-accounts): Add rotating token resource

volcanonoodle:
Related to https://github.com/grafana/terraform-provider-grafana/issues/1705

Adds a rotating token resource for Service Account tokens like done with Cloud Access Policy tokens in https://github.com/grafana/terraform-provider-grafana/pull/2390.

I think a concern here is that AFAIK we do not have an API in Grafana to fetch specific tokens by ID, but rather we have to list all the tokens that belong to a Service Account ([endpoint](https://github.com/grafana/grafana/blob/db0b35f6dfd6fd2396d5b85166e135314fe97f7f/pkg/services/serviceaccounts/api/token.go#L66)) and we cannot filter out expired tokens either. This means that in the long term, as this new Terraform resource keeps rotating tokens, the list of expired tokens returned by Grafana on each TF plan will keep growing. I don't think we should worry about it for now, as it seems unlikely that someone will accumulate too many rotated tokens under one Service Account in the near future, but at some point we might need to consider adding a filter for expired tokens to the endpoint mentioned previously.


internal/resources/grafana/resource_service_account_rotating_token.go
```diff
@@ -0,0 +1,151 @@
...
+		schema,
+	)
+}
+
+func serviceAccountRotatingTokenCreate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	namePrefix := d.Get("name_prefix").(string)
+	ttl := d.Get("seconds_to_live").(int)
+
+	expiration := ServiceAccountRotatingTokenNow().Add(time.Duration(ttl) * time.Second)
+	name := fmt.Sprintf("%s-%d", namePrefix, expiration.Unix())
```
cinaglia, line 120: This is probably not an issue but there's a chance the expiration unix timestamp might not match exactly the action expiration timestamp stored in Grafana.

internal/resources/grafana/resource_service_account_rotating_token.go
```diff
@@ -0,0 +1,151 @@
...
+		ReadContext:   serviceAccountRotatingTokenRead,
+		UpdateContext: serviceAccountRotatingTokenRead,
+		DeleteContext: serviceAccountRotatingTokenDelete,
+
+		CustomizeDiff: func(ctx context.Context, d *schema.ResourceDiff, meta any) error {
+			secondsToLive := d.Get("seconds_to_live").(int)
+			earlyRotationWindowSec := d.Get("early_rotation_window_seconds").(int)
+
+			if earlyRotationWindowSec > secondsToLive {
+				return fmt.Errorf("`early_rotation_window_seconds` cannot be bigger than `seconds_to_live`")
```
cinaglia: nit:
```suggestion
				return fmt.Errorf("`early_rotation_window_seconds` cannot be greater than `seconds_to_live`")
```

cinaglia (approved):
LGTM. I've been able to test this locally. One thing I'm not particularly in love with is how we're exposing `ServiceAccountRotatingTokenNow`, though I don't think there's a much better way to do this.

internal/resources/grafana/resource_service_account_rotating_token.go
volcanonoodle: Yeah, it probably won't 😕  We could use a random string for the name suffix instead but I don't think there's much value in that given that the name suffix shouldn't be relied on for anything. Now that you mention this, maybe I should replace the description of the `name_prefix` from `<name_prefix>-<expiration timestamp>` to `<name_prefix>-<computed_suffix>`, so that we are not misleading anyone making them think that the suffix is the accurate expiration timestamp
https://github.com/grafana/terraform-provider-grafana/blob/40df264c37ae521e36adaea04dbe24218c16b67e/internal/resources/grafana/resource_service_account_rotating_token.go#L66

I'll go ahead and edit the description if that sounds good to you @cinaglia . Thanks for pointing this out! 😄

volcanonoodle:
> One thing I'm not particularly in love with is how we're exposing `ServiceAccountRotatingTokenNow`, though I don't think there's a much better way to do this.

@cinaglia I totally agree. It doesn't look very good but I couldn't come up with something better given that all the resources live under the same `cloud` package and the tests are in a separate package. Oh well, if anyone comes up with something more elegant we can always change it in the future 😄 Thank you for the review!

internal/resources/grafana/resource_service_account_rotating_token.go
cinaglia: That sounds like a great idea!

cinaglia (approved):

cinaglia (approved):

volcanonoodle:
I just realized that a couple of tests were setting the time based on the time when the test started, rather than on the token's expiration time, when we actually want the latter because it's the expiration time that's used to determine if the token needs rotation and there could be a second difference between the 2 depending on how long the initial request to create the token took. I fixed it in https://github.com/grafana/terraform-provider-grafana/pull/2436/commits/d368bb87f3315361b78805fe48e4a52bc3f501b1.

volcanonoodle:
I'll wait to merge this until https://github.com/grafana/terraform-provider-grafana/pull/2445 has been approved, so that we can release both resources under the same minor version bump of the provider.

PR #2439, opened 2025-11-19
delete pdc fields from json_data when generating updated state

dafydd-t:
Unit test run with `make testacc-oss-docker TESTARGS="-run ^TestAccDatasource_PDCPropertiesRemovedFromState$"`

Related to https://github.com/grafana/support-escalations/issues/18903

Prior to this change, when applying a `grafana_data_source` resource that has a `private_data_source_connect_network_id` property, the `json_data_encoded` field for this data source in the returned state would contain a `enableSecureSocksProxy` and a `enableSecureSocksProxy` key. 

This diff between the config and the state is suppressed by the  `DiffSuppressFunc` for data source json, but there may be times when the marshalling of this map would produce different results, since go maps have no ordering guarantees. 

I have not been able to reproduce the issue seen in https://github.com/grafana/support-escalations/issues/18903, but it seems like a good idea to remove this diff between state and config anyway.

### DIFF
--- a/internal/resources/grafana/resource_data_source.go
+++ b/internal/resources/grafana/resource_data_source.go
@@ -19,7 +19,13 @@ import (
 	"github.com/grafana/terraform-provider-grafana/v4/internal/common"
 )
 
+const (
+	pdcEnableSecureSocksProxy   = "enableSecureSocksProxy"
+	pdcSecureSocksProxyUsername = "secureSocksProxyUsername"
+)
+
 func resourceDataSource() *common.Resource {
+
 	schema := &schema.Resource{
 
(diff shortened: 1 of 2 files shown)


aangelisc (approved):
LGTM. Technically, users can specify these fields but that's not the supported mechanism for marking a data source as supporting PDC.

dafydd-t:
> LGTM. Technically, users can specify these fields but that's not the supported mechanism for marking a data source as supporting PDC.

I'm going to add a validation check which will fail if either of these fields are defined in a data source's jsonData.

aangelisc (approved):

PR #2446, opened 2025-11-24
Faro: Update endpoint-logic with region cutoff dates

t00mas:
For a specific region, a cutoff date needs to be taken into account to point to the correct faro endpoint.


t00mas:
> [unit tests / docs (pull_request)](https://github.com/grafana/terraform-provider-grafana/actions/runs/19637411192/job/56231513858?pr=2446)Failing after 1m

unrelated failure, will re-merge from main again when fixed

kpelelis (approved):

PR #2453, opened 2025-11-26
chore: remove dependabot in favor of renovate

Duologic:
title says it.


suntala (commented):
> In order to lower resource usage and have a faster runtime, PRs will not run Cloud tests automatically.
> To do so, a Grafana Labs employee must trigger the cloud acceptance tests workflow manually.

Are PRs going to be mergeable without running the cloud acceptance tests? I assume we don't want this to be the case and am wondering how this is going to be documented so that we don't forget to do it.

cindy (approved):

PR #2467, opened 2025-12-05
Add a ListCollectors cache the collectorResource

spartan0x117:
Uses a cached `ListCollectors` response in the `collectorResource`'s `Read()` function to reduce the number of calls while performing `terraform plan`. 

Note that `terraform apply` will still make N `Create`/`Update`/`DeleteCollector` requests + N `GetCollector` requests. This was left as-is for now as updates aren't done in bulk and caching a `ListCollectors` request would have stale collectors for all but the first call to `Update()`.

### DIFF
--- a/internal/resources/fleetmanagement/resource_collector.go
+++ b/internal/resources/fleetmanagement/resource_collector.go
@@ -117,6 +117,9 @@ func (r *collectorResource) ImportState(ctx context.Context, req resource.Import
 		return
 	}
 
+	// Invalidate cache after import to ensure subsequent reads get fresh data
+	r.resetCache()
+
 	state, diags := collectorMessageToResourceModel(ctx, getResp.Msg)
 	resp.Diagnostics.Append(diags...)
 	if resp.Diagnostics.HasError() {
--- a/internal/resources/fleetmanagement/resource_collector.go
+++ b/internal/resources/fleetmanagement/resource_collector.go
@@ -153,6 +153,9 @@ func (r *collectorResource) Create(ctx context.Context, req resource.CreateReque
 		return
 	}
 
+	// Invalidate cache after mutation
+	r.resetCache()
+
 	getReq := &collectorv1.GetCollectorRequest{
 		Id: collector.Id,
 	}
@@ -247,6 +250,9 @@ func (r *collectorResource) Update(ctx context.Context, req resource.UpdateReque
 		return
 	}
 
+	// Invalidate cache after mutation
+	r.resetCache()
+
 	getReq := &collectorv1.GetCollectorRequest{
 		Id: collector.Id,
 	}
@@ -285,4 +291,14 @@ func (r *collectorResource) Delete(ctx context.Context, req resource.DeleteReque
 		resp.Diagnostics.AddError("Failed to delete collector", err.Error())
 		return
 	}
+
+	// Invalidate cache after mutation
+	r.resetCache()
+}
+
+// resetCache invalidates the ListCollectors cache so the next Read will fetch fresh data
+func (r *collectorResource) resetCache() {
+	r.listOnce = sync.Once{}
+	r.collectorCache = nil
+	r.listErr = nil
 }
--- a/internal/resources/fleetmanagement/resource_collector.go
+++ b/internal/resources/fleetmanagement/resource_collector.go
@@ -2,6 +2,7 @@ package fleetmanagement
 
 import (
 	"context"
+	"sync"
 
 	"connectrpc.com/connect"
 	collectorv1 "github.com/grafana/fleet-management-api/api/gen/proto/go/collector/v1"
@@ -34,6 +35,11 @@ var (
 
 type collectorResource struct {
 	client collectorv1connect.CollectorServiceClient
+
+	// Cache for ListCollectors result for plan/refresh
+	listOnce       sync.Once
+	collectorCache map[string]*collectorv1.Collector
+	listErr        error
 }
 
 func newCollectorResource() *common.Resource {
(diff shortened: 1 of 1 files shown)


internal/resources/fleetmanagement/resource_collector.go
```diff
@@ -177,24 +186,38 @@ func (r *collectorResource) Read(ctx context.Context, req resource.ReadRequest,
 		return
 	}
 
-	getReq := &collectorv1.GetCollectorRequest{
-		Id: data.ID.ValueString(),
+	// Use cached ListCollectors result for plan/refresh
+	r.listOnce.Do(func() {
```
johannaojeling, line 193: What will happen if the initial `ListCollectors` call fails due to a transient network error or temporary API issue? Will the ListCollectors call not be retried, i.e. is there a risk that all subsequent Read operations fail for the entire lifetime of this resource instance because `sync.Once` would prevent retry attempts and `listErr` stays non-nil?

internal/resources/fleetmanagement/resource_collector.go
```diff
@@ -265,4 +291,14 @@ func (r *collectorResource) Delete(ctx context.Context, req resource.DeleteReque
 		resp.Diagnostics.AddError("Failed to delete collector", err.Error())
 		return
 	}
+
+	// Invalidate cache after mutation
+	r.resetCache()
+}
+
+// resetCache invalidates the ListCollectors cache so the next Read will fetch fresh data
+func (r *collectorResource) resetCache() {
```
johannaojeling, line 303: Should the cache also be reset when a Collector resource is imported with [ImportState](https://github.com/grafana/terraform-provider-grafana/blob/62104cf75c56eccc581c930e5e9de70ea6442002/internal/resources/fleetmanagement/resource_collector.go#L110)? It will be created outside of the [Create](https://github.com/grafana/terraform-provider-grafana/blob/62104cf75c56eccc581c930e5e9de70ea6442002/internal/resources/fleetmanagement/resource_collector.go#L133) flow, so the cache might not reflect the newly imported collector.

johannaojeling (commented):
Thanks for addressing this! 🙌

internal/resources/fleetmanagement/resource_collector.go
spartan0x117: Great question!

I checked and [we use](https://github.com/grafana/terraform-provider-grafana/blob/main/pkg/provider/configure_clients.go#L422) a [`retryablehttp`](https://github.com/hashicorp/go-retryablehttp) client, which transparently will retry the `ListCollectors` request if there's an issue, which happens _inside_ the `sync.Once`. 

If the retries were exhausted, then the plan will fail with an error message something like:

```
Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: Failed to list collectors
│ 
│   with grafana_fleet_management_collector.collector1,
│   on main.tf line 15, in resource "grafana_fleet_management_collector" "collector1":
│   15:   resource "grafana_fleet_management_collector" "collector1" {
│ 
│ unavailable: POST http://localhost:8080/collector.v1.CollectorService/ListCollectors giving up after 4 attempt(s)
╵
╷
│ Error: Failed to list collectors
│ 
│   with grafana_fleet_management_collector.collector2,
│   on main.tf line 24, in resource "grafana_fleet_management_collector" "collector2":
│   24:   resource "grafana_fleet_management_collector" "collector2" {
│ 
│ unavailable: POST http://localhost:8080/collector.v1.CollectorService/ListCollectors giving up after 4 attempt(s)
╵
╷
│ Error: Failed to list collectors
│ 
│   with grafana_fleet_management_collector.collector3,
│   on main.tf line 33, in resource "grafana_fleet_management_collector" "collector3":
│   33:   resource "grafana_fleet_management_collector" "collector3" {
│ 
│ unavailable: POST http://localhost:8080/collector.v1.CollectorService/ListCollectors giving up after 4 attempt(s)
```

spartan0x117: The number of retries are [configurable](https://github.com/grafana/terraform-provider-grafana/blob/main/pkg/provider/framework_provider.go#L106) at the provider level and default to `3`

spartan0x117: Note this error is a bit different from what would happen today if one of the collectors were to fail repeatedly:

```
Terraform planned the following actions, but then encountered a problem:

  # grafana_fleet_management_collector.collector1 will be updated in-place
  ~ resource "grafana_fleet_management_collector" "collector1" {
      ~ enabled           = true -> false
        id                = "test-collector-1"
        # (1 unchanged attribute hidden)
    }

  # grafana_fleet_management_collector.collector3 will be updated in-place
  ~ resource "grafana_fleet_management_collector" "collector3" {
      ~ enabled           = true -> false
        id                = "test-collector-3"
        # (1 unchanged attribute hidden)
    }

Plan: 0 to add, 2 to change, 0 to destroy.
╷
│ Error: Failed to get collector
│ 
│   with grafana_fleet_management_collector.collector2,
│   on main.tf line 24, in resource "grafana_fleet_management_collector" "collector2":
│   24:   resource "grafana_fleet_management_collector" "collector2" {
│ 
│ unavailable: POST http://localhost:8080/collector.v1.CollectorService/GetCollector giving up after 4 attempt(s)
```

internal/resources/fleetmanagement/resource_collector.go
spartan0x117: Good catch, we should reset the cache in ImportState. I'll add the reset!

internal/resources/fleetmanagement/resource_collector.go
johannaojeling: That's great. Thanks!

johannaojeling (approved):
LGTM! 🚀

PR #2492, opened 2026-01-07
Update Fleet Management collector model to support OTel collectors

markobachvarovski:
This PR adds collector typing to Terraform, so now users can choose to specify whether they want to create alloy or otel collectors

### DIFF
--- a/internal/resources/fleetmanagement/model_collector.go
+++ b/internal/resources/fleetmanagement/model_collector.go
@@ -14,6 +14,7 @@ type collectorDataSourceModel struct {
 	RemoteAttributes types.Map    `tfsdk:"remote_attributes"`
 	LocalAttributes  types.Map    `tfsdk:"local_attributes"`
 	Enabled          types.Bool   `tfsdk:"enabled"`
+	CollectorType    types.String `tfsdk:"collector_type"`
 }
 
 type collectorDataSourcesModel struct {
@@ -24,6 +25,7 @@ type collectorResourceModel struct {
 	ID               types.String `tfsdk:"id"`
 	RemoteAttributes types.Map    `tfsdk:"remote_attributes"`
 	Enabled          types.Bool   `tfsdk:"enabled"`
+	CollectorType    types.String `tfsdk:"collector_type"`
 }
 
 func collectorMessageToDataSourceModel(ctx context.Context, msg *collectorv1.Collector) (*collectorDataSourceModel, diag.Diagnostics) {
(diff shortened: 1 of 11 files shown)


internal/resources/fleetmanagement/model_collector.go
```diff
@@ -68,9 +72,34 @@ func collectorResourceModelToMessage(ctx context.Context, model *collectorResour
...
 
+// collectorTypeToString converts the proto enum to a Terraform-friendly string.
+func collectorTypeToString(ct collectorv1.CollectorType) string {
+	switch ct {
+	case collectorv1.CollectorType_COLLECTOR_TYPE_ALLOY:
+		return "ALLOY"
+	case collectorv1.CollectorType_COLLECTOR_TYPE_OTEL:
+		return "OTEL"
+	default:
+		return ""
```
tpaschalis, line 87: What is the correct behavior for this (and the next) helper? Further above we say that `Defaults to ALLOY if not specified`.

tpaschalis: Or is the `Default` field in the collectorResource enough?

tpaschalis (approved):
LGTM % the question about the default field.

tpaschalis:
FYI a PR to fix the CI issues was merged, you might have to rebase from main before you run `cloud acceptance tests`.

internal/resources/fleetmanagement/model_collector.go
markobachvarovski: >Or is the Default field in the collectorResource enough?

Yes, that's right, the `Default` [here](https://github.com/grafana/terraform-provider-grafana/pull/2492/changes/BASE..c3a95a5c55adde473a9a773e2224d61fa91307a4#diff-5db815bd552539f2a7a00b6ce5b8e6d81ff3f1f13d5eebd144af1380d2d5e5c5R113) will set CollectorType = ALLOY if it's not explicitly provided. I ran tests with CollectorType set explicitly and left it empty, and both worked as designed, so in normal operation, we should never get to an empty string or `COLLECTOR_TYPE_UNSPECIFIED` as the output of these helpers

internal/resources/fleetmanagement/resource_collector.go
```diff
@@ -103,6 +106,15 @@ Required access policy scopes:
 				Computed: true,
 				Default:  booldefault.StaticBool(true),
 			},
+			"collector_type": schema.StringAttribute{
+				Description: "Type of the collector. Must be one of: ALLOY, OTEL. Defaults to ALLOY if not specified.",
```
johannaojeling, line 112: Should we mention that Fleet Management for OTel Collectors is in private preview and that users should contact customer support to get it enabled?

markobachvarovski: We're adding that to the docs :)

PR #2500, opened 2026-01-15
feat: add secrets manager resources

PoorlyDefinedBehaviour:


internal/common/secretsapi/client.go
```diff
@@ -0,0 +1,244 @@
...
+		defaultHeaders: defaultHeaders,
+		namespace:      namespace,
+	}, nil
+}
+
+func (c *Client) Namespace() string {
+	return c.namespace
+}
+
+type ObjectMetadata struct {
```
macabu: TBH dont know but could it make sense to import the grafana/grafana/apps/secret for the types here?

examples/resources/grafana_secure_value/resource.tf
```diff
@@ -0,0 +1,21 @@
+resource "grafana_apps_secret_securevalue_v1beta1" "db_password" {
+  metadata {
+    uid = "db-password"
```
macabu, line 3: I don't know where to comment this but is it standard for the k8s apps to use `uid` instead of `name` for this field?

internal/resources/appplatform/secret_keeper_resource.go
```diff
@@ -0,0 +1,255 @@
...
+				SpecBlocks: map[string]schema.Block{
+					"aws": schema.SingleNestedBlock{
+						Description: "AWS Secrets Manager configuration.",
+						Attributes: map[string]schema.Attribute{
+							"region": schema.StringAttribute{
+								Required:    true,
+								Description: "AWS region.",
+							},
+						},
+						Blocks: map[string]schema.Block{
```
macabu, line 57: We won't support the access key + secret key pair initially here?

macabu (approved):
Just a few questions

internal/resources/appplatform/secret_keeper_resource.go
PoorlyDefinedBehaviour: No because the api does't implement it yet

examples/resources/grafana_secure_value/resource.tf
PoorlyDefinedBehaviour: From what I've seen, that is the case. Other resources in the resources/appplatform folder use the `uid` field to set the name and that's why I did the same

internal/common/secretsapi/client.go
PoorlyDefinedBehaviour: Trying it

PoorlyDefinedBehaviour: That works

internal/resources/appplatform/secret_secure_value_resource.go
```diff
@@ -453,18 +460,22 @@ func parseSecureValueSpec(ctx context.Context, src types.Object) (SecureValueSpe
 		spec.Decrypters = decrypters
 	}
 
+	var rawValue string
 	if !data.Value.IsNull() && !data.Value.IsUnknown() {
-		spec.Value = data.Value.ValueString()
+		rawValue = data.Value.ValueString()
+		value := v1beta1.SecureValueExposedSecureValue(rawValue)
```
macabu: I am surprised this works, but maybe I am missing something.

We are implementing MarshalJSON on this type to "[redacted]", which TBH I think we could remove to relax the restriction.

https://github.com/grafana/grafana/blob/main/apps/secret/pkg/apis/secret/v1beta1/exposed_secure_value.go#L62-L65

PoorlyDefinedBehaviour: done

internal/resources/appplatform/resource.go
```diff
@@ -539,7 +595,7 @@ func SaveResourceToModel[T sdkresource.Object](
 		}
 	}
 
-	dst.ID = meta.UUID
+	dst.ID = meta.UID
```
radiohead: This is actually supposed to be UUID, since the `ID` is supposed to be a server-generated field.

PoorlyDefinedBehaviour: undid these changes

radiohead (approved):

PR #2511, opened 2026-01-19
feat: [KG] Add Knowledge Graph trace configuration resource

vpadi:
Issues: https://github.com/grafana/asserts-adi/issues/5149

## What

This adds support for managing Knowledge Graph trace integration configurations, enabling users to programmatically configure the trace drilldown view for their entities.

## Why 
  Currently, users must manually configure trace drilldown mappings through the Grafana UI. This resource allows teams to:
  - Define trace configurations as code alongside their infrastructure
  - Version control their trace drilldown settings
  - Ensure consistent trace configurations across environments
  - Automate the setup of entity property to trace label mappings

  The resource follows the same patterns as the existing `grafana_asserts_log_config` resource and supports the Asserts API's trace configuration endpoints.


matt-metivier (approved):
LGTM, the public docs will need to be updated to align with these changes

Issue #870, opened 2023-04-04
Allow grafana_rule_group to have config_json option 


novy4:
Grafana provider has config_json option to import data from JSON file for [grafana_dashboard](https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/dashboard) resource:

```
resource "grafana_dashboard" "metrics" {
  config_json = file("grafana-dashboard.json")
}
```

Would be nice to have exactly the same config_json option for alert [rule_group](https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/rule_group). At this moment its a bit annoying to process JSON for alerts in TF code.

Grafana API is allowing to import JSON, then why config_json is not present for alerts as for dashboards?

bursztyn-pl:
I know, this is not a solution you want, but maybe this will help you?

```
locals {
  alertrule_files = fileset(path.module, "../alertrules/*.json")
}

resource "grafana_rule_group" "alertrule" {
  provider = grafana.asdf
  org_id = 1
  for_each = { for file in local.alertrule_files : file => jsondecode(file("${path.module}/${file}")) }
  name = replace(each.value.ruleGroup, "/","|") # for now terraform gets 404 response when slash is in name.  See #883 
  folder_uid = each.value.folderUID
  interval_seconds = 60
  rule {
    name = each.value.title
    for = each.value.for
    condition = each.value.condition
    no_data_state = each.value.noDataState
    annotations = each.value.annotations
    exec_err_state = each.value.execErrState
    is_paused = each.value.isPaused
    labels = each.value.labels
    dynamic "data" {
      for_each = each.value.data
      content {
        datasource_uid = data.value.datasourceUid
        ref_id = data.value.refId
        relative_time_range {
          from = data.value.relativeTimeRange.from
          to = data.value.relativeTimeRange.to
        }
        query_type = data.value.queryType
        model = jsonencode(data.value.model)
      }
    }
  }
}
```

novy4:
> I know, this is not a solution you want, but maybe this will help you?
> 
> ```
> locals {
>   alertrule_files = fileset(path.module, "../alertrules/*.json")
> }
> 
> resource "grafana_rule_group" "alertrule" {
>   provider = grafana.asdf
>   org_id = 1
>   for_each = { for file in local.alertrule_files : file => jsondecode(file("${path.module}/${file}")) }
[...]

PR #2516, opened 2026-01-20
feat: Add grafana_alerting_rule resource for JSON-based alert management

jakubno:
Introduces a new grafana_alerting_rule resource that allows managing Grafana alerting rules using raw JSON configuration. This provides an alternative to the existing grafana_rule_group resource with more flexible control over individual alert rules.

Key features:
- Manage individual alert rules via JSON configuration (config_json)
- Support for disable_provenance flag to allow external modifications
- Full CRUD operations (Create, Read, Update, Delete)
- Import support for existing rules
- Automatic normalization of Grafana-managed fields

Implementation details:
- NormalizeConfigJSON removes server-managed fields (id, orgID, updated, provenance)
- Normalizes relativeTimeRange.to when set to 0 (Grafana default)
- Removes empty relativeTimeRange objects automatically added by Grafana
- Normalizes duration fields to minute format (e.g., "60s" not "1m0s")
- CustomizeDiff forces new resource when rule UID changes

The resource ensures stable terraform plans by normalizing default values while preserving drift detection for intentional changes made outside of Terraform.


jakubno:
Should resolve #870

jakubno:
Tagging @titolins as requested

examples/resources/grafana_alerting_rule/import.sh
titolins, line 1: is this being used?

jakubno:
Fixed the tests, forgot to also normalize the expected output

titolins (commented):
I hadn't realized the proposal in the issue. I thought the idea was to introduce a new resource for alert groups, but with json support.

So there are actually 2 things being proposed here right?

1. `config_json` - Allow the import of alert rules via raw json
2. Alert rule support - Support management of alert rules independently of rule groups (which is also being requested in #1507)

Regarding 2., I'm not sure if that's by design tbh. Alert rules depend on group, so it might be the case that support for this wasn't added intentionally. I'm reaching out to the team to understand if that's the case or not.

Regarding 1., I'd be against having this implementation if that's the only way to create alert rule resources. If we're implementing a new resource, I believe it should have proper support for hcl as well. And perhaps we could take advantage of that for validating jsons as well. If we could tie those 2 together, I don't see a reason for not accepting. But let me first understand 2. properly and get back to you 👍

jakubno:
Thanks for the quick responses — appreciated.

> I hadn’t realized the proposal in the issue. I thought the idea was to introduce a new resource for alert groups, but with json support.

Just to clarify: would your preference be to extend the existing [rule_group](https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/rule_group)￼ resource with a `config_json` field, rather than introducing a new resource? I was taking inspiration from [grafana_dashboard](https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/dashboard)￼, which supports raw JSON.

I agree with the concern around losing HCL-level validation. The primary motivation for JSON support on my side is enabling generation of alerting configuration via the [Grafana Foundation SDK](https://github.com/grafana/grafana-foundation-sdk)￼. 

Happy to adjust the approach once there’s clarity around whether independent alert rule management or extending rule group is the way we want to take.

Looking forward to your findings on (2).

jakubno:
Any update here @titolins?

titolins:
Hey @jakubno , sry for the delay. I brought this up to the team and will provide a better update on the issue by the end of the day. Some context:
- The alerting apis are currently under migration to the new app platform model (using the [grafana-app-sdk](https://github.com/grafana/grafana-app-sdk)) and includes redesigning the rules API
- The provisioning API's will be marked for deprecation as soon as we have the new API's work finished (which should happen in the next few months)
- The new API is already scoped on individual rules, so the concept of groups will not be exposed anymore
  - This API is already available, but it's still under work and considered unstable
  - There's a new terraform resource for alert rule using the new API already, but it's lacking documentation (tf provider reference [here](https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/apps_rules_alertrule_v0alpha1))
- One of the advantages of using the app sdk is that we get [grafanactl](https://github.com/grafana/grafanactl) support for free
  - This allows managing supported resources in json format
- In parallel to that, there's also the gitsync work under way ([docs](https://grafana.com/docs/grafana/latest/as-code/observability-as-code/git-sync/))
  - This is not yet supported by alerting and will take longer than the new API's work stream, but it is on our backlog
- Also, usage of the provisioning alert_rule API is discouraged as that is considered tech debt: the provisioning API favors the rule group by design

Summing up: 
  - provisioning API for alert rule is considered tech debt and all provisioning API's will be marked for deprecation soon
  - the new API is scoped for individual rules
  - there will be, in the near future, other tooling allowing for json provisioning

So we'll probably not accept this and close the issue. I have a call later today with a colleague that's working directly on the new rules API and will bring this up with them to confirm this is the case. If it is, I'll close this and the respective issue (and probably the other issue I've linked in my first message as well).

PR #2521, opened 2026-01-30
perf: Improve access policies data source performance

cinaglia:
This PR improves the performance of the `grafana_cloud_access_policies` data source by:

- Parallelizing multi-region API calls (10 concurrent requests)
- Leverage server-side name filtering to reduce network traffic
- Updating documentation to clarify performance implications of the `region_filter`


dmihai (approved):
It looks very good, nice work!

PR #2522, opened 2026-02-02
refactor(cloud): migrate cloud_ips and cloud_organization data sources from sdkv2 to framework

Duologic:
- **refactor: migrate cloud IPs data source to framework SDK**
- **refactor: migrate cloud organization data source to framework SDK**

This is my first PR using Claude code as an assistant, I've read through the code and didn't spot anything odd. I also did some manual testing with this code against a live Grafana Cloud instance, both resources returned the expected data.

### DIFF
--- a/internal/resources/cloud/data_source_cloud_organization.go
+++ b/internal/resources/cloud/data_source_cloud_organization.go
@@ -4,65 +4,122 @@ import (
 	"context"
 	"strconv"
 
-	"github.com/grafana/grafana-com-public-clients/go/gcom"
 	"github.com/grafana/terraform-provider-grafana/v4/internal/common"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
+var _ datasource.DataSource = &CloudOrganizationDataSource{}
+var _ datasource.DataSourceWithConfigure = &CloudOrganizationDataSource{}
+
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 //go:generate go run ./tools/genimports examples
-//go:generate go run github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs
+//go:generate go run github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs generate --provider-name "terraform-provider-grafana"
```
Duologic, line 14: When running `make docs` from a directory that is named differently than `terraform-provider-grafana`, it'll fail. This ensure the name is set correctly.

I ran into this as I use worktrees and check out the code in directory named after the branch instead of the repo.

Duologic:
The acceptance tests for the cloudinstance can't run in parallel with other PRs, will run those when the other PRs are done.

internal/resources/cloud/data_source_cloud_organization.go
```diff
@@ -4,65 +4,122 @@ import (
...
-				Type:     schema.TypeString,
-				Computed: true,
+			"created_at": schema.StringAttribute{
+				Computed:    true,
+				Description: "The date and time the organization was created.",
 			},
-			"updated_at": {
-				Type:     schema.TypeString,
-				Computed: true,
+			"updated_at": schema.StringAttribute{
```
nachogiljaldo, line 68: Not something you introduced, but this is an odd value to add to a datasource, will it not cause constant drift if internal updates would change the field?

nachogiljaldo (approved):
Not an expert on terraform API but what I see seems 👌 . Just raised a question but definitively not a blocker since that was already there.

internal/resources/cloud/data_source_cloud_organization.go
Duologic: I can't say either, let's do that in a follow up.

PR #2527, opened 2026-02-06
refactor(slo): migrate to Terraform Plugin Framework

Duologic:
## Summary

Migrates the `grafana_slo` resource and `grafana_slos` data source from SDKv2 to Terraform Plugin Framework.

- Go slices for nested block models (not `types.List`)
- Multi-file layout: `resource_slo.go` (schema + CRUD), `resource_slo_model.go` (pack functions + resource model), `data_source_slos.go` (datasource + shared models + convert functions), `common_framework.go` (shared base structs)
- Shared model types between resource and datasource
- `ResourceWithConfigValidators` pattern for required `ListNestedBlock` validation (Framework doesn't support `Required` on blocks)
- Enrichment support preserved
- Custom `ImportState` with nil-based 404 handling
- Removed `utils.go` (SDKv2 conversion functions no longer needed)

### Key correctness fixes applied

- **Query type normalization**: API returns `"grafanaQueries"` (camelCase) but schema expects `"grafana_queries"` (snake_case) — normalized in `convertQueryToModel`
- **Empty alerting metadata**: API returns non-nil fastburn/slowburn even when unconfigured — `isAlertingMetadataEmpty()` guard treats empty ones as absent
- **Silent objective dropping**: Old code used `continue` on validation failure in `packObjectives`, silently dropping objectives — removed in favor of schema-level validators
- **Nil-based 404 handling**: Replaced string-matching on diagnostic summary with nil check from `readSLO`

## Review guide

Start with `internal/resources/slo/resources.go` to see registration (`makeResourceSlo`, `makeDatasourceSlo`).

Then review in this order:
1. **`data_source_slos.go`** — Contains all shared model types (`queryModel`, `alertingModel`, `enrichmentModel`, `objectiveModel`, etc.) and `convert*ToModel` functions (API → TF state). Also contains the datasource schema and Read implementation.
2. **`resource_slo_model.go`** — Contains `pack*` functions (TF state → API request) and the `sloResourceModel` struct. This is the resource-specific counterpart to the shared models.
3. **`resource_slo.go`** — Resource schema, CRUD operations, validators (`grafanaQuerySchemaValidator`, `requiredBlockValidator`, `nonEmptyStringValidator`), `formatAPIError`, `readSLO`.
4. **`common_framework.go`** — Shared `Configure` base structs for resource and datasource.

### Schema changes (non-breaking)

All changes are cosmetic/documentation improvements:
- Description format `markdown` → `plain`
- ID field: `optional+computed` → `computed` only
- `max_items`/`min_items` replaced by Framework list validators
- Added descriptions to `freeform` and `ratio` query blocks
- Required blocks marked with `**Required.**` prefix in descriptions (Framework workaround)

Existing Terraform configurations work without modification. Verified via `terraform providers schema -json` comparison.

## Test plan

All acceptance tests pass against Grafana Cloud dev instance:

- [x] `TestAccDataSourceSlo` — datasource read + list
- [x] `TestAccResourceSlo` — 8 steps: freeform, ratio, threshold, advanced queries, alerting, enrichments
- [x] `TestAccSLO_recreate` — external delete + automatic re-create
- [x] `TestAccResourceInvalidSlo` — invalid config rejection
- [x] `TestValidateGrafanaQuery` — 5 sub-tests for grafana query validation
- [x] `TestAccResourceSloWithCustomUUID` — custom UUID support

### Issue Ref
https://github.com/grafana/deployment_tools/issues/539315
https://github.com/grafana/deployment_tools/issues/539316


internal/resources/slo/resource_slo.go
```diff
@@ -28,1009 +35,748 @@ const (
...
-		},
-		Schema: map[string]*schema.Schema{
-			"name": {
-				Type:         schema.TypeString,
-				Required:     true,
-				Description:  `Name should be a short description of your indicator. Consider names like "API Availability"`,
-				ValidateFunc: validation.StringLenBetween(0, 128),
+`,
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
```
elainevuong, line 94: question: is it common to have both a Terraform resource ID "id" and a "uuid" attribute? Why are both needed? Just out of curiosity, not opposed to having both

internal/resources/slo/resource_slo.go
```diff
@@ -28,1009 +35,748 @@ const (
...
-				Description: `Destination Datasource sets the datasource defined for an SLO`,
-				Elem: &schema.Resource{
-					Schema: map[string]*schema.Schema{
-						"uid": {
-							Type:        schema.TypeString,
-							Description: `UID for the Datasource`,
+		},
+		Blocks: map[string]schema.Block{
+			"destination_datasource": schema.ListNestedBlock{
+				MarkdownDescription: "**Required.** Destination Datasource sets the datasource defined for an SLO.",
```
elainevuong, line 135: comment: I think we need a `Required: true` here, because the `destination_datasource` is a required block.

internal/resources/slo/resource_slo.go
```diff
@@ -28,1009 +35,748 @@ const (
...
-				Required:    true,
-				Description: `Query describes the indicator that will be measured against the objective. Freeform Query types are currently supported.`,
-				Elem: &schema.Resource{
-					Schema: map[string]*schema.Schema{
-						"type": {
-							Type:         schema.TypeString,
-							Description:  `Query type must be one of: "freeform", "query", "ratio", "grafana_queries" or "threshold"`,
-							ValidateFunc: validation.StringInSlice([]string{"freeform", "query", "ratio", "threshold", "grafana_queries"}, false),
-							Required:     true,
+			"query": schema.ListNestedBlock{
```
elainevuong, line 152: comment: I think we also need a `Required: true` here for the `query` block as well

internal/resources/slo/resource_slo.go
Duologic: Every terraform resource needs an explicit `id` field, for this resource the value of the UUID was chosen in the past. This refactor maintains that behavior.

internal/resources/slo/resource_slo.go
```diff
@@ -28,1009 +35,748 @@ const (
...
+	apiReq := apiClient.DefaultAPI.V1SloPost(ctx).SloV00Slo(sloModel)
+	response, _, err := apiReq.Execute()
 	if err != nil {
-		return apiError("Unable to create SLO - API", err)
+		resp.Diagnostics.AddError("Unable to create SLO", formatAPIError(err))
+		return
 	}
 
-	// Note: UUID after a POST is always returned; this can be a vanity UUID or a random UUID
-	d.SetId(response.Uuid)
```
Duologic, line 410: Old uuid -> id mapping.

internal/resources/slo/resource_slo.go
```diff
@@ -28,1009 +35,748 @@ const (
...
-				Required:    true,
-				Description: `Over each rolling time window, the remaining error budget will be calculated, and separate alerts can be generated for each time window based on the SLO burn rate or remaining error budget.`,
-				Elem: &schema.Resource{
-					Schema: map[string]*schema.Schema{
-						"value": {
-							Type:         schema.TypeFloat,
-							Required:     true,
-							ValidateFunc: validation.FloatBetween(0, 1),
-							Description:  `Value between 0 and 1. If the value of the query is above the objective, the SLO is met.`,
+			"objectives": schema.ListNestedBlock{
```
elainevuong, line 226: comment: missing `Required: true` on the `objectives` top level schema

internal/resources/slo/resource_slo.go
```diff
@@ -28,1009 +35,748 @@ const (
...
-							Elem:        keyvalueSchema,
+				Validators: []validator.List{
+					listvalidator.SizeAtMost(1),
+				},
+				NestedObject: schema.NestedBlockObject{
+					Blocks: map[string]schema.Block{
+						"label": schema.ListNestedBlock{
+							MarkdownDescription: "Labels will be attached to all alerts generated by any of these rules.",
+							NestedObject: schema.NestedBlockObject{
+								Attributes: map[string]schema.Attribute{
```
elainevuong: question: we previously used the label key/value type in three places - at the top-level label, within the labels on the alerting block, and within the annotations on the alerting block. It's also used in the labels and annotations that sit on both the fastburn and slowburn blocks within the alerting block.

we previously had this defined in one place - in `keyvalueSchema`, which was re-used in the alerting-labels and alerting-annotations block.

```
			"label": {
				Type:        schema.TypeList,
				Optional:    true,
				Description: `Additional labels that will be attached to all metrics generated from the query. These labels are useful for grouping SLOs in dashboard views that you create by hand. Labels must adhere to Prometheus label name schema - "^[a-zA-Z_][a-zA-Z0-9_]*$"`,
				Elem:        keyvalueSchema,
			},
```

```
var keyvalueSchema = &schema.Resource{
		"key": {
			Required:    true,
			Description: `Key for filtering and identification`,
		},
		"value": {
			Required:    true,
			Description: `Templatable value`,
		},
	},
}
```

Is this no longer doable with the new framework?

internal/resources/slo/resource_slo.go
Duologic: In the Plugin Framework, `ListNestedBlock` does not support a `Required` field — that is an SDKv2-only concept. The equivalent is the `listvalidator.SizeAtLeast(1)` validator on lines 117-118, which will fail the plan if the block is omitted. This is the standard pattern for required blocks in Framework resources across the provider.

internal/resources/slo/resource_slo.go
Duologic: Same as above — `ListNestedBlock` in the Plugin Framework has no `Required` field. The `listvalidator.SizeAtLeast(1)` on lines 134-136 enforces that at least one `query` block must be present, serving the same purpose.

internal/resources/slo/resource_slo.go
Duologic: Same here — the `listvalidator.SizeAtLeast(1)` on lines 222-224 enforces that at least one `objectives` block is present. Plugin Framework blocks do not have a `Required` property; validators are the intended mechanism.

internal/resources/slo/resource_slo.go
```diff
@@ -28,1009 +35,748 @@ const (
...
-		diag.Diagnostic{
-			Severity: diag.Error,
-			Summary:  action,
-			Detail:   detail,
-		},
+	return detail
+}
+
+// ValidatePrometheusWindow validates that window is a valid Prometheus duration
+func ValidatePrometheusWindow(value string) error {
```
elainevuong: comment: I think this function isn't needed and isn't used anywhere. It was likely used for validating the `window` field on `objectives`, but looks like that's now covered off by regex validation directly on the schema.

internal/resources/slo/data_source_slos.go
```diff
@@ -6,263 +6,633 @@ import (
...
+
+	alerting := alertingModel{
+		Label:      convertLabelsToModel(apiAlerting.Labels),
+		Annotation: convertLabelsToModel(apiAlerting.Annotations),
+	}
 
-	if metaData.Annotations != nil {
-		retAnnotations := unpackLabels(metaData.Annotations)
-		labelsAnnotsStruct["annotation"] = retAnnotations
+	// Convert FastBurn — treat API-returned empty metadata as absent to avoid phantom blocks
```
elainevuong, line 504: thanks for fixing this bug!! 🙏

elainevuong (approved):
LGTM, some minor nits. Built the provider and tested this against all our query types and it's working as expected. Thank you!

internal/resources/slo/resource_slo.go
Duologic: Good point. In the Plugin Framework, `schema.ListNestedBlock` is a value type (struct) rather than a pointer like `*schema.Resource` in SDKv2, so you can't share a single variable by reference. However, extracting a helper function like `keyValueNestedBlock(description string) schema.ListNestedBlock` achieves the same thing — call it at each of the 7 sites instead of duplicating the struct literal. That would cut ~50 lines and keep the definition in one place. I'll do that.

PR #2529, opened 2026-02-06
feat: add alertmanager_uid support for contact points and notification policies

### LINKED ISSUE #799: Support different alertmanager in resources like grafana_contact_point and grafana_notification_policy
In the current version of provider during creating grafana_contact_point and grafana_notification_policy they will be created in default alertmanager Grafana. But if you have **ngalertmanager** alertmanger you couldn't choose it. This functionality available in the web gui but not in provider.

darkweaver87:
## Summary

- Adds `alertmanager_uid` attribute to `grafana_contact_point` and `grafana_notification_policy` resources
- Enables managing alerting resources on external alertmanagers (e.g., `grafanacloud-ngalertmanager`) via the AM Config API
- Automatically detects Grafana-managed vs native alertmanager format and converts contact point settings accordingly

Closes #799

## Changes

- **New AM Config API client** (`alerting_amconfig_api.go`): Testable HTTP client for the `/api/alertmanager/{recipient}/config/api/v1/alerts` endpoint
- **Conversion helpers** (`alerting_amconfig_helpers.go`): Functions to convert between Grafana and native Alertmanager formats (routes, receivers, matchers)
- **Contact point support**: CRUD operations via AM Config API when `alertmanager_uid` is set
- **Notification policy support**: CRUD operations via AM Config API when `alertmanager_uid` is set
- **Field mapping**: Handles differences between Grafana and native AM (e.g., `og_priority` ↔ `priority`, `teams` ↔ `msteams_configs`)
- **Grafana-only settings filtered**: Settings like `overridePriority`, `sendTagsAs` are excluded for native AMs

### DIFF
--- a/docs/resources/contact_point.md
+++ b/docs/resources/contact_point.md
@@ -43,6 +43,7 @@ resource "grafana_contact_point" "my_contact_point" {
 ### Optional
 
 - `alertmanager` (Block Set) A contact point that sends notifications to other Alertmanager instances. (see [below for nested schema](#nestedblock--alertmanager))
+- `alertmanager_uid` (String) The UID of the Alertmanager to manage contact points for. When set, uses the Alertmanager Config API instead of the provisioning API. This allows managing contact points on external alertmanagers (e.g., `grafanacloud-ngalertmanager`).
 - `dingding` (Block Set) A contact point that sends notifications to DingDing. (see [below for nested schema](#nestedblock--dingding))
 - `disable_provenance` (Boolean) Allow modifying the contact point from other sources than Terraform or the Grafana API. Defaults to `false`.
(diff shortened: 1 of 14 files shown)


titolins:
Thank you for the contribution and the effort that went into this. It's clear a lot of work was put into it.
After internal discussion, we've decided not to merge this PR. The Grafana Terraform provider is intentionally scoped to managing resources that live within Grafana itself. Using it as a proxy to configure external alertmanagers conflicts with a deliberate design boundary we maintain. We've pushed back on similar patterns in the past, for example managing alerting rules on external datasources via Grafana's proxy APIs.
Beyond the design concern, accepting this would create a Terraform dependency on Grafana's DMA proxy APIs, which would make it significantly harder to deprecate or remove those APIs in the future.
We'll follow up in the issue with more context on the current direction and what this means going forward.

PR #2530, opened 2026-02-12
Migrate grafana_team to use Framework SDK

suntala:
### **Review Guide**

This PR migrates `grafana_team` from SDKv2 to the Plugin Framework and fixes a latent shared-instance bug in the provider's Plugin Framework resource factory.

  **`resources.go` (grafana package)** — trivial rename only (`resourceTeam` → `makeResourceTeam`).

  **`resource_team.go`** — standard Framework CRUD rewrite. Notable points:
  - `preferences` and `team_sync` use `schema.ListNestedBlock` / `schema.SetNestedBlock` (not nested attributes) for protocol v5 mux compatibility — using nested attributes here would panic at startup
  - `members` is `Optional+Computed` so it can be unknown on the first create (e.g. when set from another resource's output); the old SDK handled this implicitly, the Framework requires an explicit `IsUnknown()` guard before calling `ElementsAs`
  - `org_id` + API key validation is added explicitly in `Create` — the SDKv2 version had this injected automatically via `grafanaOrgIDResourceValidation`; the Framework resource must enforce it inline
  - `team_id` / `team_uid` use `UseStateForUnknown` plan modifiers so they don't force replacement on every plan

  **`docs/`** — generated; run `make docs` to verify.

  ### **Issue Ref**
https://github.com/grafana/deployment_tools/issues/522297

### DIFF
--- a/docs/resources/team.md
+++ b/docs/resources/team.md
@@ -40,16 +40,13 @@ resource "grafana_team" "test-team" {
 ### Optional
 
 - `email` (String) An email address for the team.
-- `ignore_externally_synced_members` (Boolean) Ignores team members that have been added to team by [Team Sync](https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-team-sync/).
-Team Sync can be provisioned using [grafana_team_external_group resource](https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/team_external_group).
- Defaults to `true`.
-- `members` (Set of String) A set of email addresses corresponding to users who should be given membership
-to the team. Note: users specified here must already exist in Grafana.
-- `org_id` (String) The Organization ID. If not set, the Org ID defined in the provider block will be used.
-- `preferences` (Block List, Max: 1) (see [below for nested schema](#nestedblock--preferences))
-- `team_sync` (Block List, Max: 1) Sync external auth provider groups with this Grafana team. Only available in Grafana Enterprise.
-	* [Official documentation](https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-team-sync/)
-	* [HTTP API](https://grafana.com/docs/grafana/latest/developers/http_api/team_sync/) (see [below for nested schema](#nestedblock--team_sync))
+- `ignore_externally_synced_members` (Boolean) Ignores team members that have been added to team by [Team Sync](https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-team-sync/). Team Sync can be provisioned using [grafana_team_external_group resource](https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/team_external_group).
+- `members` (Set of String) A set of email addresses corresponding to users who should be given membership to the team. Note: users specified here must already exist in Grafana.
+- `org_id` (String) The Organization ID. If not set, the default organization is used for basic authentication, or the one that owns your service account for token authentication.
+- `preferences` (Block List) (see [below for nested schema](#nestedblock--preferences))
+- `team_sync` (Block List) Sync external auth provider groups with this Grafana team. Only available in Grafana Enterprise.
+* [Official documentation](https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-team-sync/)
+* [HTTP API](https://grafana.com/docs/grafana/latest/developers/http_api/team_sync/) (see [below for nested schema](#nestedblock--team_sync))
 
 ### Read-Only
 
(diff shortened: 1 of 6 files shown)


internal/resources/grafana/resource_team.go
```diff
@@ -32,432 +39,593 @@ const (
...
+		if common.IsNotFoundError(err) {
+			return nil, diags
 		}
+		diags.AddError("Failed to read team", err.Error())
+		return nil, diags
 	}
 
-	return ReadTeam(ctx, d, meta)
-}
+	emailVal := types.StringNull()
```
mihai-turdean: claude` flagged this, but do why do we need to either return a null string or a non-empty string?

> If a user sets email = "" in config, the plan will have types.StringValue("") but read() returns types.StringNull(). This mismatch would cause a perpetual diff. Consider using  types.StringValue("") instead when the API returns empty, or ensure the schema marks email as Computed too. The SDKv2 version didn't have this issue because SDKv2 treats "" and absent the same for strings.                                                                                                                                                                    
> 
> This is likely a real bug — users who explicitly set email = "" will see a plan diff on every apply.

internal/resources/grafana/resource_team.go
```diff
@@ -32,432 +39,593 @@ const (
...
-						"theme": {
-							Type:         schema.TypeString,
-							Optional:     true,
-							ValidateFunc: validation.StringInSlice([]string{"light", "dark", "system", ""}, false),
-							Description:  "The default theme for this team. Available themes are `light`, `dark`, `system`, or an empty string for the default theme.",
-							Default:      "",
+		},
+		// preferences and team_sync use Blocks (not Attributes) for protocol v5 mux compatibility.
+		Blocks: map[string]schema.Block{
+			"preferences": schema.ListNestedBlock{
```
mihai-turdean, line 150: The old schema had `MaxItems: 1`. The new `ListNestedBlock` doesn't enforce this — a user could supply multiple `preferences {}` blocks and only [0] would be used. The Plugin Framework supports `listvalidator.SizeAtMost(1)` via terraform-plugin-framework-validators. Without it, extra blocks are silently ignored.

internal/resources/grafana/resource_team.go
```diff
@@ -32,432 +39,593 @@ const (
...
-				MaxItems: 1,
-				Elem: &schema.Resource{
-					Schema: map[string]*schema.Schema{
-						"groups": {
-							Type:     schema.TypeSet,
-							Optional: true,
-							Elem: &schema.Schema{
-								Type: schema.TypeString,
-							},
+			"team_sync": schema.ListNestedBlock{
```
mihai-turdean, line 186: The old schema had `MaxItems: 1`. The new `ListNestedBlock` doesn't enforce this — a user could supply multiple `preferences {}` blocks and only [0] would be used. The Plugin Framework supports `listvalidator.SizeAtMost(1)` via terraform-plugin-framework-validators. Without it, extra blocks are silently ignored.

internal/resources/grafana/resource_team.go
rooneyshuman: Updating so email is computed too

internal/resources/grafana/resource_team.go
rooneyshuman: Adding a validator to enforce max of 1 as was done in the old SDKv2

internal/resources/grafana/resource_team.go
rooneyshuman: same as above

docs/resources/team.md
```diff
@@ -40,16 +40,13 @@ resource "grafana_team" "test-team" {
 ### Optional
 
 - `email` (String) An email address for the team.
```
mihai-turdean, line 42: Docs may need regenerating after the `email` attribute was changed to `Optional+Computed` — the current diff still shows `email (String)` without a `Computed` marker.

internal/resources/grafana/resource_team.go
```diff
@@ -32,432 +40,597 @@ const (
...
-			WeekStart:        d.Get("preferences.0.week_start").(string),
+	// Preferences
+	prefsResp, err := client.Teams.GetTeamPreferences(teamIDStr)
+	if err != nil {
+		diags.AddError("Failed to read team preferences", err.Error())
+		return nil, diags
+	}
+	prefs := prefsResp.GetPayload()
+	if prefs.Theme != "" || prefs.Timezone != "" || prefs.HomeDashboardUID != "" || prefs.WeekStart != "" {
+		nullIfEmpty := func(s string) types.String {
```
mihai-turdean: The `nullIfEmpty` helper here can cause a plan/state mismatch for explicitly-set empty strings. If a user writes `theme = ""` (valid per the `OneOf` validator), the plan will contain `types.StringValue("")`, but `read()` will return `types.StringNull()` for that field — causing a perpetual diff.

Consider using `types.StringValue(s)` unconditionally inside the preferences block instead of `nullIfEmpty`, since `""` is a valid configured value for all four fields.

rooneyshuman: updated to remove the helper and use `types.StringValue(s)` instead

docs/resources/team.md
rooneyshuman: docs are up to date - tried to regen with no changes. There's also a unit test in the CI suite that checks for docs being outdated which is passing. The email field appears under `Optional` in `docs/resources/team.md`, which is the correct section for Optional+Computed attributes in `tfplugindocs`.

mihai-turdean (approved):

PR #2343, opened 2025-09-09
fix(oncall): make labels/dynamic_labels order-insensitive

sidharth-chauhan:
This PR fixes [#2169](https://github.com/grafana/terraform-provider-grafana/issues/2169).

The `labels` and `dynamic_labels` fields for `grafana_oncall_integration` were previously treated as lists, making Terraform detect diffs when only the order of labels changed. Since the OnCall API does not treat label order as significant, this caused unnecessary perma-diffs.

### Changes
- Updated schema definitions for `labels` and `dynamic_labels` to use `TypeSet` instead of `TypeList`.
- Adjusted expand/flatten helpers for labels to support the new schema.
- Updated acceptance tests to validate label handling.
- Updated documentation examples to reflect the new `labels` and `dynamic_labels` format.

### Outcome
Terraform will now treat `labels` and `dynamic_labels` as order-insensitive, preventing diffs when only their order changes.

### Notes
- All local unit tests pass.
- Acceptance tests skipped locally (no Grafana Cloud credentials available). These will run in CI.

PR #2536, opened 2026-02-17
fix(oncall): make integration labels idempotent

### LINKED ISSUE #2169: Oncall integration labels order creates perma-diff
### Terraform Version

OpenTofu v1.8.8

provider registry.opentofu.org/grafana/grafana v3.25.0

### Grafana Version

Grafana Cloud

### Affected Resource(s)

* grafana_oncall_integration

```hcl
data "grafana_oncall_label" "source" {
  key = "source"
  value = "test"
}

data "grafana_oncall_label" "severity" {
  key = "severity"
  value = "warning"
}

data "grafana_oncall_label" "customer" {
  key = "customer"
  value = "acme"
}

resource "grafana_oncall_integration" "test_integration" {
  provider = grafana.oncall
  name     = "TEST INTEGRATION"
  type     = "webhook"

  default_route {
  }

  labels = [
    data.grafana_oncall_label.source,
    data.grafana_oncall_label.severity,
    data.grafana_oncall_label.customer,
  ]
}
```

### Expected Behavior

Since the API does not treat the order of labels as significant the terraform provider should not either.

Changing the order of labels in terraform should not result in any diff.

### Actual Behavior

The terraform provider seems to treat the order as significant resulting in a perma-diff when the order of labels is changed in terraform.

### Steps to Reproduce

1. Apply the test terraform config above
2. Reorder the labels, eg
  ```
  labels = [
    data.grafana_oncall_label.source,
    data.grafana_oncall_label.customer,
    data.grafana_oncall_label.severity,
  ]
  ```
3. Apply the reordered labels
4. Perform another terraform plan to see that there is still a diff. Example output below.

### Important Factoids

_No response_

### References

Oncall integration labels added to the provider in #2125 

When managing Oncall integration labels via Terraform the order of labels is treated a significant by the provider however it seems that the order is not significant to the API. This can result in perma-diff when every plan and apply results in the same diff. The apply does not seem to actually make any difference to the API. If the order is not significant to the API there should not have been any diff in the first place.

We were managing the labels via the UI however since label support was added to the terraform provider we migrated our config to terraform for consistency (although there is still missing functionality for managing integrations via terraform #1581).

Example of the perma-diff

```
  # grafana_oncall_integration.test_integration will be updated in-place
  ~ resource "grafana_oncall_integration" "test_integration" {
        id     = "CIAEFVAYPK4QV"
      ~ labels = [
            {
                "id"    = "source"
                "key"   = "source"
                "value" = "test"
            },
          - {
              - "id"    = "severity"
              - "key"   = "severity"
              - "value" = "warning"
            },
            {
                "id"    = "customer"
                "key"   = "customer"
                "value" = "acme"
            },
          + {
              + "id"    = "severity"
              + "key"   = "severity"
              + "value" = "warning"
            },
        ]
        name   = "TEST INTEGRATION"
        # (2 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }
```

### Work around

It is possible to work around this issue by removing the labels from an integration and re-adding them however the difference in behavior between the API and provider should still be considered a bug.

darkweaver87:
# Motivation

This PR fixes https://github.com/grafana/terraform-provider-grafana/issues/2169 and is not breaking as https://github.com/grafana/terraform-provider-grafana/pull/2343 is.

The `labels` and `dynamic_labels` attributes caused spurious diffs on
every plan due to two issues:

1. flattenLabels() injected a synthetic "id" field (set to the label
   key name) that is never present in user configuration.

2. TypeList compares elements positionally, but the API may return
   labels in a different order than the config, causing labels to
   appear swapped or removed.

Fix by adding a DiffSuppressFunc that:
- Suppresses the synthetic "id" and map-size diffs from old state
- Compares labels as a set of {key, value} pairs, ignoring order

Also remove the synthetic "id" from flattenLabels() so new state
no longer includes it.

### DIFF
--- a/internal/resources/oncall/resource_integration.go
+++ b/internal/resources/oncall/resource_integration.go
@@ -237,8 +237,9 @@ func resourceIntegration() *common.Resource {
 						Type: schema.TypeString,
 					},
 				},
-				Optional:    true,
-				Description: "A list of string-to-string mappings for static labels. Each map must include one key named \"key\" and one key named \"value\" (using the `grafana_oncall_label` datasource).",
+				Optional:         true,
+				DiffSuppressFunc: labelsDiffSuppress,
+				Description:      "A list of string-to-string mappings for static labels. Each map must include one key named \"key\" and one key named \"value\" (using the `grafana_oncall_label` datasource).",
 			},
 			"dynamic_labels": {
 				Type: schema.TypeList,
@@ -248,8 +249,9 @@ func resourceIntegration() *common.Resource {
 						Type: schema.TypeString,
 					},
 				},
-				Optional:    true,
-				Description: "A list of string-to-string mappings for dynamic labels. Each map must include one key named \"key\" and one key named \"value\" (using the `grafana_oncall_label` datasource).",
+				Optional:         true,
+				DiffSuppressFunc: labelsDiffSuppress,
+				Description:      "A list of string-to-string mappings for dynamic labels. Each map must include one key named \"key\" and one key named \"value\" (using the `grafana_oncall_label` datasource).",
 			},
 		},
 	}
@@ -795,6 +797,9 @@ func expandDefaultRoute(input []any) *onCallAPI.DefaultRoute {
 	defaultRoute := onCallAPI.DefaultRoute{}
 
 	for _, r := range input {
+		if r == nil {
+			continue
+		}
 		inputMap := r.(map[string]any)
 		id := inputMap["id"].(string)
 		defaultRoute.ID = id
@@ -825,6 +830,9 @@ func expandLabels(input []any) []*onCallAPI.Label {
 	labelsData := make([]*onCallAPI.Label, 0, 1)
 
 	for _, r := range input {
+		if r == nil {
+			continue
+		}
 		inputMap := r.(map[string]any)
 		key, keyExists := inputMap["key"]
 		value, valueExists := inputMap["value"]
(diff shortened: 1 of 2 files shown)


internal/resources/oncall/resource_integration.go
```diff
@@ -841,15 +849,62 @@ func expandLabels(input []any) []*onCallAPI.Label {
...
 }
+
+// labelsDiffSuppress suppresses spurious diffs on the labels and dynamic_labels attributes.
+// It handles two cases:
+//  1. The synthetic "id" field that flattenLabels previously stored in state.
+//  2. Label ordering: the API may return labels in a different order than the config, but TypeList
+//     compares positionally. We suppress the diff when the set of {key,value} pairs is identical.
+func labelsDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
+	// Suppress the synthetic "id" field that was previously stored in state
+	if strings.HasSuffix(k, ".id") {
```
gerardocorea, line 874: A small note about how this is only needed during the initial transition period to avoid displaying a diff showing the removal of the id field would be good for removal.

gerardocorea (changes requested):
Overall change looks good, tested it locally and order is ignored for labels and dynamic labels as well as the unused id field. Thanks for this.

internal/resources/oncall/resource_integration.go
darkweaver87: Thanks @gerardocorea for your review. Is that [comment](https://github.com/grafana/terraform-provider-grafana/pull/2536/changes/3b3de9c42b1c6227ab5686868b11d10efdc682e0) fine ?

gerardocorea: Looks good, thank you.

willgallego-grafana (approved):
Hey @darkweaver87 thanks for your patience on a review and for cutting the PR as well. This looks good and is a needed boost to some more stable TF code for the repo. Appreciate the help!

gerardocorea (approved):
lgtm

PR #2537, opened 2026-02-17
Add secure values support for App Platform resources

ArturWierzbicki:
This PR introduces a shared secure-values framework for app platform resources. 

It adds a write-only secure block with name and create inputs, uses `secure_version` to control when secrets are applied or rotated and centralizes secure-aware update/removal behavior in the generic resource flow. The follow-up [git-sync-resources PR](https://github.com/grafana/terraform-provider-grafana/pull/2539) is the first consumer of this framework


radiohead:
This comment has been generated by Claude Code

## `secure_version` is effectively required when using `secure {}` — worth making this clearer

`secure_version` is declared `Optional` but not `Computed`. This creates a usability gap: a user can create a resource with values in the `secure {}` block and never set `secure_version`. If they later want to rotate a secret (or change which keys are configured), they have no way to trigger re-application without also making an unrelated change to a `spec` field.

A few options worth considering:
1. Make `secure_version` `Required` when `SecureValueAttributes` is configured (schema-level enforcement)
2. Keep it `Optional` but strengthen the docs — explicitly call out that `secure_version` must be set for any future secure rotation to work reliably
3. Consider `Computed: true` with a hash of the secure key names, so at minimum the presence/absence of keys triggers an update automatically

The current state means a user who omits `secure_version = 1` on first apply is effectively stuck — no Terraform-only path to trigger re-application later.

radiohead (approved):

spinillos (approved):

PR #2538, opened 2026-02-18
fix: sort issue with frontend_o11y_app

ChrisSyme:
## What
This PR fixes an issue where `terraform plan` incorrectly detects changes in `grafana_frontend_o11y_app` resource's `allowed_origins` attribute even when no actual changes were made aside from changing the order.

## Why
The `allowed_origins` list was not being sorted when reading from the API, causing Terraform to detect spurious diffs when the API returned origins in a different order between runs. This resulted in unnecessary plan changes and potential resource updates even though the actual set of origins remained unchanged.

## Changes
- Sort `allowed_origins` alphabetically in `convertClientModelToTFModel` to ensure consistent ordering
- Add unit test to verify origins are always sorted regardless of API response order

## Example Issue

Before this fix, running `terraform plan` multiple times could show changes like:

```hcl
  ~ resource "grafana_frontend_o11y_app" "example" {
      ~ allowed_origins = [
          - "https://z.example.com",
          - "https://a.example.com",
          + "https://a.example.com",
          + "https://z.example.com",
        ]
    }
```

After this fix, `terraform plan` correctly shows no changes when origins haven't actually changed.

## Testing
Added `TestConvertClientModelToTFModel_OriginsSorted` to verify that origins are consistently sorted alphabetically regardless of the order returned by the API.


t00mas (approved):

ChrisSyme:
@t00mas thanks for approving. Could you please trigger the pending checks as this is restricted to maintainers?

mar4uk:
acceptance tests are failing, probably should be fixed in https://github.com/grafana/terraform-provider-grafana/pull/2550

Duologic:
That is not related to linked PR, it's because this is a contribution from outside grafana, it doesn't have access to those  credentials. Let me see what I can do.

internal/resources/frontendo11y/models_test.go
```diff
@@ -0,0 +1,47 @@
...
+		},
+	}
+
+	result, diags := convertClientModelToTFModel(123, app)
+	if diags.HasError() {
+		t.Fatalf("unexpected diagnostics: %v", diags)
+	}
+
+	var origins []types.String
+	diags = result.AllowedOrigins.ElementsAs(nil, &origins, false)
```
mar4uk, line 27: could you please pass context.TODO here instead of nil?
The golangci-lint is [throwing](https://github.com/grafana/terraform-provider-grafana/actions/runs/22403022244/job/64854785961?pr=2551) an error for this line:

> internal/resources/frontendo11y/models_test.go:27:43: SA1012: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (staticcheck)
	diags = result.AllowedOrigins.ElementsAs(nil, &origins, false)

mar4uk (changes requested):

PR #2541, opened 2026-02-18
fix(syntheticmonitoring): add mutex to prevent concurrent API request issues

saas-zach-dunton:
## Summary
- Add mutex to serialize Synthetic Monitoring API operations
- Fixes silent failures when deleting multiple `grafana_synthetic_monitoring_check_alerts` resources concurrently

## Problem
The Synthetic Monitoring API doesn't handle concurrent requests well. When multiple resources are deleted simultaneously (e.g., during `terraform destroy`), Terraform reports success for all deletions, but only some actually succeed. The remaining resources are silently left as orphans in the API.

## Root Cause
When Terraform tries to delete multiple `grafana_synthetic_monitoring_check_alerts` resources in parallel, the concurrent `PUT /check/{id}/alerts` requests to the SM API interfere with each other. The API returns success (200 OK) for all requests, but only processes some of them.

## Solution
Add a mutex to serialize all SM API operations, similar to the existing mutexes for alerting (`alertingMutex`), folder (`folderMutex`), and dashboard (`dashboardMutex`) resources.

## Testing
Tested with 11 concurrent deletions:
- **Without fix**: 6 of 11 deletions silently failed (alerts remained in API despite Terraform reporting "destroyed")
- **With fix**: All 11 deletions succeeded correctly

## Test plan
- [x] Build passes
- [x] Manual testing with multiple concurrent create/delete operations
- [x] Verified fix resolves the issue
- [x] Verified issue exists without the fix

### DIFF
--- a/internal/common/client.go
+++ b/internal/common/client.go
@@ -53,6 +53,7 @@ type Client struct {
 	alertingMutex  sync.Mutex
 	folderMutex    sync.Mutex
 	dashboardMutex sync.Mutex
+	smapiMutex     sync.Mutex
 }
 
 // WithAlertingMutex is a helper function that wraps a CRUD Terraform function with a mutex.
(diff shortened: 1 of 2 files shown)


ka3de:
Thank you for the submission.
As a first step we should reproduce the issue and verify that it's actually an issue in the API, and discard any other possible issue, for example in the terraform provider. If the issue is confirmed to be in the API, IMO we should prioritize first fixing it in the API, not through serialization of requests from the terraform provider. Otherwise users that interact with the HTTP API directly will face the same issue.

Could you provide a specific example of resources you are creating and destroying?

ka3de:
We were able to reproduce the bug and we have created an issue to handle this from the Synthetic Monitoring side.
A deeper review of the implementation will follow-up, but we have an initial fix that should solve this for the Terraform provider. We'll get back to you when that change is deployed to production so you can confirm from your side @saas-zach-dunton.

ka3de:
@saas-zach-dunton we have deployed an initial fix that should prevent orphaned alert thresholds.
Could you check from your side? A more thoughtful review will follow up in order to improve concurrent check and alert requests.

saas-zach-dunton:
This seems resolved now. Thanks!

PR #2546, opened 2026-02-24
Migrate grafana_annotation to use Framework SDK

suntala:
For https://github.com/grafana/terraform-provider-grafana/issues/2534

This PR updates the grafana_annotation resource to use the Framework SDK. More details can be found in the issue and parent issue. Platform Monitoring is migrating resources to use Framework SDK instead of Terraform Plugin SDKv2 since support for SDKv2 is ending. 

The resource is reimplemented from the legacy SDK (schema + CreateContext/ReadContext/…) to the Framework. Added: `annotationResource` implementing `resource.Resource`, typed `resourceAnnotationModel`, and Framework schema with plan modifiers and validators. 

Optional attributes (tags, dashboard_uid, panel_id) are normalized in the read method so that when the API returns empty or zero, state is set to null, matching an unset config and avoiding “Provider produced inconsistent result after apply.” RFC3339 validation for time and time_end is preserved via a custom Framework string validator. The resource stays registered as before in resources.go via `makeResourceAnnotation()` and continues to use the org-scoped lister.


cindy (commented):
This looks good to me 👍

cbodonnell (approved):
LGTM :)

PR #2556, opened 2026-02-26
grafana_dashboard: write to apis/ when passing k8s shape

harisrozajac:
With https://github.com/grafana/grafana/pull/118946 we allow passing k8s resource using grafana_dashboard

This PR adds redirection logic in the provider itself, so we hit apis/ endpoint directly


internal/resources/appplatform/dashboard_k8s_for_legacy_resource.go
```diff
@@ -0,0 +1,259 @@
...
+
+	"github.com/grafana/grafana-app-sdk/resource"
+	"github.com/grafana/grafana/apps/dashboard/pkg/apis/dashboard/v1beta1"
+	"github.com/grafana/grafana/pkg/apimachinery/utils"
+	"github.com/grafana/terraform-provider-grafana/v4/internal/common"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+)
+
+const (
+	k8sDashboardIDPrefix   = "v1beta1:"
```
ryantxu: ideally we can make this agnostic to the exact version, and just the standard "k8s.io/client-go/dynamic" with an unstructured request

We can likely copy most of https://github.com/grafana/grafana/blob/main/pkg/api/dashboard.go#L509 for that

internal/resources/grafana/resource_dashboard.go
```diff
@@ -145,7 +153,33 @@ func CreateDashboard(ctx context.Context, d *schema.ResourceData, meta any) diag
...
+	d.SetId(id)
+	normalized := NormalizeDashboardConfigJSON(outJSON)
+	d.Set("config_json", normalized)
+	return readDashboardK8s(ctx, d, meta)
+}
+
 func ReadDashboard(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	if _, _, ok := appplatform.ParseK8sDashboardID(d.Id()); ok {
+		tflog.Debug(ctx, "grafana_dashboard: state id is v1beta1, using App Platform API for read")
+		return readDashboardK8s(ctx, d, meta)
```
ryantxu: is this required?   If it is, the "fix" in https://github.com/grafana/grafana/pull/118946 may just be a more delayed error 😬

harisrozajac: i've made the updates so that client is version agnostic and ReadDashboard now always reads from /api/dashboards/uid/{uid}

PR #2558, opened 2026-02-26
fix(slo): improve error messaging for invalid SLO Terraform configuration

### LINKED ISSUE #2557: [SLO] Clarify Error Messaging for Invalid SLO Terraform Configuration
Latest -  4.25.0 

### Grafana Version
Latest

### Affected Resource(s)
*grafana_slo

### Steps to Reproduce
See escalation. When invalid fields are specified on the SLO Terraform resource, better error messaging should be provided to the customer to enable them to troubleshoot the root cause of the issue.

It is anticipated that the incidence should decrease as SLO has implemented an Export to HCL functionality with the UI. 

### References

https://github.com/grafana/support-escalations/issues/20872

elainevuong:
### Summary
                                                                        
This PR improves error messaging in the SLO resource packing logic to surface actionable, user-facing errors rather than runtime panic when a user's Terraform configuration is malformed or unexpected. See escalation [here](https://github.com/grafana/support-escalations/issues/20872). 

Normally these fields would be enforced by the Terraform schema (i.e. `type: schema.TypeString`, `required: true` etc), but the user was using a var.tf file; when the Terraform validation runs, it sees a valid reference to a variable, but when it starts to execute the `apply` is when issues occur. 

Changes:
- Replaced unsafe type assertions with guarded assertions that return descriptive errors, across all pack* functions: `packSloResource`, `packQuery`, `packObjectives`, `packLabels`, `packAlerting`, `packAlertMetadata`, `packDestinationDatasource`
- Updated pack* functions to return actionable error messages 
- Error messages now include the field path (e.g. `query.ratio.success_metric`, `destination_datasource.uid`) and informs the user to validate their Terraform configuration
- Added a test case (sloEmptyDestinationDatasourceUID) that verifies an empty `destination_datasource.uid` produces the expected error message

Fixes #2557


elainevuong:
Testing Video - for context

https://github.com/user-attachments/assets/f93dd751-2412-4e44-81b1-d68bcdcd8e27

internal/resources/slo/resource_slo.go
```diff
@@ -626,38 +697,60 @@ func packQuery(query map[string]any) (slo.SloV00Query, error) {
...
 
-func packObjectives(tfobjectives []any) []slo.SloV00Objective {
+func packObjectives(tfobjectives []any) ([]slo.SloV00Objective, error) {
 	objectives := []slo.SloV00Objective{}
 
 	for ind := range tfobjectives {
-		tfobjective := tfobjectives[ind].(map[string]any)
+		tfobjective, ok := tfobjectives[ind].(map[string]any)
+		if !ok {
+			return nil, fmt.Errorf("objectives[ind]: unexpected type - check your Terraform configuration")
```
cesararmada: Do we want to include the literal "objectives[ind]" in the message, or the actual index ?  If we want to use the index we should use `objectives[%d]` instead.

internal/resources/slo/resource_slo.go
```diff
@@ -626,38 +697,60 @@ func packQuery(query map[string]any) (slo.SloV00Query, error) {
...
 
-func packLabels(tfLabels []any) []slo.SloV00Label {
+func packLabels(tfLabels []any) ([]slo.SloV00Label, error) {
 	labelSlice := []slo.SloV00Label{}
 
 	for ind := range tfLabels {
-		currLabel := tfLabels[ind].(map[string]any)
+		currLabel, ok := tfLabels[ind].(map[string]any)
+		if !ok {
+			return nil, fmt.Errorf("label[ind]: unexpected type — check your Terraform configuration")
```
cesararmada: Same question here and other logs that are using a literal string.

cesararmada (approved):
LGTM, just some questions to check before merge.

PR #2569, opened 2026-03-06
feat(codegen): generate resources from cue schemas

spinillos:
It adds a new function to be able to generate the resource from CUE schema. 

It uses cog for the generator but its still WIP.

---

Manifests are parsed with github.com/grafana/grafana-app-sdk for consistency, basic types (TF models, types and attributes) are generated with cog. A custom codegen template generates the boilerplate code required to define [`NamedResources`](https://github.com/grafana/terraform-provider-grafana/blob/1aafb4d7ccbad1f410881ebe18c1d804b51fbbf8/internal/resources/appplatform/resource.go#L154)

The generated output should always be reviewed and customized (to improve the user's experience, add validation, deal with secret values, …)


scripts/add-resource/codegen/jenny_go_generator.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// We're finally ready to do some codegen!
+	files, err := cog.TypesFromSchema().
+		CUEValue(
+			filepath.Base(jenny.outputDir),
+			v,
+			cog.ForceEnvelope(envelopeName),
+			cog.NameFunc(nameFunc),
+		).
+		SchemaTransformations(cog.PrefixObjectsNames(jenny.kindName)).
```
konsalex, line 105: Maybe I miss context here, but wouldn't this conflict with other Kinds of a different version? Shouldn't the prefix include a version in the identifier?

scripts/add-resource/dev/main.go
```diff
@@ -0,0 +1,63 @@
...
+	manifestProperties := manifest.Properties()
+	config := codegen.Config{
+		PackageValue:      &kindInfo.Schema,
+		KindName:          kindInfo.Kind,
+		PluralName:        kindInfo.PluralName,
+		Version:           version,
+		AppName:           manifestProperties.AppName,
+		GroupOverride:     manifestProperties.FullGroup,
+		GoTypesImportPath: goTypesImportPath,
+		ResourceOutputDir: "appplatform",
```
konsalex, line 51: One issue caught by Claude:
```
Only appplatform is a valid output directory. The CLI accepts any directory, while the templates hardcode package appplatform_test and reference App Platform symbols unqualified. Other selections generate uncompilable packages. Hardcode/validate the directory or parameterize the templates.
```

PR #2570, opened 2026-03-09
refactor(grafana_user): migrate to plugin framework

suntala:
Fixes https://github.com/grafana/deployment_tools/issues/507740

Migrates grafana_user from SDKv2 to the Plugin Framework (global client, ModifyPlan for case-insensitive email/login, lister uses org 0, and admin flag updates only when is_admin changes). 

Hardens acceptance tests and CI by avoiding process-wide GRAFANA_AUTH clashes: explicit provider blocks via ConfigWithBasicAuthProvider / ConfigWithTokenProvider, a simplified ProtoV5ProviderFactories, orgScopedTest returning tokens, and OSS GRAFANA_BASIC_AUTH in acc workflows; Docker Compose sets admin credentials explicitly. 

Also extends config generation for Framework resources (replace_null_sensitive, broader “missing required” plan errors) and tightens parseAuth so tokens with : are not parsed as basic auth.

[Thread](https://raintank-corp.slack.com/archives/C06PXDT9Y85/p1775722360267019), [thread2](https://raintank-corp.slack.com/archives/C06PNGH5H28/p1776093865515959)

### Equivalence test

To compare the migrated provider against the `grafana_user` goldens (requires https://github.com/grafana/terraform-provider-grafana/pull/2871 for `EQUIV_FILTERS`):

```sh
git fetch origin suntala/eqtest-grafana_user
git checkout origin/suntala/eqtest-grafana_user -- equivalence-tests/tests/grafana_user equivalence-tests/goldens/grafana_user
make equivalence-test-diff-local EQUIV_FILTERS=grafana_user
```

Exit `0` = pass.


pkg/generate/testdata/generate/dashboard/resources.tf
```diff
@@ -43,5 +43,5 @@ resource "grafana_user" "admin" {
   email    = "admin@localhost"
   is_admin = true
   login    = "admin"
-  password = "SENSITIVE_VALUE_TO_REPLACE"
+  password = "SENSITIVE_VALUE_TO_REPLACE" # sensitive
```
suntala, line 46: See https://github.com/grafana/terraform-provider-grafana/pull/1715

The legacy provider’s “unsensitive” mode exists so Terraform doesn’t over-redact (e.g. drop whole blocks) when generating config; making attributes non-sensitive during generate achieves that. The fact that generated config had no # sensitive comment seems to be a side effect of that, not the main goal—Terraform only adds the comment when the provider reports an attribute as sensitive.

mem (commented):
LGTM

internal/resources/grafana/resource_user.go
```diff
@@ -9,198 +9,347 @@ import (
...
 	}
 	if _, err = client.Users.UpdateUser(id, &u); err != nil {
-		return diag.FromErr(err)
+		resp.Diagnostics.AddError("Failed to update user", err.Error())
+		return
 	}
-	if d.HasChange("password") {
-		f := models.AdminUpdateUserPasswordForm{Password: models.Password(d.Get("password").(string))}
+
+	if !plan.Password.IsNull() && !plan.Password.IsUnknown() && plan.Password.ValueString() != "" {
```
colin-stuart: I think this will trigger a password change on every request (since password is Required in the schema). Instead of `IsNull` we can probably make the check similar to the admin check below, and use `!plan.Password.Equal(state.Password)`

PR #2573, opened 2026-03-10
Migrate grafana_playlist to use Framework SDK

suntala:
For https://github.com/grafana/deployment_tools/issues/507742

Migrates the grafana_playlist resource from the legacy Terraform Plugin SDKv2 to the Plugin Framework. The resource now uses Framework types and CRUD, keeps org-scoped IDs (org_id/uid), and uses the existing lister. Playlist items are represented as a set of objects (id, order, type, value). No API or Terraform config changes; only the provider implementation is updated.

[Thread](https://raintank-corp.slack.com/archives/C04RVCAG9B5/p1774028329866709)

### DIFF
--- a/docs/resources/playlist.md
+++ b/docs/resources/playlist.md
@@ -40,26 +40,32 @@ resource "grafana_playlist" "test" {
 ### Required
 
 - `interval` (String)
-- `item` (Set of Object) (see [below for nested schema](#nestedatt--item))
 - `name` (String) The name of the playlist.
 
 ### Optional
 
+- `item` (Block Set) (see [below for nested schema](#nestedblock--item))
 - `org_id` (String) The Organization ID. If not set, the Org ID defined in the provider block will be used.
 
 ### Read-Only
 
 - `id` (String) The ID of this resource.
 
-<a id="nestedatt--item"></a>
+<a id="nestedblock--item"></a>
 ### Nested Schema for `item`
 
 Required:
 
-- `id` (String)
 - `order` (Number)
+
+Optional:
+
 - `type` (String)
 - `value` (String)
+
+Read-Only:
+
+- `id` (String)
 
 ## Import
 
--- a/docs/resources/playlist.md
+++ b/docs/resources/playlist.md
@@ -3,10 +3,13 @@
 page_title: "grafana_playlist Resource - terraform-provider-grafana"
 subcategory: "Grafana OSS"
 description: |-
+  Manages Grafana playlists.
   Official documentation https://grafana.com/docs/grafana/latest/dashboards/create-manage-playlists/HTTP API https://grafana.com/docs/grafana/latest/developers/http_api/playlist/
 ---
 
 # grafana_playlist (Resource)
+
+Manages Grafana playlists.
 
 * [Official documentation](https://grafana.com/docs/grafana/latest/dashboards/create-manage-playlists/)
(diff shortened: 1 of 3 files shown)


docs/resources/playlist.md
```diff
@@ -40,11 +40,11 @@ resource "grafana_playlist" "test" {
 ### Required
 
 - `interval` (String)
-- `item` (Block Set, Min: 1) (see [below for nested schema](#nestedblock--item))
 - `name` (String) The name of the playlist.
 
 ### Optional
 
+- `item` (Block Set) (see [below for nested schema](#nestedblock--item))
```
suntala, line 50: A validator on item ([see](https://github.com/grafana/terraform-provider-grafana/pull/2573/changes#diff-f398641dc1c6eeee65d0fb0b4865c3acd1345bd0bc466690d65916bb1718c3bcR126)) ensures that it has at least one item.

internal/resources/grafana/resource_playlist.go
```diff
@@ -10,198 +10,411 @@ import (
...
+
+	createResp, err := client.Playlists.CreatePlaylist(&cmd)
+	if err != nil {
+		resp.Diagnostics.AddError("Error creating playlist", err.Error())
+		return
+	}
+
+	id := createResp.Payload.UID
+	if id == "" {
+		id = strconv.FormatInt(createResp.Payload.ID, 10)
```
spinillos, line 184: You are trying to convert empty id here.

internal/resources/grafana/resource_playlist.go
```diff
@@ -10,198 +10,411 @@ import (
...
+}
 
+func (r *playlistResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
+	var plan playlistResourceModel
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	client, _, split, parseErr := r.clientFromExistingOrgResource(resourcePlaylistID, plan.ID.ValueString())
```
spinillos: You are doing the same in `Update`, `ReadPlaylist` and `ImportState`. Couldn't be better extract this check into a function?

internal/resources/grafana/resource_playlist.go
suntala: Thanks for the review.

Seems to me that this matched legacy behavior. See legacy code: https://github.com/grafana/terraform-provider-grafana/pull/2573/changes#diff-f398641dc1c6eeee65d0fb0b4865c3acd1345bd0bc466690d65916bb1718c3bcL110-L114

However, I did go ahead and tighten up the logic to only use the numeric id when it's non-zero: https://github.com/grafana/terraform-provider-grafana/pull/2573/changes/9b4beef617af7e3d4d4e6c154f0ecd7b27c81dbe#diff-f398641dc1c6eeee65d0fb0b4865c3acd1345bd0bc466690d65916bb1718c3bcR181-R187

internal/resources/grafana/resource_playlist.go
suntala: Yes, makes sense to reduce repetition. Change made in [9b4beef](https://github.com/grafana/terraform-provider-grafana/pull/2573/commits/9b4beef617af7e3d4d4e6c154f0ecd7b27c81dbe)

spinillos (approved):

PR #2577, opened 2026-03-11
CI: Add SDKv2 usage check

suntala:
For https://github.com/grafana/deployment_tools/issues/511486

Adds a CI check that warns when a PR or push to main adds new SDKv2 resource or datasource registration in internal/resources/. The job stays green but posts a PR comment and a workflow warning with migration guidance and links. The plan is to switch this to a hard failure in April. AGENTS.md is updated to describe the check and the PR comment.

Steps:
* Send out comms (#terraform-provider? #engineering?):
  * Remind about previous comms if there are any.
  * Provide context about the migration taking place.
  * Immediately there will be CI warning for new resources/datasources using SDKv2
  * In April it will be switched to hard failure.
* Merge this PR
* In April switch from exit 0 to exit 1 (hard failure) and remove the Action to comment on PRs as it will no longer be necessary.

Feedback on the [comms draft](https://docs.google.com/document/d/1_kNp2DLZvE1wNOy5IID1XtI6OWBG3hjI9JnNk8veUb8/edit?tab=t.u0hgisc7kowa#bookmark=id.3pcukujal0q6) would be great as well.

[Thread](https://raintank-corp.slack.com/archives/C02S85R5064/p1773309102259809)


.github/workflows/sdkv2-migration-check.yml
```diff
@@ -0,0 +1,73 @@
+name: SDKv2 migration check
+on:
+  pull_request: {}
+  push:
+    branches:
+      - main
```
cindy, line 6: Does that mean we're also running this job when something is merged to main?

cindy (approved):

.github/workflows/sdkv2-migration-check.yml
suntala: Yes, that's right. WDYT?

suntala: Updated the PR description just now to include this info.

cindy: Sounds good. One thing about the hard failure. I'm assuming if we switch to a hard check, any file modified will also fail?

.github/workflows/sdkv2-migration-check.yml
```diff
@@ -0,0 +1,73 @@
...
+            echo "sdkv2_detected=true" >> "$GITHUB_OUTPUT"
+            {
+              echo '**SDKv2 migration notice**'
+              echo ''
+              echo 'This PR adds new SDKv2 (legacy) resource or datasource registration. This codebase is migrating to the Terraform Plugin Framework; **new resources and data sources must use the Framework**, not SDKv2.'
+              echo ''
+              echo '- Use **`common.NewResource`** (with a type implementing `resource.ResourceWithConfigure`) for new resources.'
+              echo '- Use the Framework datasource pattern for new data sources (not `NewLegacySDKDataSource`).'
+              echo ''
+              echo 'See **AGENTS.md** (Three-Layer Resource Architecture) for details. See https://github.com/grafana/deployment_tools/issues/475444 and https://github.com/grafana/terraform-provider-grafana/issues/2216 for more context and examples of how to use the Plugin Framework.'
```
cindy: I created an overview (non-monitoring) issue for the migration. Will add issues for the remaining resources and datasources there. 
```suggestion
              echo 'See **AGENTS.md** (Three-Layer Resource Architecture) for details. See https://github.com/grafana/terraform-provider-grafana/issues/2580 and https://github.com/grafana/terraform-provider-grafana/issues/2216 for more context and examples of how to use the Plugin Framework.'
```

Duologic (approved):

.github/workflows/sdkv2-migration-check.yml
suntala: I hope I understand your question correctly--this should only fail if new resources/datasources are added with SDKv2. Modifications to unmigrated resources/datasources should not result in failure.

.github/workflows/sdkv2-migration-check.yml
suntala: Nice! I'll add the issue you created alongside the others.

PR #2578, opened 2026-03-11
feat: Add allow_ui_updates option for v1 and v2 dashboard resources

harisrozajac:
For file provisioning of dashboards, we have `allowUiUpdates` property, which makes it possible to edit the v2 provisioned dashboard in the UI. This PR supports this option in TF for v1 and v2 dashboards resource.

Requested [here](https://raintank-corp.slack.com/archives/C03KVDHTWAH/p1773227060770389). 

Also introduced a generic `OptionsAttributes` field on `ResourceSpecSchema` so individual AppPlatform resources can extend the options block without affecting other resources, following the same pattern used by `SpecAttributes`/`SpecBlocks`

Demo

https://github.com/user-attachments/assets/1b1d2a8f-58c2-40e9-a9e6-e69fb0dbe855

Example config for testing:

```terraform
resource "grafana_apps_dashboard_dashboard_v2beta1" "test-" {
 metadata {
   uid = "tr-v2"
 }

 spec {
   json = file("${path.module}/test-dash-v2.json")
 }

 options {
   overwrite = true
   allow_ui_updates = true
 }
}
```


spinillos:
The CI is failing because:
* You have to define `allow_ui_updates` inside ResourceOptionsModel with `tfsdk` tag.
* Execute `make docs` to update documentation.

harisrozajac:
> You have to define allow_ui_updates inside ResourceOptionsModel with tfsdk tag.

@spinillos I did that initially 🙂 but that affected all app platform resources - so when i run `make docs`, `allow_ui_updates` will show as an option in alerting and other docs. With the current approach, only dashboard docs are updated

I've updated my approach to have a generic `OptionsAttributes` field on `ResourceSpecSchema` so individual AppPlatform resources can extend the options block without affecting other resources, following the same pattern used by `SpecAttributes`/`SpecBlocks`. This is the commit https://github.com/grafana/terraform-provider-grafana/pull/2578/commits/0b019addb3356253c3e4435a2fa2653848c0c154 wdyt?

also something is flaky with `make linkcheck` - it looks like it's because of too many requests when it hits those urls, but it passed after i rerun it 2 times

internal/resources/appplatform/resource.go
```diff
@@ -90,6 +86,9 @@ type SecureValueAttribute struct {
 	Description         string
 	MarkdownDescription string
 	DeprecationMessage  string
+	SpecAttributes      map[string]schema.Attribute
+	SpecBlocks          map[string]schema.Block
+	OptionsAttributes   map[string]schema.Attribute
```
radiohead: Looks like these were added by mistake?

radiohead (commented):
I think this is a really weird feature to request, but that's OK. Overall LGTM, just a few extra fields that I think are not necessary?

internal/resources/appplatform/resource.go
harisrozajac: ah yes, this was added while resolving a merge conflict. i removed them

harisrozajac:
> I think this is a really weird feature to request, but that's OK

@radiohead I agree, do you think it's worth updating the docs saying that allow_ui_updates will cause a drift?

radiohead (approved):

radiohead:
> > I think this is a really weird feature to request, but that's OK
> 
> @radiohead I agree, do you think it's worth updating the docs saying that allow_ui_updates will cause a drift?

Yeah, I think that'd be a good idea. Maybe ask docs team? Otherwise this looks good, thank you – approved!

markd-forcepoint:
As someone who is after this change, here is the reason - in case there is a better way to do things.
Currently, I have a set of json dashboards in a git repo, that can be deployed to multiple folders and grafana instances, in order to allow a production deployment, a development deployment and multiple dev branch deployments. The tooling modifies folders, links and ids to allow this.
This allows a workflow where a developer can have a full set of dashboards for multiple parts of a large service on a branch deployment in a branch named folder. They can then modify the dashboards, export the json and save it in a repo. CI then redeploys in a PR folder while in PR and then deploys in the main dev folder when merged. After deploying successfully in dev, the dashboards can then be deployed in the production instance.
So, for the branch deploys, I want to be able to edit in the UI, so that the json can be exported (NB the v2 save to json when trying to edit when you are not allowed is not the same as export). For the main dev and production deployments, I will probably have the allow edit off.

internal/resources/appplatform/dashboard_resource.go
```diff
@@ -51,6 +51,12 @@ Manages Grafana dashboards using the new Grafana APIs.
 						ElementType: types.StringType,
 					},
 				},
+				OptionsAttributes: map[string]schema.Attribute{
+					"allow_ui_updates": schema.BoolAttribute{
+						Optional:    true,
+						Description: "Set to true to allow editing the resource from the Grafana UI. By default, resources managed by Terraform cannot be edited in the UI.",
```
harisrozajac: @imatwawana

internal/resources/appplatform/dashboard_v2_resource.go
```diff
@@ -51,6 +51,12 @@ Manages Grafana dashboards using the v2beta1 (Dynamic Dashboards) schema.
 						ElementType: types.StringType,
 					},
 				},
+				OptionsAttributes: map[string]schema.Attribute{
+					"allow_ui_updates": schema.BoolAttribute{
+						Optional:    true,
+						Description: "Set to true to allow editing the resource from the Grafana UI. By default, resources managed by Terraform cannot be edited in the UI.",
```
harisrozajac: @imatwawana

internal/resources/appplatform/dashboard_resource.go
imatwawana:
```suggestion
						Description: "Set to true to allow editing the resource from the Grafana UI. By default, resources managed by Terraform cannot be edited in the UI. Enabling this option can cause divergence between the Terraform configuration and the resource in Grafana.",
```

internal/resources/appplatform/dashboard_v2_resource.go
imatwawana:
```suggestion
						Description: "Set to true to allow editing the resource from the Grafana UI. By default, resources managed by Terraform cannot be edited in the UI. Enabling this option can cause divergence between the Terraform configuration and the resource in Grafana.",
```

internal/resources/appplatform/dashboard_resource.go
harisrozajac: thank you! do you mind if i switch "can cause" to "will cause"? 😄

harisrozajac:
@markd-forcepoint thank you for adding more detail! we've updated the docs so potential users will be aware that using `allow_ui_updates` will cause a drift between Terraform and Grafana.

PR #2600, opened 2026-03-12
Migrate grafana_service_account to use Framework SDK 

suntala:
For https://github.com/grafana/deployment_tools/issues/507745

Migrates the grafana_service_account resource from the legacy Terraform Plugin SDKv2 to the Plugin Framework. The resource now uses Framework types and CRUD, with org_id/id composite IDs and the existing lister. The grafana_service_account datasource uses an explicit schema because the resource no longer exposes .Schema. Includes schema fix for is_disabled (Optional + Computed + default).

[Thread](https://raintank-corp.slack.com/archives/C06PXDT9Y85/p1775473454629649)


internal/resources/grafana/resource_service_account.go
```diff
@@ -124,66 +184,161 @@ func CreateServiceAccount(ctx context.Context, d *schema.ResourceData, meta any)
...
+	}
+	if len(split) == 0 {
+		resp.Diagnostics.AddError("Invalid resource ID", "Resource ID has no parts")
+		return
+	}
+	idInt, ok := split[0].(int64)
+	if !ok {
+		resp.Diagnostics.AddError("Invalid resource ID", "Service account ID is not an integer")
+		return
+	}
```
mgyongyosi: nit: It seems that this is the same ~10 lines (calling clientFromExistingOrgResource, validate split length, and type-assert split) in `Delete` and `read` methods.

Might be worth extracting it to a function.

mgyongyosi (approved):
LGTM, left a suggestion. Thank you!

internal/resources/grafana/resource_service_account.go
suntala: Thanks for the feedback. Addressed in [aea4585](https://github.com/grafana/terraform-provider-grafana/pull/2600/commits/aea4585559f9137b9188369b5de7d52755d8d0a7)

PR #2608, opened 2026-03-18
Migrate grafana_folder_permission to Plugin Framework

rooneyshuman:
### **Review Guide**
  This PR migrates `grafana_folder_permission` from SDKv2 to the Plugin Framework and extracts
  shared bulk-permission infrastructure for reuse.

  **`resources.go`** - trivial rename only (`resourceFolderPermission` →
  `makeResourceFolderPermission`).

  **`common_resource_permission_sdk2.go`** - `updateResourcePermissions` now delegates to the new
   `setResourcePermissions` helper instead of containing the diff logic inline. Behavior is
  identical.

  **`common_resource_permission.go`** - three things added:
  - `bulkPermissionItemModel`: typed model for a single permission entry
  - `bulkPermissionsSchemaAttribute()`: returns `schema.SetNestedBlock` - **not**
  `SetNestedAttribute`, because the mux provider's protocol v5 adapter doesn't support nested
  attributes inside attributes; using the wrong type would panic at startup
  - `setResourcePermissions()`: the extracted diff function - compare against the deleted code in
   `_sdk2.go` to confirm parity; logic reads current managed+non-inherited permissions, emits
  removals for anything not in desired, emits additions for anything not in current, then calls
  the API once

  **`common_plugin_framework.go`** - adds `stripOrgScopedIDPlanModifier` on `team_id`/`user_id`.
  Needed because `grafana_team.id` returns `"orgID:localID"` while `grafana_user.id` returns a
  plain ID - without normalization this causes "inconsistent result after apply". The modifier
  strips the org prefix unconditionally on every plan.

  **`resource_folder_permission.go`** - standard Framework CRUD rewrite. Notable:
  - `folder_uid` validator: was `folderUIDValidation` func, now
  `stringvalidator.RegexMatches(^[a-zA-Z0-9\-\_]+$)` - confirm these are equivalent
  - `team_id`/`user_id` no longer default to `"0"` - this is intentional and reflected in the
  docs
  - Delete passes `nil` to `applyBulkPermissions`, clearing all managed permissions
  - `Update` uses `clientFromNewOrgResource` (reads from plan) since `folder_uid` has
  `RequiresReplace`

  **`docs/`** - generated; run `make docs` to verify.

  ### Things to double-check
  - `SplitOrgResourceID` behavior on a plain ID (no colon): confirm
  `stripOrgScopedIDPlanModifier` handles `"10"` correctly and doesn't silently zero out the value
  - `setResourcePermissions` diff logic is byte-for-byte equivalent to the deleted SDKv2 code
  - No existing configs/tests depend on `team_id`/`user_id` defaulting to `"0"`

### **Issue Ref**
https://github.com/grafana/deployment_tools/issues/520857


internal/resources/grafana/common_resource_permission.go
```diff
@@ -233,3 +236,183 @@ func (r *resourcePermissionBase) writeItem(itemID string, data *resourcePermissi
...
+// readBulkPermissions fetches the current permissions from the API.
+// team_id and user_id are returned as plain local IDs (no org prefix); the
+// stripOrgScopedIDPlanModifier ensures plan values are normalized to the same format.
+func (r *resourcePermissionBulkBase) readBulkPermissions(client *goapi.GrafanaHTTPAPI, resourceUID string) ([]bulkPermissionItemModel, diag.Diagnostics) {
+	resp, err := client.AccessControl.GetResourcePermissions(resourceUID, r.resourceType)
+	if err != nil {
+		return nil, diag.Diagnostics{diag.NewErrorDiagnostic("Failed to read permissions", err.Error())}
+	}
+
+	var items []bulkPermissionItemModel
```
mihai-turdean, line 305: Maybe initialize items as an empty list `[]` instead of `nil`. This will avoid ni vs empty list diffs.

mihai-turdean (approved):

PR #2619, opened 2026-03-19
docs(app_platform): Add overwrite deprecation message

spinillos:
`overwrite` is used only in dashboard v1beta1 resource, but its propagated to the rest of the resources.

This flag allows to do an update when we create an existent resource in Grafana, skipping the normal behaviour of Terraform.
With the exception of this resource, the rest of them have to import them into Terraform state before apply them.


internal/resources/appplatform/resource.go
```diff
@@ -926,8 +926,9 @@ func setManagerProperties(obj sdkresource.Object, clientID string, allowUIUpdate
 func (r *Resource[T, L]) optionsSchemaAttributes() map[string]schema.Attribute {
 	attrs := map[string]schema.Attribute{
 		"overwrite": schema.BoolAttribute{
-			Optional:    true,
-			Description: "Set to true if you want to overwrite existing resource with newer version, same resource title in folder or same resource uid.",
+			Optional:           true,
+			Description:        "Set to true if you want to overwrite existing resource with newer version, same resource title in folder or same resource uid.",
```
ArturWierzbicki: should we change the description to mention it works only with dashboard v1 beta?

internal/resources/appplatform/resource.go
```diff
@@ -926,8 +926,9 @@ func setManagerProperties(obj sdkresource.Object, clientID string, allowUIUpdate
 func (r *Resource[T, L]) optionsSchemaAttributes() map[string]schema.Attribute {
 	attrs := map[string]schema.Attribute{
 		"overwrite": schema.BoolAttribute{
```
ArturWierzbicki, line 928: also - what about adding this conditionally? ie we could have a hardcoded list of all existing resources to retain overwrite so that its _not_ added automatically to any other resource, cc @radiohead

radiohead: Yeah, as much of a hack it is, I think I'd be OK with that, considering it's going to be temporary anyway.

radiohead (approved):

PR #2620, opened 2026-03-19
Migrate grafana team datasource to Plugin Framework SDK

rooneyshuman:
## Review Guide
  This PR migrates the `grafana_team` data source from SDKv2 (`NewLegacySDKDataSource`) to the
  Plugin Framework, replacing the `CloneResourceSchemaForDatasource` pattern with a standalone
  typed schema.

  ### Recommended reading order

  **`data_source_team.go`** — the only logic file changed (+220/-26). Full rewrite; the old file
  was a thin wrapper around `readTeamFromID` (shared with the resource). The new file is
  self-contained. Key things to check:

  - **Schema equivalence**: the old schema was cloned from `resourceTeam().Schema` with
  `CloneResourceSchemaForDatasource`. The new schema is written out explicitly. Confirm all
  fields are present: `id`, `org_id`, `name`, `team_id`, `team_uid`, `email`, `members`,
  `read_team_sync`, `preferences`, `team_sync`. The old schema also had
  `ignore_externally_synced_members` which is explicitly set to `nil` (dropped) — this was a
  resource-only write field that has no meaning on a data source.

  - **`preferences` and `team_sync` as `ListNestedBlock`**: same mux/protocol v5 reason as #2608
  — `ListNestedAttribute` would panic at startup. These were previously `List of Object` in
  SDKv2; they're now `Block List`. Functionally equivalent, just different schema representation.

  - **Preferences population logic**: only sets `data.Preferences` if at least one field is
  non-empty (`prefs.Theme+prefs.Timezone+...!= ""`). In the old SDKv2 code this was handled by
  `flattenTeamPreferences` in the resource. Verify this is equivalent — specifically that an
  all-empty preferences response doesn't write a spurious empty block to state.

  - **`read_team_sync` default removed**: the old schema had `Default: false`. The new schema has
   no default — the check is `!data.ReadTeamSync.IsNull() && data.ReadTeamSync.ValueBool()`. This
   means if `read_team_sync` is omitted from config, `IsNull()` is true and sync is skipped.
  Equivalent behavior, but worth confirming existing configs that omit `read_team_sync` still
  work.

  - **`dsReadTeamMembersSlice`**: new private helper that replaces the shared resource helper.
  Filters out `admin@localhost` and members with labels (externally synced). Confirm this matches
   the original filtering in `readTeamFromID`.

  - **Team not found**: old code silently returned `diag.Errorf`; new code returns
  `resp.Diagnostics.AddError`. Same user-visible behavior.

  **`docs/data-sources/team.md`** — generated. The `nestedatt--` anchors became `nestedblock--`
  (reflects the schema change from attribute to block). Run `make docs` to verify.

  ### Things to double-check
  - All fields from the old cloned schema are present in the new explicit schema — particularly
  `team_uid`, which could easily be missed
  - The `ignore_externally_synced_members` field being dropped is intentional (write-only
  resource field, meaningless on a data source)
  - Preferences empty-check behavior: does the old `flattenTeamPreferences` ever write an empty
  block? If so, the new condition `!= ""` would be a behavior change
  - Members filtering in `dsReadTeamMembersSlice` matches the original resource helper exactly

### **Issue ref**
https://github.com/grafana/deployment_tools/issues/522311


internal/resources/grafana/data_source_team.go
```diff
@@ -2,55 +2,249 @@ package grafana
...
+	data.Email = types.StringValue(team.Email)
+
+	// Preferences
+	prefsResp, err := client.Teams.GetTeamPreferences(strconv.FormatInt(teamID, 10))
+	if err != nil {
+		resp.Diagnostics.AddError("Failed to read team preferences", err.Error())
+		return
+	}
+	prefs := prefsResp.GetPayload()
+	if prefs.Theme+prefs.Timezone+prefs.HomeDashboardUID+prefs.WeekStart != "" {
```
mihai-turdean: Use a more explicit check
```suggestion
	if prefs.Theme != "" || prefs.Timezone != "" || prefs.HomeDashboardUID != "" || prefs.WeekStart != "" {
```

mihai-turdean (approved):

PR #2623, opened 2026-03-20
Fix state comparison for k8s-shaped dashboards in `grafana_dashboard` resource

ArturWierzbicki:
with local k8s file, we need to compare remote dashboard json to local `dashboard.spec` rather than the whole file


radiohead (approved):

internal/resources/grafana/resource_dashboard.go
```diff
@@ -247,6 +234,76 @@ func isKubernetesStyleDashboard(dashboardJSON map[string]any) bool {
...
+func cloneDashboardJSON(dashboardJSON map[string]any) (map[string]any, error) {
+	clonedJSONBytes, err := json.Marshal(dashboardJSON)
+	if err != nil {
+		return nil, err
+	}
+	clonedDashboardJSON, err := UnmarshalDashboardConfigJSON(string(clonedJSONBytes))
+	if err != nil {
+		return nil, err
+	}
+	return clonedDashboardJSON, nil
```
spinillos, line 304:
```suggestion
	return UnmarshalDashboardConfigJSON(string(clonedJSONBytes))
```

PR #2624, opened 2026-03-20
use apiversion query param for getting dashboards in `grafana_dashboard`

ArturWierzbicki:
## Summary

 This PR updates the `grafana_dashboard` read path to use Grafana’s new optional `apiVersion` query parameter on `GET /api/dashboards/uid/:uid` when the configured `config_json` is Kubernetes-shaped.

The goal is to make refresh/read use the same dashboard API version as the user’s configured file, so Terraform compares the local Kubernetes `spec` to the correct remote dashboard body and does not show false drift due to representation mismatches.

  ## Problem

  `grafana_dashboard` can be configured with Kubernetes-shaped dashboard JSON, for example:

  ```json
  {
    "apiVersion": "dashboard.grafana.app/v2beta1",
    "kind": "Dashboard",
    "metadata": {
      "name": "example"
    },
    "spec": {
      "title": "Example"
    }
  }
```

Before this change, the provider always read dashboards with: `GET /api/dashboards/uid/:uid` without specifying the preferred dashboard API version. After https://github.com/grafana/grafana/pull/120842, callers can ask for a specific representation with: `GET /api/dashboards/uid/:uid?apiVersion=v2beta1`

Without using that query parameter, Kubernetes-shaped dashboard configs could refresh into a different representation than the configured spec, which caused false drift on terraform plan.

This PR is intentionally scoped to the read path only. For `grafana_dashboard`:

  - detect whether the configured `config_json` is Kubernetes-shaped
  - read the configured JSON from Terraform config
  - extract the version suffix from top-level apiVersion
  - call `GET /api/dashboards/uid/:uid?apiVersion=<version>` for that case
  - keep existing behavior for legacy dashboard JSON

 The provider continues comparing local `config_json.spec` to the remote dashboard body since the GET response still returns the normal outer envelope (meta + dashboard) rather than the full Kubernetes wrapper.

## Why Raw Config instead of state

This uses raw Terraform config as the source of truth for `apiVersion` so it also works when `store_dashboard_sha256 = true`. If the provider tried to recover `apiVersion` from state, that would break or become unreliable once state stores only a hash instead of the full configured dashboard JSON.

## Caveat

This provider change depends on the target Grafana instance including changes from https://github.com/grafana/grafana/pull/120842 - If a remote Grafana environment does not yet include the corresponding server-side change, the provider may still receive the older representation and the same drift may remain.

### CODE REVIEW

File: internal/resources/grafana/resource_dashboard_internal_test.go [hunk 1]
```diff
@@ -79,3 +88,107 @@ func TestNormalizeDashboardConfigJSONForState(t *testing.T) {
...
 }
+
+func TestPreferredDashboardAPIVersion(t *testing.T) {
+	t.Run("extracts version from kubernetes dashboard config", func(t *testing.T) {
+		configJSON := `{"apiVersion":"dashboard.grafana.app/v2beta1","kind":"Dashboard","metadata":{"name":"test-dashboard"},"spec":{"title":"test dashboard"}}`
+
+		got := preferredDashboardAPIVersion(configJSON)
+		if got != "v2beta1" {
+			t.Fatalf("expected v2beta1, got %q", got)
+		}
```
spinillos, line 99: Why don't use assert for tests? Same with the rest of the tests

```suggestion
assert.Equal(t, "v2beta1",  preferredDashboardAPIVersion(configJSON))
```

radiohead (approved):

spinillos (approved):
LGTM. Only a nit for tests

internal/resources/grafana/resource_dashboard.go
```diff
@@ -149,7 +154,9 @@ func ReadDashboard(ctx context.Context, d *schema.ResourceData, meta any) diag.D
 	metaClient := meta.(*common.Client)
 	client, orgID, uid := OAPIClientFromExistingOrgResource(meta, d.Id())
 
-	resp, err := client.Dashboards.GetDashboardByUID(uid)
+	preferredAPIVersion := preferredDashboardAPIVersion(getDashboardReadConfigJSON(d))
```
ArturWierzbicki, line 157: I pushed a small follow-up - 

Originally I switched from `d.Get("config_json")` to `getDashboardReadConfigJSON(d)` for the whole read path. That worked for the new apiVersion query-param logic, but it also changed the state refresh/normalization

so now, config is just for the preferred api version extraction and the rest relies on state again - 

```
  configJSON := d.Get("config_json").(string)
  configJSON, err = normalizeDashboardConfigJSONForState(configJSON, remoteDashJSON)
```

cc @spinillos @radiohead

internal/resources/grafana/resource_dashboard_internal_test.go
ArturWierzbicki: good point - will do in a followup

spinillos (approved):

PR #2635, opened 2026-03-27
Add Grafana Cloud Integrations provider

Dasomeone:
# What is this PR and why do we need it?
This PR introduces a new `grafana_cloud_integrations` provider to perform simple management of [Grafana Cloud Integrations](https://grafana.com/docs/grafana-cloud/monitor-infrastructure/integrations/)

Cloud Integrations a bit of an anti-pattern currently, as the API doesn't manage much of the actual lifecycle, and so control and configuration options are a bit more limited than normally expected.

Either way, this PR introduces this new provider based on the newer Plugin Framework, and supports the following operations:
- Installing Integrations
- Uninstalling Integrations
- Importing Integrations
- Reinstalling the integration on configuration change
- Support for both Mimir and Grafana Alerting (controlled via the integration itself, not user-selectable)

**please note, this provider does not supply an update/upgrade functionality as Cloud Integrations are "updated" by a reinstall to ensure getting a clean set of dashboards and alerts**

Unit tests have been added for some internals, alongside client interaction tests.

Heavily inspired by original work in this direction by @clementduveau in https://github.com/grafana/terraform-provider-grafana/pull/2191 (thank you! :pray:)

CC: @aditya-konarde @mariaalons @Alex3k


internal/common/resource.go
```diff
@@ -28,6 +28,7 @@ var (
 	CategoryFrontendO11y        ResourceCategory = "Frontend Observability"
 	CategoryAsserts             ResourceCategory = "Knowledge Graph"
 	CategoryK6                  ResourceCategory = "k6"
+	CategoryIntegration         ResourceCategory = "Integration"
```
mariaalons: `CategoryIntegration` was added but It's not being used

internal/resources/cloudintegrations/resource_cloud_integration.go
```diff
@@ -0,0 +1,328 @@
...
+)
+
+var (
+	resourceCloudIntegrationName = "grafana_cloud_integration"
+	resourceCloudIntegrationID   = common.NewResourceID(common.StringIDField("slug"))
+)
+
+func resourceCloudIntegration() *common.Resource {
+	return common.NewResource(
+		common.CategoryCloud,
```
mariaalons: I'm not sure about this one, but you added `CategoryIntegration`, this should be `CategoryIntegration` with subcategory `CategoryCloud`?

mariaalons (commented):
Nice work! Test it locally works as expected 👏 
Left only two comments that are related

internal/resources/cloudintegrations/resource_cloud_integration.go
Dasomeone: Ah indeed, thank you for catching these! I also hadn't updated it to be in line with `CloudIntegrations` naming as there are a couple other uses of "integration" in this codebase. Updated now

internal/common/resource.go
Dasomeone: Good catch, was outdated and not in use. Updated now :)

internal/common/cloudintegrationsapi/client.go
```diff
@@ -0,0 +1,460 @@
...
+	dashboardsResponse, err := c.PostDashboards(ctx, slug, config)
+	if err != nil {
+		return fmt.Errorf("failed to post dashboards: %w", err)
+	}
+
+	dashboardFolder := integration.Data.DashboardFolder
+	folderUID := c.generateFolderUID(dashboardFolder)
+	err = c.CreateFolder(ctx, dashboardFolder, folderUID)
+	if err != nil {
+		if !strings.Contains(err.Error(), "412") && !strings.Contains(err.Error(), "already exists") {
```
simaarfania: String matching on error can be brittle, could we check "412"  via the HTTP status code or a typed error? Consider using `errors.As` against the OpenAPI error type to check

internal/common/cloudintegrationsapi/client.go
```diff
@@ -0,0 +1,460 @@
...
+			return fmt.Errorf("failed to create folder: %w", err)
+		}
+	}
+
+	for i, dashboard := range dashboardsResponse.Data {
+		err = c.CreateDashboard(ctx, dashboard, folderUID)
+		if err != nil {
+			_ = c.DeleteFolder(ctx, folderUID)
```
simaarfania, line 205: if this folder pre-existed before installlation, deleting it here on dashboard failure would destroy data. we may want to track whether the folder was newly created vs already present before deciding to delete on rollback

internal/common/cloudintegrationsapi/client.go
```diff
@@ -0,0 +1,460 @@
...
+	if err != nil {
+		return err
+	}
+
+	folderUID := c.generateFolderUID(integration.Data.DashboardFolder)
+
+	var success bool
+	defer func() {
+		if !success {
+			_ = c.DeleteFolder(ctx, folderUID)
```
simaarfania: errors are silently swallowed here (e.g. if a permission denied on DeleteFolder or a network failure on UninstallIntegrationRules, it would be completely invisible)

simaarfania: maybe we should add a log here?

internal/common/cloudintegrationsapi/client.go
```diff
@@ -0,0 +1,460 @@
...
+// Resources are cleaned up before calling the uninstall API, matching the plugin's
+// order of operations.
+func (c *Client) UninstallIntegration(ctx context.Context, slug string) error {
+	integration, err := c.GetIntegration(ctx, slug)
+	if err != nil {
+		return fmt.Errorf("failed to get integration details: %w", err)
+	}
+
+	folderUID := c.generateFolderUID(integration.Data.DashboardFolder)
+	_ = c.DeleteFolder(ctx, folderUID)
```
simaarfania: silently discarding all errors means permission failures and API errors are invisible to the operator. we might want to consider surfacing as a warning diagnostic or at minimum logging them.

internal/common/cloudintegrationsapi/client.go
```diff
@@ -0,0 +1,460 @@
...
+	return &response, nil
+}
+
+// generateFolderUID generates a folder UID from the given dashboard_folder
+func (c *Client) generateFolderUID(folderName string) string {
+	return strings.ReplaceAll(strings.ToLower(folderName), " ", "-")
+}
+
+// CreateFolder creates a dashboard folder
+func (c *Client) CreateFolder(ctx context.Context, title, uid string) error {
```
simaarfania, line 116: OpenAPI clients support WithContext variants I believe! passing ctx here would make these respect cancellation/timeouts.

internal/common/cloudintegrationsapi/client.go
```diff
@@ -0,0 +1,460 @@
...
+const (
+	adminBasePath = "/api/plugin-proxy/grafana-easystart-app/integrations-api-admin"
+
+	grafanaCloudPromUID = "grafanacloud-prom"
+	rulesConvertAPIPath = "/api/convert/prometheus/config/v1/rules"
+
+	defaultRetries = 3
+	defaultTimeout = 90 * time.Second
+
+	RolloutLevelMimir       = 0
```
simaarfania: we may want to make these typed but i'll leave this up to you!

internal/resources/cloudintegrations/resource_cloud_integration.go
```diff
@@ -0,0 +1,331 @@
...
+	config := toAPIConfig(plan.AlertsEnabled)
+	var installErr error
+	r.commonClient.WithFolderLock(func() {
+		r.commonClient.WithDashboardLock(func() {
+			installErr = r.client.InstallIntegration(ctx, slug, config)
+		})
+	})
+	if installErr != nil {
+		resp.Diagnostics.AddError("Failed to install integration", installErr.Error())
+		resp.State.RemoveResource(ctx)
```
simaarfania: If uninstall succeeds but reinstall fails, dropping state here means Terraform forgets the resource even though there may be orphaned dashboards/folders sitting in Grafana. Maybe we should return the error and keep state instead?

internal/common/cloudintegrationsapi/client.go
Dasomeone: This is on purpose to mirror the exact behaviour taken by Grafana Cloud Integrations. These folders will be, hopefully, created as readonly in the future.

internal/resources/cloudintegrations/resource_cloud_integration.go
Dasomeone: Great catch, thanks! Have removed the state reset & added a bit more detail to the error statement

internal/common/cloudintegrationsapi/client.go
Dasomeone: Made them typed, good suggestion!

internal/common/cloudintegrationsapi/client.go
Dasomeone: It does indeed, but it's a bit hidden via the parameters! Thanks for letting me know, updated all the OpenAPI calls appropriately

internal/common/cloudintegrationsapi/client.go
Dasomeone: Added additional logs here and throughout

simaarfania (approved):

PR #2642, opened 2026-03-30
SLO: Add alert enrichment support to fastburn and slowburn alerting

cesararmada:
Add the enrichment block to fastburn and slowburn alerting metadata, allowing users to configure alert enrichments (e.g., assistantInvestigation) on SLO alert rules. Updates the slo-openapi-client dependency to include the new AlertEnrichment model.

### DIFF
--- a/docs/resources/slo.md
+++ b/docs/resources/slo.md
@@ -101,6 +101,9 @@ resource "grafana_slo" "test" {
         key   = "description"
         value = "Error budget is burning too fast"
       }
+      enrichment {
+        type = "assistantInvestigation"
+      }
     }
 
     slowburn {
@@ -111,6 +114,9 @@ resource "grafana_slo" "test" {
       annotation {
         key   = "description"
         value = "Error budget is burning too fast"
+      }
+      enrichment {
+        type = "assistantInvestigation"
       }
     }
   }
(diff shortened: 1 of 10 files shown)


docs/resources/slo.md
```diff
@@ -367,6 +377,14 @@ Required:
...
 
+<a id="nestedblock--alerting--slowburn--enrichment"></a>
+### Nested Schema for `alerting.slowburn.enrichment`
+
+Required:
+
+- `type` (String) Type of the alert enrichment. Currently only "assistantInvestigation" is supported.
+
+
 <a id="nestedblock--alerting--slowburn--label"></a>
```
elainevuong, line 394: Comment: in the docs/resources/slo.md we have three examples, you can also see them here

https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/slo

It'd be great if we could update some of the examples to show the correct usage of alertEnrichments so that users can see where it's supposed to fit on the HCL schema as a documentation examples.

elainevuong (approved):
LGTM. I built the terraform-provider and tested this against our dev instance without any issue. Minor nit on updating the examples in our documentation section to demonstrate usage of the Alert Enrichment fields

docs/resources/slo.md
cesararmada: Thanks for the suggestion, added to the examples here [c00c04a](https://github.com/grafana/terraform-provider-grafana/pull/2642/commits/c00c04aa57300eec8a60106446a35046dbbeaf35)

PR #2645, opened 2026-04-01
Add notify_next_on_call_from_schedule escalation type

musculman:
Support the newly exposed `STEP_NOTIFY_SCHEDULE_NEXT` public API step (notify_next_on_call_from_schedule) in the `grafana_oncall_escalation` resource. This type notifies the next person on-call from a schedule rather than the current one, and reuses the existing `notify_on_call_from_schedule` field to reference the schedule ID.


the-it (approved):
lgtm

PR #2651, opened 2026-04-01
fix(cloud): Update cloud wake path to grafana_com login page

janthoe:
Since the mt-fe service is now at 100% rollout, `/login` no longer wakes stacks, as those calls are proxied to that service. Instead, we can trigger an instance wake-up by selecting the `grafana_com` provider. These calls are routed directly to the stack, triggering a wake-up.


TrueWill (approved):
LGTM, but I'm not an expert.

hairyhenderson (approved):

janthoe:
after chatting with @hairyhenderson we realized that this is not necessary. Closing in favour of other solutions.

PR #2653, opened 2026-04-02
Migrate grafana_dashboard_permission to Plugin Framework

rooneyshuman:
### **Review Guide**

This PR migrates `grafana_dashboard_permission` from SDKv2 to the Plugin Framework, following the same pattern established in #2608 (`grafana_folder_permission`).

**`resources.go`** — trivial rename only (`resourceDashboardPermission` → `makeResourceDashboardPermission`).

**`resource_dashboard_permission.go`**
- Embeds `resourcePermissionBulkBase` with `dashboardsPermissionsType`, reusing the shared `readBulkPermissions` / `applyBulkPermissions` helpers from `common_resource_permission.go`
- Dashboard existence is verified via `client.Dashboards.GetDashboardByUID` in Create, Read, and ImportState (mirrors what `resourceDashboardPermissionGet` did in the SDKv2 version)
- `dashboard_uid` is now `Required` — in SDKv2 it was `Optional+Computed`, but there was no code path to derive the UID without user input; omitting it would always error at apply time. This makes the schema match reality.
  - This matches the precedent set by `folder_permissions` 
- `team_id` / `user_id` no longer default to `"0"` — intentional, consistent with `grafana_folder_permission`
- Delete passes `nil` to `applyBulkPermissions`, clearing all managed permissions

**`docs/resources/dashboard_permission.md`** — regenerated via `make docs`; no manual edits.

### **Things to double-check**
- `dashboard_uid` schema change (`Required` vs `Optional+Computed`) — confirm no existing configs rely on the field being optional
- `team_id` / `user_id` dropping `Default: "0"` — confirm no `ImportStateVerify` tests depend on those fields being `"0"` vs `null`

### **Issue Ref**
https://github.com/grafana/deployment_tools/issues/538219


internal/resources/grafana/resource_dashboard_permission.go
mihai-turdean, line 1: In the folder permission resource, Create and ImportState both re-read the UID from the API response (`folderResp.Payload.UID`), which normalizes it (e.g., if casing differs). The dashboard resource uses the user-provided UID directly (`data.DashboardUID.ValueString()`) without normalizing from the `GetDashboardByUID` response. If the API ever normalizes UIDs differently from user input, this could cause state drift. Consider:
```go
dashResp, err := client.Dashboards.GetDashboardByUID(dashboardUID)
// ...
dashboardUID = dashResp.Payload.Dashboard.(map[string]interface{})["uid"].(string)
```

rooneyshuman: suggestion applied

mihai-turdean (approved):

PR #2657, opened 2026-04-03
App Platform: generic resource

ArturWierzbicki:
Generic resource that accepts any k8s-shaped manifest, features
- it determines the API route based on API Group and API Discovery (for plural kind name) 
- it auto-discovers `stack_id` for cloud
   - if autodiscovered `stack_id` doesn't match the provided one -> error, else -> use it
   - if can't autodiscover but there is a set `stack_id` -> use it
   - if can't autodiscover, and no `stack_id` -> use `org_id`
- it supports the `secure` section as write-only attributes
- it accepts namespaces-resources only
- it checks drift differently for `metadata` and `spec`
   - for `spec`, config is the source of truth as a whole - anything added/changed compared to config causes drift
   - for `metadata`, drift is calculated only for configured fields; eg extra labels/annotation _not_ defined in config won't show up as drift, they are implicitly marked as server-managed
- the `metadata.uid` behavior matches the typed app platform resources - ie. `metadata.uid` is an alias for `metadata.name`. the apiserver-created `metadata.uid` aka `uuid` is stored internally and doesn't show up in the config + doesn't cause drift

nice to have things in the future..
- check configured `spec` against server's openapi spec during plan to surface incompatibilities 
- add `server-owned-paths` that would take a list of jsonpaths on `spec` which would NOT be taken into account when doing drift calculation. useful eg for mutation hooks


internal/resources/appplatform/generic/generic_acc_test.go
```diff
@@ -0,0 +1,294 @@
...
+			if err != nil {
+				lastErr = err
+			} else if check == nil {
+				return nil
+			} else if err := check(resource); err == nil {
+				return nil
+			} else {
+				lastErr = err
+			}
+
```
spinillos, line 122:
```suggestion
			} else if err := check(resource); err != nil {
				lastErr = err
			}
```

internal/resources/appplatform/generic/generic_acc_test.go
```diff
@@ -0,0 +1,294 @@
...
+	isNotFound genericNotFoundFunc,
+) error {
+	deadline := time.Now().Add(30 * time.Second)
+	var lastErr error
+
+	for time.Now().Before(deadline) {
+		if _, err := getter(ctx, client, uid); err == nil {
+			lastErr = fmt.Errorf("%s %s still exists", resourceLabel, uid)
+			time.Sleep(1 * time.Second)
+			continue
```
spinillos: I think that we don't need this `continue` here.

internal/resources/appplatform/generic/generic_acc_test.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+}
+
+func waitForProvisioningAPI(t *testing.T) {
+	t.Helper()
+
+	baseURL := strings.TrimRight(os.Getenv("GRAFANA_URL"), "/")
+	if baseURL == "" {
+		t.Fatal("GRAFANA_URL must be set")
+	}
```
spinillos: This information is check in `CheckCloudInstanceTestsEnabled`

internal/resources/appplatform/generic/generic_acc_test.go
```diff
@@ -0,0 +1,294 @@
...
+
+	t.Fatalf("timed out waiting for provisioning API at %s (last result: %s)", reqURL, lastResult)
+}
+
+func setGrafanaAuth(req *http.Request) {
+	auth := os.Getenv("GRAFANA_AUTH")
+	if auth == "" {
+		return
+	}
```
spinillos: Same here. This is checked in `CheckCloudInstanceTestsEnabled`

internal/resources/appplatform/generic/generic_resource.go
```diff
@@ -0,0 +1,2509 @@
...
+			"Failed to autodiscover the Grafana Cloud stack namespace from `/bootdata`: %s. %s",
+			discoveryErr.Error(),
+			detail,
+		)
+	}
+	diags.AddError("Failed to resolve namespace", detail)
+	return "", diags
+}
+
+func (r *genericResource) grafanaGet(ctx context.Context, subpath string) ([]byte, error) {
```
spinillos, line 716: Shouldn't be all client configuration set [here](https://github.com/grafana/terraform-provider-grafana/blob/main/pkg/provider/configure_clients.go#L181)?

internal/resources/appplatform/generic/generic_resource.go
```diff
@@ -0,0 +1,2509 @@
...
+	if _, status := stringFieldAtPath(model.Manifest, "kind"); status == attrPathUnknown {
+		return true
+	}
+
+	if _, status := stringFieldAtPath(model.Manifest, "metadata", "uid"); status == attrPathUnknown {
+		return true
+	}
+	if uid, status := stringFieldAtPath(model.Manifest, "metadata", "uid"); status == attrPathKnown && uid != "" {
+		return false
+	}
```
spinillos:
```suggestion
uid, status := stringFieldAtPath(model.Manifest, "metadata", "uid")
if status == attrPathUnknown {
    return true
}
if status == attrPathKnown && uid != "" {
	return false
}
```

internal/resources/appplatform/generic/generic_resource.go
```diff
@@ -0,0 +1,2509 @@
...
+		}
+
+		normalized[key] = str
+	}
+
+	if diags.HasError() {
+		return nil, diags
+	}
+
+	return normalized, diags
```
spinillos:
```suggestion
	return normalized, diags
```

spinillos (approved):
I added some comments, but it seems to be fine.

internal/resources/appplatform/generic/generic_resource.go
ArturWierzbicki: moved here https://github.com/grafana/terraform-provider-grafana/pull/2657/changes/cbb5e0d63c6c8418927acaf37b8d0a12ac3e85c4

PR #2658, opened 2026-04-06
Migrate grafana report resource

rooneyshuman:
## Migration: `grafana_report` SDKv2 → Plugin Framework

This PR migrates the `grafana_report` resource from the legacy SDKv2 to the Plugin Framework.

### Changes

**`internal/resources/grafana/resource_report.go`**
- Rewrote resource using Plugin Framework (`resource.Resource`, `resource.ResourceWithConfigure`, `resource.ResourceWithImportState`, `resource.ResourceWithModifyPlan`)
- Renamed factory `resourceReport()` → `makeResourceReport()` to follow Framework naming convention
- Replaced SDKv2 `schema.Resource` with typed Go models (`resourceReportModel`, `resourceReportScheduleModel`, etc.) and Framework schema
- `ModifyPlan` normalizes `workdays_only` to `false` for frequencies that don't support it — matches SDKv2 behavior
- `preserveScheduleTimes` helper (called from `Create`, `Update`, and `Read`) writes the user-provided start/end time format back into state after each API round-trip, avoiding "Provider produced inconsistent result after apply" errors

**`internal/resources/grafana/resources.go`**
- Updated registration: `resourceReport()` → `makeResourceReport()`

**`internal/resources/grafana/resource_report_test.go`**
- Updated two time-value assertions to expect config format instead of API-normalized format (see below)
- Replaced `dashboards.1.time_range.0.from = ""` assertion with `dashboards.1.time_range.# = 0` — Framework correctly represents an absent block as an empty list rather than a block with empty strings

**`docs/resources/report.md`**
- Regenerated via `go generate ./...`

### Behavioral change: start_time / end_time are stored in config format

The v5 mux strictly requires plan values for `Optional+Computed` attributes inside `ListNestedBlock`s to match the config value exactly — no normalization is allowed. Because of this, state now stores the user-provided format (e.g. `"2020-01-01T00:00:00-07:00"`) rather than the API-normalized form (`"2020-01-01T07:00:00Z"`). Two test assertions were updated to reflect this. Both strings represent the same moment, so subsequent plans produce no diff.

### Known limitation: `schedule` block cannot be marked `Required` in the schema

`schedule` uses `ListNestedBlock` (not `ListNestedAttribute`) because the v5 mux [does not support](https://github.com/hashicorp/terraform-plugin-docs/issues/363) `NestedType` on attributes. `ListNestedBlock` has no `Required` flag; enforcement is done via `listvalidator.SizeAtLeast(1)` + `listvalidator.SizeAtMost(1)` validators, and the description reads `"(Required) Schedule of the report."` as a hint. 

## Issue Ref
- https://github.com/grafana/deployment_tools/issues/540919

### DIFF
--- a/internal/resources/grafana/resource_report.go
+++ b/internal/resources/grafana/resource_report.go
@@ -408,10 +408,11 @@ func (r *reportResource) Read(ctx context.Context, req resource.ReadRequest, res
 		resp.State.RemoveResource(ctx)
 		return
 	}
-	// Preserve the user-provided time format from the existing state. The v5 mux requires
-	// that plan values for Optional+Computed attributes in blocks match the config exactly
-	// (no plan modification is allowed), so we store times in config format and avoid drift.
-	preserveScheduleTimes(readData, &data)
+	// Preserve the user-provided time format from the existing state only when the API
+	// value represents the same instant. The v5 mux requires plan values for
+	// Optional+Computed attributes in blocks to match config exactly (no plan modifier
+	// allowed), so we avoid format-only diffs while still surfacing genuine time changes.
+	preserveScheduleTimesIfSemanticEqual(readData, &data)
 	resp.Diagnostics.Append(resp.State.Set(ctx, readData)...)
 }
 
@@ -622,7 +623,7 @@ func modelToReport(ctx context.Context, data *resourceReportModel) (models.Creat
 		Formats: []models.Type{reportFormatPDF},
 	}
 
-	if !data.Formats.IsNull() {
+	if !data.Formats.IsNull() && len(data.Formats.Elements()) > 0 {
 		var formatStrs []string
 		diags.Append(data.Formats.ElementsAs(ctx, &formatStrs, false)...)
 		if diags.HasError() {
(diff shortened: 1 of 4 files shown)


internal/resources/grafana/resource_report.go
```diff
@@ -48,364 +55,566 @@ var (
...
+func modelToReport(ctx context.Context, data *resourceReportModel) (models.CreateOrUpdateReport, diag.Diagnostics) {
+	var diags diag.Diagnostics
+
+	var recipients []string
+	diags.Append(data.Recipients.ElementsAs(ctx, &recipients, false)...)
+	if diags.HasError() {
+		return models.CreateOrUpdateReport{}, diags
+	}
+
+	schedule := data.Schedule[0]
```
mmandrus, line 608: nit: check that Schedule is non-empty

internal/resources/grafana/resource_report.go
```diff
@@ -414,240 +623,291 @@ func schemaToReport(d *schema.ResourceData) (models.CreateOrUpdateReport, error)
...
-
 	return number, unit, nil
 }
 
-func validateTimezone(i any, path cty.Path) diag.Diagnostics {
-	timezone := i.(string)
-	_, err := time.LoadLocation(timezone)
-	return diag.FromErr(err)
+func formatDate(date string, timezone *time.Location) (*strfmt.DateTime, error) {
+	parsedDate, err := time.Parse(timeDateShortFormat, date)
```
mmandrus:
```suggestion
	parsedDate, err := time.ParseInLocation(timeDateShortFormat, date, timezone)
```

From cursor: "time.Parse without a zone uses UTC for the short layout. Then .In(timezone) converts that instant to the schedule timezone for the API"

mmandrus (approved):

PR #2664, opened 2026-04-07
feat: add azure managed identity and workload identity support to sso settings

### LINKED ISSUE #2214: Add Azure managed Identity/workload Identity support to SSO settings
Add support for configuring Azure managed identities, workload identities in the `sso_settings_resource` for the Azure AD connector.

Related PRs/issues:
- https://github.com/grafana/grafana/pull/104807 (https://github.com/grafana/grafana/issues/78249)
- https://github.com/grafana/grafana/pull/95455 (https://github.com/grafana/grafana/issues/95168)

dennismdejong:
## Summary
- Added `client_authentication`, `managed_identity_client_id`, `workload_identity_token_file`, and `federated_credential_audience` to `grafana_sso_settings` (fixes #2214).
- Updated `client_id` to be optional, since Azure AD can use `managed_identity_client_id` or `client_authentication` instead of a plain client ID.
- Added required structural validations for Azure AD and other OAuth2 providers (so `client_id` is still enforced where strictly required).
- Regenerated schema and documentation.

## What has been tested
- Verified unit and integration tests run successfully, catching appropriate validation errors when variables are omitted.


internal/resources/grafana/resource_sso_settings.go
```diff
@@ -992,7 +1012,7 @@ func getSettingsFromResourceData(d *schema.ResourceData, settingsKey string) (ma
 
 	// sometimes the settings set contains some empty items that we want to ignore
 	// we are only interested in the settings that have one of the following:
-	// - the client_id set because the client_id is a required field for OAuth2 providers
+	// - the client_id or managed_identity_client_id or client_authentication set because they are required fields for OAuth2 providers
 	// - a non-empty config for LDAP
 	// - the private_key or private_key_path set because those are required fields for SAML
```
volcanonoodle: I think this could be a bit confusing because `managed_identity_client_id` is not generic to all OAuth2 providers. Perhaps we could make the comment something like the following, WDYT?
```
// - the client_id set because the client_id is a required field for OAuth2 providers
// - the managed_identity_client_id or client_authentication for relevant EntraID OAuth setups
// ...
```

internal/resources/grafana/resource_sso_settings_test.go
```diff
@@ -350,7 +350,7 @@ func TestSSOSettings_resourceWithEmptySettings(t *testing.T) {
 		Steps: []resource.TestStep{
 			{
 				Config:      testConfigWithEmptySettings,
-				ExpectError: regexp.MustCompile("Missing required argument"),
+				ExpectError: regexp.MustCompile("client_id must be set for the provider okta|Missing required argument"),
```
volcanonoodle, line 394: Does this need to match 2 different error messages?

volcanonoodle (commented):
Thank you very much for the contribution!! Just added some small comments, this looks very nice 😄

volcanonoodle:
I think we'd also want to add tests that check that these new fields can be set and read back from the API, like in https://github.com/grafana/terraform-provider-grafana/blob/576724e793e28008f675a35e48d966de8c7909c8/internal/resources/grafana/resource_sso_settings_test.go#L202
but since those might need to be cloud-only from looking at the comment in the code, maybe we can deal with those later on 😄

volcanonoodle:
Actually, I've just been made aware that we're already placing specific EntraID config options (like `force_use_graph_api`) in the custom fields section here:
https://github.com/grafana/terraform-provider-grafana/blob/576724e793e28008f675a35e48d966de8c7909c8/internal/resources/grafana/resource_sso_settings.go#L315-L322

Could you move these there too please? 🙏

dennismdejong:
Thanks for the feedback! I've moved the EntraID-specific configuration fields into the custom fields map to align with existing practices and updated the validation logic and comments accordingly. Regarding the test, I'll add the suggested test cases to ensure these fields can be set and read back. I kept the OR regex in the test to account for potential variations in the Terraform error output between provider-level and schema-level validation.

internal/resources/grafana/resource_sso_settings.go
dennismdejong: "Thanks for the suggestion! That is indeed much clearer and more accurate regarding how EntraID setups work compared to standard OAuth2 providers. I have updated the code comment to reflect that managed_identity_client_id or client_authentication are valid alternatives to client_id for these specific scenarios."

internal/resources/grafana/resource_sso_settings_test.go
dennismdejong: Yes, that's intentional. Depending on how the resource is configured in the test, either the standard Terraform SDK schema validation (Missing required argument) or our custom validation logic (at least one of...) might trigger first. Using the regex pipe (|) ensures the test remains stable and passes regardless of which validation layer catches the missing configuration first."

PR #2668, opened 2026-04-07
Migrate disabled alert config

rooneyshuman:
### **Review Guide**
This PR migrates `grafana_asserts_suppressed_assertions_config` from SDKv2 to the Plugin Framework.

**`resource_disabled_alert_config.go`** - standard Framework rewrite. Notable:
- `disabledAlertConfigResource` holds `*assertsapi.APIClient` and `int64 stackID` directly, injected via `Configure`
- `id` is a new computed attribute with `UseStateForUnknown()` (set to `name` on read, consistent with other asserts resources)
- `name` retains `RequiresReplace()` — equivalent to the previous `ForceNew: true`
- `match_labels` is `schema.MapAttribute{ElementType: types.StringType}`, populated with `types.MapValueFrom`
- Private `put` helper handles upsert via `PutDisabledAlertConfig`
- Private `read` helper retains eventual-consistency retry via `withRetryRead` (SDKv2 `helper/retry` aliased as `sdkretry`)
- `ImportState` implemented using the `read` helper rather than `resource.ImportStatePassthroughID`, so import populates the full model

**`docs/`** - generated; `make docs` produces no changes to the asserts suppressed assertions config page.

### Issue ref
https://github.com/grafana/deployment_tools/issues/542767


rooneyshuman:
@codex review

internal/resources/asserts/resource_disabled_alert_config.go
```diff
@@ -4,204 +4,253 @@ import (
...
+	dto := assertsapi.DisabledAlertConfigDto{
+		Name:      &name,
+		ManagedBy: getManagedByTerraform(),
 	}
-	name := d.Id()
 
-	// Retry logic for read operation to handle eventual consistency
+	if !data.MatchLabels.IsNull() && len(data.MatchLabels.Elements()) > 0 {
+		var matchLabels map[string]string
+		if diags := data.MatchLabels.ElementsAs(context.Background(), &matchLabels, false); diags.HasError() {
```
jgehrett:
```suggestion
		if diags := data.MatchLabels.ElementsAs(ctx, &matchLabels, false); diags.HasError() {
```
Should this use the passed `ctx`?

internal/resources/asserts/resource_disabled_alert_config.go
```diff
@@ -4,204 +4,253 @@ import (
...
+			return fmt.Errorf("failed to read match_labels")
```
jgehrett: Can you include some detail from the `diags` response in the message?

jgehrett (commented):
I'm new to this codebase so asked Claude for some help. I only have a couple questions / suggestions

internal/resources/asserts/resource_disabled_alert_config.go
rooneyshuman: can do!

internal/resources/asserts/resource_disabled_alert_config.go
rooneyshuman: done!

jgehrett (approved):
LGTM

PR #2670, opened 2026-04-08
fix(sm): Add validation for private probe labels

Pokom:
Private probes should only have a max of 3 labels. This updates the behavior to provide this as an error the user during planning isntead of failing silently. It also updates the generated documentation to provide the maximum number of labels.

- fixes https://github.com/grafana/synthetic-monitoring/issues/535


jmichalek132 (approved):
LGTM.

PR #2684, opened 2026-04-14
feat(api): update all resources for Grafana v13

macabu:
## Summary

This PR updates the OpenAPI schema for Grafana v13. It introduces some non-breaking but notable changes namely in the following APIs:

- Folders: No longer supportes numeric folder `id`, and if your Terraform state is using them, we will attempt to migrate to `uid` instead.
- Roles: The `version` is no longer used in the request, and the server manages versioning automatically. This is being skipped from the Terraform diff, as there might be version changes, which do not affect any functionality.


macabu:
@codex review

macabu:
@codex review

macabu:
@codex review

internal/resources/grafana/resource_role.go
alexanderzobnin, line 191: This is from my fellow Claude friend:

All six of these fields changed from value types to pointers (string → *string, int64 → *int64). While SDKv2's d.Set() does use mapstructure.WeakDecode internally which can dereference pointers, this is fragile and inconsistent with the rest of the PR. If any field is ever nil (e.g., during a partial API response or error scenario), this would silently set zero values rather than producing an explicit error. These should be explicitly dereferenced.

So probably makes sense to verify.

alexanderzobnin (approved):

internal/resources/grafana/resource_role.go
macabu: Good catch! hmm tests failed, so I reverted, but Claude also says that we shouldnt be changing it

Duologic (changes requested):
The breaking changes listed in the PR description can be avoided or are already mitigated:

**Roles:** Already handled -- `version` and `auto_increment_version` are properly deprecated with `DiffSuppressFunc` that always returns true, and the `ExactlyOneOf` constraint is removed. Existing configs continue to plan without changes. Nice work.

**Folders:** The `GetFolderByIDorUID` removal doesn't need to be a breaking change. Since `GetFolderByID` is gone from the v13 client, we can't resolve numeric IDs, but we *can* detect them and return a clear actionable error instead of letting `GetFolderByUID("123")` 404 silently (which Terraform treats as "resource deleted" -- potentially cascading into destroying contents).

Something like this in `ReadFolder`:

```go
func ReadFolder(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
	metaClient := meta.(*common.Client)
	client, orgID, idStr := OAPIClientFromExistingOrgResource(meta, d.Id())

	// Detect legacy numeric folder IDs that can no longer be resolved
	if _, err := strconv.ParseInt(idStr, 10, 64); err == nil {
		return diag.Diagnostics{{
			Severity: diag.Error,
			Summary:  "Numeric folder IDs are no longer supported",
			Detail: fmt.Sprintf(
				"The folder resource %q uses a numeric ID (%s). "+
					"Grafana v13 removed the numeric folder ID API. "+
					"Please remove this resource from state and re-import it using the folder UID:\n\n"+
					"  terraform state rm %s\n"+
					"  terraform import %s <orgID>:<folderUID>",
				d.Id(), idStr,
				d.Id(), d.Id(),
			),
		}}
	}

	response, err := client.Folders.GetFolderByUID(idStr)
	// ... rest unchanged
```

This turns a silent data-loss scenario into a clear one-time migration instruction. With this, none of the changes in this PR are breaking for end users.

**Separately:** the `resource_role_version_diff_test.go` file should be fully deleted (not left as a one-line stub) -- looks like this was already addressed in the latest commit.

examples/resources/grafana_apps_provisioning_connection_v0alpha1/private-key.pem
Duologic, line 1: **Private key PEM files:** On `main` these examples used `"replace-me"` as a placeholder. This PR introduces `filebase64("${path.module}/private-key.pem")` with actual PEM files committed to the repo, which triggers TruffleHog. These can be generated at test time instead -- the `examples_test.go` already resolves `${path.module}` to absolute paths, so the TF examples don't need to change.

Something like this in `examples_test.go`:

```go
// Add to imports:
// "crypto/rand"
// "crypto/rsa"
// "crypto/x509"
// "encoding/pem"
// "os"

// At the start of TestAccExamples, before the test loop:
func TestAccExamples(t *testing.T) {
	if testing.Short() {
		t.Skip("skipping long test")
	}

	// Generate a dummy PEM key for examples that reference private-key.pem
	generateDummyPEM(t)

	// ... rest of function unchanged
}

func generateDummyPEM(t *testing.T) {
	t.Helper()

	key, err := rsa.GenerateKey(rand.Reader, 2048)
	if err != nil {
		t.Fatalf("failed to generate RSA key: %v", err)
	}

	keyBytes, err := x509.MarshalPKCS8PrivateKey(key)
	if err != nil {
		t.Fatalf("failed to marshal private key: %v", err)
	}

	pemBlock := pem.EncodeToMemory(&pem.Block{
		Type:  "PRIVATE KEY",
		Bytes: keyBytes,
	})

	// Write to all example dirs that need a private-key.pem
	for _, dir := range []string{
		"../../examples/resources/grafana_apps_provisioning_connection_v0alpha1",
		"../../examples/resources/grafana_apps_provisioning_repository_v0alpha1",
	} {
		path := filepath.Join(dir, "private-key.pem")
		if err := os.WriteFile(path, pemBlock, 0600); err != nil {
			t.Fatalf("failed to write %s: %v", path, err)
		}
		t.Cleanup(func() { os.Remove(path) })
	}
}
```

Then delete the committed PEM files and add `private-key.pem` to `.gitignore`. The TF examples keep using `filebase64("${path.module}/private-key.pem")` unchanged -- they just won't be checked in.

macabu: Done!

internal/resources/grafana/resource_folder.go
```diff
@@ -176,10 +178,27 @@ func ReadFolder(ctx context.Context, d *schema.ResourceData, meta any) diag.Diag
 	metaClient := meta.(*common.Client)
 	client, orgID, idStr := OAPIClientFromExistingOrgResource(meta, d.Id())
 
-	folder, err := GetFolderByIDorUID(client.Folders, idStr)
+	// Detect legacy numeric folder IDs that can no longer be resolved since Grafana v13
+	if _, err := strconv.ParseInt(idStr, 10, 64); err == nil {
+		return diag.Diagnostics{{
```
Duologic: The diagnostic error is still a breaking experience for users. 


Since `GetFolders` returns `FolderSearchHit` with both `ID` and `UID`, we can transparently migrate numeric IDs instead, **the team owning the Folder resource might know about better options to do this**:

```go
// Replace the numeric ID error block in ReadFolder with:

	// Transparently migrate legacy numeric folder IDs to UIDs.
	// GetFolderByID was removed in the Grafana v13 client, so we resolve
	// numeric IDs by listing folders and matching on the deprecated ID field.
	if numericalID, err := strconv.ParseInt(idStr, 10, 64); err == nil {
		uid, err := resolveFolderNumericID(client, numericalID)
		if err != nil {
			return diag.FromErr(err)
		}
		idStr = uid
		d.SetId(MakeOrgResourceID(orgID, uid))
	}
```

```go
// Add this helper:

func resolveFolderNumericID(client *goapi.GrafanaHTTPAPI, numericalID int64) (string, error) {
	page := int64(1)
	for {
		resp, err := client.Folders.GetFolders(folders.NewGetFoldersParams().WithPage(&page))
		if err != nil {
			return "", fmt.Errorf("failed to resolve numeric folder ID %d: %w", numericalID, err)
		}
		for _, f := range resp.Payload {
			if f.ID == numericalID {
				return f.UID, nil
			}
		}
		if len(resp.Payload) == 0 {
			break
		}
		page++
	}
	return "", fmt.Errorf("folder with numeric ID %d not found, it may have been deleted", numericalID)
}
```

This way users with legacy numeric IDs in state get a seamless upgrade -- the state is silently corrected on the first `terraform plan`/`apply` with no manual intervention needed.

Duologic (changes requested):
I think we can do a little bit better on the folder resource, other than that this PR looks good to go.

macabu:
@codex review

internal/resources/grafana/resource_folder.go
```diff
@@ -176,10 +178,23 @@ func ReadFolder(ctx context.Context, d *schema.ResourceData, meta any) diag.Diag
...
 
-	folder, err := GetFolderByIDorUID(client.Folders, idStr)
+	// Transparently migrate legacy numeric folder IDs to UIDs.
+	// GetFolderByID was removed in the Grafana v13 client, so we resolve
+	// numeric IDs by listing folders and matching on the deprecated ID field.
+	if numericalID, err := strconv.ParseInt(idStr, 10, 64); err == nil {
+		uid, err := resolveFolderNumericID(client, numericalID)
+		if err != nil {
+			return diag.FromErr(err)
+		}
```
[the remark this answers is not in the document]
macabu: hmm i dont think we want to remove from state here

internal/resources/grafana/resource_folder.go
macabu: Done!

internal/resources/grafana/resource_folder.go
```diff
@@ -128,10 +129,11 @@ func CreateFolder(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
 func UpdateFolder(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
 	client, _, idStr := OAPIClientFromExistingOrgResource(meta, d.Id())
 
-	folder, err := GetFolderByIDorUID(client.Folders, idStr)
+	response, err := client.Folders.GetFolderByUID(idStr)
```
Duologic, line 137: **Bug (medium severity):** `UpdateFolder` previously called `GetFolderByIDorUID`, which handled both numeric and UID-based lookups. It now calls `GetFolderByUID` directly without the numeric ID migration that was added to `ReadFolder`.

If a user's state still holds a legacy numeric folder ID (e.g., `1:42`) and they run `terraform apply -refresh=false`, `ReadFolder` is skipped and `idStr` arrives here as `"42"`. `GetFolderByUID("42")` will look up a folder with UID `"42"` instead of resolving numeric ID 42, likely returning a 404 or the wrong folder.

In normal usage Terraform calls `Read` before `Update`, which migrates the ID. But `-refresh=false` bypasses that.

Consider adding the same `strconv.ParseInt` + `resolveFolderNumericID` migration from `ReadFolder` here, or extracting it into a shared helper called from both.

macabu: @Duologic Done!

Duologic (approved):

PR #2688, opened 2026-04-14
chore: update codeowners

jtroy:
It seems that the Identity Squad owns all things **Team**, so update CODEOWNERS to reflect this.


Duologic:
There are also catalog-info.yaml files that might need an update as well.

jtroy:
> There are also catalog-info.yaml files that might need an update as well.

Thanks! I think I got them now.

Jguer (approved):

PR #2689, opened 2026-04-14
fix(dashboard_permission): normalize team_id/user_id '0' to null in plan

rooneyshuman:
### **Changes Made**
**Root cause**

The SDKv2 version of `grafana_dashboard_permission` (and `grafana_folder_permission`) defaulted `team_id`/`user_id` to `"0"`. Many users carried `team_id = "0"` / `user_id = "0"` into their configs to mean "not set".

After the Framework migration in #2653, `readBulkPermissions` correctly returns `types.StringNull()` for zero IDs:

```go
if perm.TeamID > 0 {
    item.TeamID = types.StringValue(...)
} else {
    item.TeamID = types.StringNull()  // "0" becomes null
}
```

But the plan preserved the literal `"0"` from config. Since `permissions` is a `SetNestedBlock`, elements are matched by their full value — `{team_id="0"}` ≠ `{team_id=null}` → **"provider produced inconsistent result after apply"** (support escalation grafana/support-escalations#21762).

**Fix**

Extend `stripOrgScopedIDPlanModifier` (already applied to both fields) to normalize `"0"` → `null` at plan time, so plan and Read agree. `applyBulkPermissions` already handles null and `"0"` identically (`if teamID > 0` guard), so apply behaviour is unchanged.

Affects both `grafana_dashboard_permission` and `grafana_folder_permission` since both share `bulkPermissionsSchemaAttribute`.

### **Testing**
Built a dev version of the provider and used it to reproduce the steps outlined in the support issue using the same defined `main.tf` file. After the change, verified that the resources were able to be created without the `inconsistent result` errors.
```
❯ terraform apply
╷
│ Warning: Provider development overrides are in effect
│
│ The following provider development overrides are set in the CLI configuration:
│  - grafana/grafana in /Users/belen/repos/terraform-provider-grafana
│
│ The behavior may therefore not match any released version of the provider and applying changes may cause the state to become incompatible with
│ published releases.
╵

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # grafana_dashboard.home_dashboard will be created
  + resource "grafana_dashboard" "home_dashboard" {
      + config_json  = jsonencode(
            {
              + panels        = []
              + schemaVersion = 16
              + tags          = []
              + title         = "Home Dashboard"
              + uid           = null
            }
        )
      + dashboard_id = (known after apply)
      + folder       = (known after apply)
      + id           = (known after apply)
      + uid          = (known after apply)
      + url          = (known after apply)
      + version      = (known after apply)
    }

  # grafana_dashboard_permission.home_dashboard_permission will be created
  + resource "grafana_dashboard_permission" "home_dashboard_permission" {
      + dashboard_uid = (known after apply)
      + id            = (known after apply)
      + org_id        = (known after apply)

      + permissions {
          + permission = "View"
          + role       = "Viewer"
        }
    }

  # grafana_folder.bam_standard_dashboards will be created
  + resource "grafana_folder" "bam_standard_dashboards" {
      + id                           = (known after apply)
      + prevent_destroy_if_not_empty = false
      + title                        = "BAM Standard Dashboards"
      + uid                          = (known after apply)
      + url                          = (known after apply)
    }

Plan: 3 to add, 0 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

grafana_folder.bam_standard_dashboards: Creating...
grafana_folder.bam_standard_dashboards: Creation complete after 0s [id=0:bfj39v12n9b7kd]
grafana_dashboard.home_dashboard: Creating...
grafana_dashboard.home_dashboard: Creation complete after 1s [id=0:3ae8f2cb-9d57-4105-8301-994a345e833d]
grafana_dashboard_permission.home_dashboard_permission: Creating...
grafana_dashboard_permission.home_dashboard_permission: Creation complete after 0s [id=0:3ae8f2cb-9d57-4105-8301-994a345e833d]

Apply complete! Resources: 3 added, 0 changed, 0 destroyed.
```

### DIFF
--- a/pkg/generate/postprocessing/replace_references.go
+++ b/pkg/generate/postprocessing/replace_references.go
@@ -58,6 +58,7 @@ var knownReferences = []string{
 	"grafana_dashboard.uid=grafana_library_panel.uid",
 	"grafana_dashboard_permission.dashboard_uid=grafana_dashboard.uid",
 	"grafana_dashboard_permission.team_id=grafana_team.id",
+	"grafana_dashboard_permission.user_id=grafana_service_account.id",
 	"grafana_dashboard_permission.user_id=grafana_user.id",
 	"grafana_dashboard_permission_item.dashboard_uid=grafana_dashboard.uid",
 	"grafana_dashboard_permission_item.team=grafana_team.id",
--- a/internal/resources/grafana/common_plugin_framework.go
+++ b/internal/resources/grafana/common_plugin_framework.go
@@ -215,13 +215,17 @@ func (d *orgScopedAttributePlanModifier) PlanModifyString(ctx context.Context, r
 	}
 }
 
-// stripOrgScopedIDPlanModifier unconditionally strips the org prefix from plan values,
-// normalizing "orgID:localID" to "localID". Used in bulk permission blocks where
-// referenced resources may return either format.
+// stripOrgScopedIDPlanModifier normalizes team_id/user_id plan values for bulk
+// permission blocks:
+//   - Strips the org prefix: "orgID:localID" → "localID"
+//   - Normalizes "0" → null: zero means "not set", matching what readBulkPermissions
+//     returns (types.StringNull) when the API returns TeamID/UserID == 0. This
+//     prevents "inconsistent result after apply" errors when users carry over the
+//     legacy SDKv2 pattern of explicitly setting team_id = "0" or user_id = "0".
 type stripOrgScopedIDPlanModifier struct{}
 
 func (d *stripOrgScopedIDPlanModifier) Description(_ context.Context) string {
-	return "Strips the org ID prefix from resource IDs, normalizing to the local ID."
+	return "Strips the org ID prefix from resource IDs and normalizes \"0\" to null."
 }
 
 func (d *stripOrgScopedIDPlanModifier) MarkdownDescription(ctx context.Context) string {
@@ -233,6 +237,10 @@ func (d *stripOrgScopedIDPlanModifier) PlanModifyString(_ context.Context, req p
 		return
 	}
 	_, localID := SplitOrgResourceID(resp.PlanValue.ValueString())
+	if localID == "0" {
+		resp.PlanValue = types.StringNull()
+		return
+	}
 	if localID != "" {
 		resp.PlanValue = types.StringValue(localID)
 	}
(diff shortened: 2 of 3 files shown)


forsethc (commented):
LGTM

forsethc (approved):

rooneyshuman:
CI failures blocked until https://github.com/grafana/terraform-provider-grafana/pull/2697 is merged

suntala:
I would have preferred a different approach, but I am missing some context here. I will go ahead and merge because I know there is an escalation that is relying on this fix and the owner team gave their approval. Also, without deeper context on this, it does seem fine for “0” and null to be treated the same semantically.

My preferred approach would have been to keep/restore the legacy behavior of team_id/user_id defaulting to “0” (changed by [#2608](https://github.com/grafana/terraform-provider-grafana/pull/2608)) and maintain consistency by having both plan and apply set those IDs to "0" (so readBulkPermissions, applyBulkPermissions and PlanModifyString should all be aligned on this). The bottom line is that I would like to preserve legacy behavior as much as possible.

However, since there were some different callouts asserting that the "0 should be equivalent to null" approach is the correct one, I don’t feel comfortable replacing this PR with my preferred approach. 
* It was specifically called out that “team_id/user_id no longer default to "0" - this is intentional and reflected in the docs” ([PR description](https://github.com/grafana/terraform-provider-grafana/pull/2608)). → why is it better to no longer default to 0?
* “After the Framework migration in #2653, readBulkPermissions correctly returns types.StringNull() for zero IDs” ([PR description](https://github.com/grafana/terraform-provider-grafana/pull/2689))  → why is it correct for “null” to be returned for zero IDs?

pkg/generate/postprocessing/replace_references.go
```diff
@@ -58,6 +58,7 @@ var knownReferences = []string{
 	"grafana_dashboard.uid=grafana_library_panel.uid",
 	"grafana_dashboard_permission.dashboard_uid=grafana_dashboard.uid",
 	"grafana_dashboard_permission.team_id=grafana_team.id",
+	"grafana_dashboard_permission.user_id=grafana_service_account.id",
```
suntala, line 61: I would have preferred not to include this improvement as part of the migration/migration fix. It seems small enough that we can leave it in though.

PR #2713, opened 2026-04-17
feat: add support for named routing trees to alertrule resource

khalilhaji:
## Summary
Adds support for using a named routing tree in notification settings by introducing nested simplified_routing and named_routing_tree blocks. This marks the flat config as deprecated but continue to accept it to avoid this being a user-facing breaking change.

## Backwards Compatibility
- This is not a breaking change because existing configurations that use the legacy flat notification_settings fields are still accepted

## Testing
- Added unit tests covering legacy flat notification settings, nested simplified routing, named routing tree support, and invalid mixed-shape configurations.
- Unskipped the alert rule example acceptance tests locally and ran them against local grafana


internal/resources/appplatform/alertrule_resource.go
```diff
@@ -168,50 +205,202 @@ This resource is currently in alpha and is subject to change. Grafana 12.4+ user
...
+				Description:        "Deprecated. Minimum time interval for re-sending a notification if an alert is still firing.",
 				Validators: []validator.String{
 					PrometheusDurationValidator{},
 				},
 			},
 		},
+		Blocks: map[string]schema.Block{
+			"simplified_routing": schema.SingleNestedBlock{
+				Description: "Simplified routing to a contact point with optional grouping and timing overrides.",
+				Validators:  []validator.Object{requireAttrsWhenPresent("contact_point")},
```
khalilhaji, line 261: Note that `requireAttrsWhenPresent` doesn't affect the generated docs which still say this is optional. We use this in a few places with the same caveat for the generated docs.

moustafab (approved):
Changes look good to me, nice job!

PR #2724, opened 2026-04-22
fix(grafana_team): prevent mass member removal on ignore_externally_synced transition

Duologic:
## Summary

- Fixes a bug where transitioning `ignore_externally_synced_members` from `false` to `true` causes all externally synced team members to be removed via the API
- Replaces `Default([])` on `members` with a plan modifier that preserves prior state when members is not in config, preventing the empty-set diff that triggers mass removal
- Adjusts `Update()` to use the state's `ignore` value for the final read when members is unconfigured, so the returned member list matches the plan (preventing "inconsistent result after apply")

## Root Cause

The v4.31.1 migration added `Default: setdefault.StaticValue([])` on the `members` attribute. This means "members not in config" becomes `plan=[]`, which diffs against state (containing members) and triggers removal API calls. The old SDKv2 code had a `DiffSuppressFunc` that treated null and empty as equivalent, preventing this.

## The Fix

1. **Plan modifier** (`membersUseStateWhenUnconfigured`): When config members is null and prior state exists, preserve state. When null and no state (create), default to `[]`. When explicitly set (even to `[]`), let the framework handle it.
2. **Update() consistency**: When members is unconfigured, the final `r.read()` uses the **state's** `ignore` value (not the plan's) so the returned member list matches what the plan modifier predicted. The `ignore` field is then overridden to the plan value, taking effect on the next refresh cycle.

### Two-cycle convergence

When `ignore` transitions `false→true`:
- **Cycle 1**: Plan modifier preserves state members. Update makes no member API calls. Final read uses old `ignore=false` → members still in state. State saved with `ignore=true`.
- **Cycle 2**: Read refreshes with `ignore=true` → filters labeled members → state cleared. No diff, stable.

## Testing

Adds a mock unit test (`TestUnitTeam_IgnoreExternallySyncedTransition_Mock`) that:
1. Creates a team with `ignore=false` and explicit members (all externally synced via LDAP labels)
2. Transitions to `ignore=true` with no `members` in config — asserts 0 removal API calls and members preserved in state
3. Re-applies same config — asserts members are cleared from state (via Read filtering), still 0 removal API calls

The test **fails** on unfixed code (`expected members.# "2", got "0"`) and **passes** with the fix.

Relates to [support escalation #21901](https://github.com/grafana/support-escalations/issues/21901).

### DIFF
--- a/internal/resources/grafana/resource_team_mock_test.go
+++ b/internal/resources/grafana/resource_team_mock_test.go
@@ -1,0 +1,184 @@
+package grafana_test
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"sync/atomic"
+	"testing"
+
+	"github.com/grafana/terraform-provider-grafana/v4/internal/testutils"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+// TestUnitTeam_IgnoreExternallySyncedTransition_Mock tests the transition of
+// ignore_externally_synced_members from false to true when members are not in
+// config. It uses a mock Grafana API that returns members with external sync
+// labels (simulating LDAP/SAML/team sync).
+//
+// The bug (support escalation #21901): With the Default([]) on the members
+// attribute, removing members from config causes plan=[] which diffs against
+// state (has members) and triggers mass removal. The correct behavior is to
+// preserve members and silently clear them from state via Read() filtering.
+func TestUnitTeam_IgnoreExternallySyncedTransition_Mock(t *testing.T) {
+	// memberRemovals tracks how many times the mock received a DELETE request
+	// to remove a team member. The fix should result in zero removals.
+	var memberRemovals atomic.Int32
+
+	mux := http.NewServeMux()
+
+	// POST /api/teams — create team
+	mux.HandleFunc("/api/teams", func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == http.MethodPost {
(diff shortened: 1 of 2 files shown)


suntala:
> Fixes a bug where transitioning ignore_externally_synced_members from false to true (while removing members from config)

We can discuss more concretely in Slack, but I don't remember us ever mentioning removing members from config while we were working on reproducing the issue.

Duologic:
Right, the sentence is not well formed, let me adjust.

internal/resources/grafana/resource_team_mock_test.go
```diff
@@ -0,0 +1,184 @@
...
+						if n := memberRemovals.Load(); n > 0 {
+							return fmt.Errorf("expected 0 member removal API calls, got %d", n)
+						}
+						return nil
+					},
+				),
+			},
+			// Step 3: Re-apply same config. The refresh (Read) now runs with
+			// ignore=true, filtering externally synced members from state.
+			// This is the "silent state clearing" — no diff, no API calls.
```
volcanonoodle, line 167: Should this step result in an empty plan? I'd think so if previously all members had been removed from the state, but I might be wrong. If it's supposed to result in an empty plan, could/should we check that with `ExpectNonEmptyPlan`? I'm very out of my depth here, please ignore if not relevant 😄

Duologic: Yes, it should come with an empty plan, so `Expect*Non*EmptyPlan` would fail in that case.

volcanonoodle: Whoops, I thought `ExpectNonEmptyPlan: false` would check that the plan was empty. Anyway, approved, thanks 🙇

volcanonoodle (approved):

PR #2727, opened 2026-04-23
fix(cloud): fix broken cloud acceptance tests

Duologic:
## Summary

The `cloud-acc-tests` workflow had been perma-broken for months. We fixed as many issues as we found, running the full test suite 10 times at each stage to measure progress:

- v1 (before fixes): **50%** pass rate
- v2 (after timeout/404 fixes): **70%** pass rate
- v3 (after all fixes including 409 handling): **100%** pass rate

This improved the situation a lot, but there is still an upstream bug that can cause transient failures.

### Upstream bug: gateway phantom-success (hosted-grafana#7649)

The cloud-grafana-gateway's reverse proxy returns a canned `503 Retry-After: 10` with `{"code":"Loading","message":"Your instance is loading..."}` when the upstream Grafana pod connection is reset or times out — even if the upstream POST was already processed. Because `retryablehttp` retries POST requests on 503, the retry hits the already-created resource and gets `400 ErrAlreadyExists`. The provider can't recover from this because GCOM doesn't expose a search/list endpoint for service accounts (only create, get-by-id, delete, tokens), so there's no way to look up the SA by name to get its ID.

Filed as [hosted-grafana#7649](https://github.com/grafana/hosted-grafana/issues/7649) — fix in progress upstream. Until that lands, SA-related tests may occasionally fail with `400 Bad Request`.

### Next steps

- Once the upstream gateway fix lands, run another 10-run analysis to confirm sustained 100% pass rate.
- Gate PRs on cloud acceptance tests via #2745, or alternatively run them in a non-blocking (informational) mode with a comment pinging the Monitoring squad (`@grafana/platform-monitoring`) when they fail, so flaky tests get triaged rather than ignored.

---

### Provider bug fixes

- **`testutils/provider.go`: remove eager `ConfigureProvider` call on factory creation.**
  The factory was eagerly invoking `ConfigureProvider` with an all-null config when the test factory was first created. This produced a partially-initialized `*common.Client` with `GrafanaAPI == nil` that the framework server cached as `ResourceConfigureData`. During the subsequent `ValidateResourceTypeConfig` phase, that cached client caused `basePluginFrameworkResource.Configure` to emit *"The Grafana Provider is missing a configuration for the Grafana API"* errors for resources using aliased providers whose configs depend on unknown resource outputs. The fix lets Terraform invoke `ConfigureProvider` per-instance with the actual resolved config — matching real-world `terraform apply` semantics.

- **`resource_cloud_stack.go`: use correct schema prefix for the grafana IP allow list.**
  `addIPAllowListIfPresent(d, "grafana", tenant)` used the prefix `"grafana"` (singular), but the schema attribute is `"grafanas_ip_allow_list_cname"` (plural). This caused a panic whenever GCOM returned an `IpAllowListCNAME` value for the grafana tenant.

- **`resource_cloud_stack_service_account.go` / `resource_cloud_stack_service_account_token.go`: treat 404 on delete as already deleted.**
  The delete functions for service accounts and service account tokens did not handle 404 responses. When a resource was deleted outside of Terraform (or when the gateway phantom-success bug caused a retry to find the resource already gone), the destroy phase would error with `404 Not Found`. Now follows the standard Terraform convention where 404 during destroy means the resource is already absent.

- **`resource_cloud_stack.go`: refine 409 Conflict handling in `createStack`.**
  The retry loop previously string-matched `"conflict"` on `err.Error()` and retried all 409 responses with a 10s backoff. This was wrong for several reasons:
  - It retried on genuine conflicts (slug actively in use) that would never resolve by waiting.
  - It used fragile string matching on the HTTP status text instead of checking the status code.
  - It hammered the GCOM API every 10s during the ~30s grace period, generating alert noise.

  Now the retry logic:
  - Checks `httpResp.StatusCode == 409` directly instead of string matching.
  - Inspects the GCOM response body to distinguish the grace period case (`"deleted recently"`) from real conflicts.
  - **Only retries on grace period 409s** — parses the wait duration from the GCOM message (e.g. `"wait for 30s"`) and sleeps for that duration plus a 5s margin. Falls back to 35s if parsing fails.
  - **Hard fails on all other 409s** (slug actively in use, slug still in downstream systems) and surfaces the GCOM error message to the Terraform user.

### Stack-readiness / timeout fixes

- **Use schema-configured `wait_for_readiness_timeout` for `waitUntilReady` (GCOM sub-resource readiness).**
  The `waitUntilReady` function used a hardcoded 5-minute timeout for GCOM sub-resource readiness polling, while the schema's `wait_for_readiness_timeout` (default 10m) only applied to the second HTTP health check (`waitForStackReadiness`). This caused `TestResourceStack_Basic` to fail ~20% of the time when cloud provisioning took longer than 5 minutes. Now both readiness checks share the same schema-configured timeout.
- **Fix tight-loop bug in `waitUntilReady`.**
  When the GCOM readiness API returned a transient error, `waitUntilReady` looped without any sleep, hammering the API. Added a 1-second sleep on error to match the success-but-not-ready path.
- **Increase stack readiness timeout from 5m to 10m** across `resource_cloud_stack.go`, `resource_cloud_stack_service_account.go`, `resource_cloud_stack_service_account_token.go`, and the `cloud_stack` docs.
- **Bump retry budget** (`GRAFANA_RETRIES=10`, `GRAFANA_RETRY_WAIT=5`) in the CI workflow to handle transient 503s from the cloud-grafana-gateway.

### Test correctness fixes

- **`resource_cloud_stack_service_account_test.go`: pass slug literal instead of `stack.Slug`.**
  `testAccGrafanaAuthCheckServiceAccounts(stack.Slug, ...)` evaluated `stack.Slug` at test step *definition* time, when `stack` is still the zero value. Fix: pass the `slug` literal variable (already known at definition time).
- **`resource_cloud_stack_service_account_test.go`: add `testAccGrafanaAuthCheckServiceAccountAbsent` helper.**
  Step 4's "absent" check called `testAccGrafanaAuthCheckServiceAccounts(slug, "", nil)` with empty `saName=""`, which never matched any SA and always returned *"expected to find key"*. Created a dedicated absent-check helper.
- **`resource_cloud_stack_service_account_test.go`: give `TestAccGrafanaServiceAccountFromCloud_AssignRoleOrPermissions` a unique stack prefix (`tfsaperm`).**
  Both `TestAccGrafanaServiceAccountFromCloud` and `..._AssignRoleOrPermissions` used prefix `"tfsatest"`. The shared `PreCheck` calls `testAccDeleteExistingStacks(t, prefix)`, which would delete the other test's stack mid-run when both ran concurrently.
- **`resource_cloud_plugin_test.go`: assert correct plugin slug in step 2 of `TestAccResourcePluginInstallation`.**
  Step 2 asserted `pluginSlug` (`grafana-googlesheets-datasource`) instead of the actual step 2 plugin (`grafana-clock-panel`).
- **`resource_cloud_plugin_test.go`: keep both resources in step 2 of `TestAccResourcePluginInstallation`.**
  Step 2 had replaced `test-installation` with `test-installation-no-version`, so step 3's `ImportState` on `test-installation` failed.
- **`resource_private_data_source_connect_network_token_test.go`: assert correct PDC resource addresses in `TestResourcePrivateDataSourceConnectNetworkToken_Basic`.**
  Test referenced `grafana_cloud_stack.current.id` (managed resource) instead of `data.grafana_cloud_stack.current.id` (data source).
- **`resource_cloud_stack_test.go`: match expected error across newlines in `TestResourceStack_Basic`.**
  Regex needed `(?s)` to span newlines in the error output.
- **Remove private connectivity assertions from `TestResourceStack_Basic`.** The `private_connectivity_*` attributes are not populated on ephemeral test stacks.
- **Remove `sm_url` assertions** from `TestResourceStack_Basic` and `TestAccDataSourceStack_Basic`. Not populated on ephemeral test stacks.
- **Fix slug overflow in `TestAccGrafanaServiceAccountRotatingTokenFromCloud`.**
  Prefix `"tfsarottoken"` (12) + random(10) + `"-recreated"` (10) = 32 chars, exceeding the 29-char slug max. Shortened prefix to `"tfsarot"` and suffix to `"v2"`.
- **Add `delete_protection = false` to test stack configs.** Test stacks were inheriting delete-protection-enabled defaults, causing cleanup failures and zombie stacks that exhaust the org's stack quota.
- **Lint fixes** — regex raw-string literals, gosec nolint directives.
- **Regenerate `replace_references.go`** after the data source reference fix.


nachogiljaldo (approved):
LGTM, thanks!

Duologic:
I should have made this a draft, still working on it.

thomson:
Heads up — not blocking. The `cloud-acc-tests` retry-on-conflict logic trips our grafana.com API alerts: we enforce a 30-second cooldown on recreation of recently-deleted instances, and these tests recreate the same slug well inside that window.

Two steps in `TestResourceStack_Basic`:

- [Step 3](https://github.com/grafana/terraform-provider-grafana/blob/fix-cloud-acc-tests/internal/resources/cloud/resource_cloud_stack_test.go#L66-L72) — `Taint` destroys+creates back-to-back, no sleep.
- [Step 4](https://github.com/grafana/terraform-provider-grafana/blob/fix-cloud-acc-tests/internal/resources/cloud/resource_cloud_stack_test.go#L73-L88) — sleeps only 10s between out-of-band delete and recreate.

Easiest fix: bump the step 4 sleep to ≥35s and add an equivalent `PreConfig` sleep before step 3. Happy to PR if useful.

Duologic:
Step 3 is trickier as this is testing specific TF behavior, it is inherently in conflict with the 30s cooldown.

Duologic:
Covered the above by handling 409 conflicts response body.

rooneyshuman (approved):

PR #2736, opened 2026-04-27
fix(slo): allow empty fastburn {} / slowburn {} blocks to round-trip without drift

### LINKED ISSUE #2735: bug(slo): allow empty fastburn {} / slowburn {} blocks to round-trip without drift
4.33.0

### Affected Resource(s)
grafana_slo (resource)

### Summary                                                

Provider v4.33.0 has `grafana_slo resources` whose HCL declares empty `fastburn {}` and/or `slowburn {}` blocks fail to apply with:
```
                                                                                                                                          
  .alerting[0].fastburn: block count changed from 1 to 0                                                                                  
  .alerting[0].slowburn: block count changed from 1 to 0
                                                                                                          
```                                
The same configuration applied cleanly before #2527. terraform plan also incorrectly shows the blocks as + create even though they are  present in HCL and were previously materialized in state.
                                                                                                                                          
### Reproduction 
  ```
  resource "grafana_slo" "example" {                                                                                                      
    name        = "Example SLO"                          
    description = "..."
                                                                                                                                          
    destination_datasource { uid = "grafanacloud-prom" }
    objectives             { value = 0.995, window = "28d" }                                                                              
    query {                                                                                                                               
      type = "freeform"
      freeform { query = "..." }                                                                                                          
    }                                                    

    alerting {
      fastburn {}
      slowburn {}
    }
  }
```

  Run terraform plan then terraform apply. Plan output:                                                                                   
  ```
  ~ alerting {                                                                                                                            
    + fastburn {}                                        
    + slowburn {}
  }                                                                                                                                       
  ```

  Apply output:                                                                                                                           
```                                                         
  When applying changes to grafana_slo.example, provider produced an unexpected
  new value: .alerting[0].fastburn: block count changed from 1 to 0.                                                                      
```
                                                                                                                                          
### Expected behavior                                                                                                                       
apply succeeds. The follow-up plan is empty (no drift). This is the pre-#2527 behavior.                                                 
  
### Root cause                                                                                                                              
`convertAlertingToModel` in `internal/resources/slo/data_source_slos.go` filters out any returned fastburn/slowburn whose Labels, Annotations, and Enrichments are all empty, via the `isAlertingMetadataEmpty` helper. 
When the user's HCL declares `fastburn {}`/`slowburn {}`, the pack sends an empty shell, the API echoes an empty shell on GET, and the filter strips it from state — producing a `length-1 → length-0` mismatch that the Plugin Framework consistency check rejects.

The filter was added in #2527 to mitigate a separate problem: the SLO API GET handler returns empty fastburn/slowburn shells even when the PUT did not include them, which previously caused perpetual terraform plan drift for users who omit those blocks from HCL. The filter is correct for that case but wrong for users who explicitly declared the blocks — both cases hit the same code path.             
                                                         
### Affected versions
Any release containing #2527.                                                                                                           
  
### Workaround                                                                                                                              
Pin the provider to a version that predates #2527 until a fix lands.

### PR DESCRIPTION
**Pull request by elainevuong**
### Summary
  - Removes the `isAlertingMetadataEmpty` filter from `convertAlertingToModel` so the resource's READ no longer strips API-returned empty `fastburn` / `slowburn` shells out of state.
  - Restores pre-#2527 behavior for users whose HCL contains alerting `{ fastburn {} slowburn {} }` — apply succeeds again, no more block count changed from 1 to 0 consistency error.   
  - Adds an acceptance-test step exercising the previously-broken shape, plus a follow-up PlanOnly step that asserts no drift on the next plan.                                                                                                                                                                                                                        
                                                         
~~The filter that's being removed was originally added to mitigate a separate, pre-existing problem: the SLO API GET handler returns empty fastburn/slowburn shells even when the PUT did not include them, causing perpetual terraform plan drift for users who omit those blocks. As such, the workaround adopted by most users is `{ fastburn {} slowburn {} }`~~

~~Removing the filter here re-exposes that drift — matching pre-#2527 behavior exactly. The proper fix is on the SLO API side (return exactly what was PUT, no shell injection); that work is [tracked](https://github.com/grafana/slo/issues/4284) separately. Once it lands, the omit-block drift is resolved at the source for new and re-PUT SLOs, and no further provider change will be needed.~~

Fixes #2735

### DIFF
--- a/internal/resources/slo/data_source_slos.go
+++ b/internal/resources/slo/data_source_slos.go
@@ -501,13 +501,11 @@ func convertAlertingToModel(apiAlerting *slo.SloV00Alerting) []alertingModel {
 		Annotation: convertLabelsToModel(apiAlerting.Annotations),
 	}
 
-	// Convert FastBurn — treat API-returned empty metadata as absent to avoid phantom blocks
-	if apiAlerting.FastBurn != nil && !isAlertingMetadataEmpty(apiAlerting.FastBurn) {
+	if apiAlerting.FastBurn != nil {
 		alerting.FastBurn = convertAlertingMetadataToModel(apiAlerting.FastBurn)
 	}
 
-	// Convert SlowBurn — same treatment as FastBurn
-	if apiAlerting.SlowBurn != nil && !isAlertingMetadataEmpty(apiAlerting.SlowBurn) {
+	if apiAlerting.SlowBurn != nil {
 		alerting.SlowBurn = convertAlertingMetadataToModel(apiAlerting.SlowBurn)
 	}
 
@@ -521,17 +519,6 @@ func convertAlertingToModel(apiAlerting *slo.SloV00Alerting) []alertingModel {
 	}
 
 	return []alertingModel{alerting}
-}
-
-// isAlertingMetadataEmpty returns true when the API-returned metadata contains
-// no user-visible data (no labels, annotations, or enrichments). The SLO API
-// always returns non-nil fastburn/slowburn objects even when the user's config
-// did not specify them, so we need to treat those empty shells as absent.
-func isAlertingMetadataEmpty(meta *slo.SloV00AlertingMetadata) bool {
-	if meta == nil {
-		return true
-	}
-	return len(meta.Labels) == 0 && len(meta.Annotations) == 0 && len(meta.Enrichments) == 0
 }
 
 func convertAlertingMetadataToModel(meta *slo.SloV00AlertingMetadata) []alertingMetadataModel {
--- a/internal/resources/slo/resource_slo_test.go
+++ b/internal/resources/slo/resource_slo_test.go
@@ -93,6 +93,25 @@ func TestAccResourceSlo(t *testing.T) {
 				),
 			},
 			{
+				// Tests that empty fastburn {} / slowburn {} blocks round-trip cleanly.
+				// Regression coverage for the "block count changed from 1 to 0" error
+				// that occurred when the unpack stripped API-returned empty metadata
+				// shells out of state while the user's HCL still declared the blocks.
+				Config: emptyBurnAlert(randomName + " - Empty Burn Alerting Check"),
+				Check: resource.ComposeTestCheckFunc(
+					testAccSloCheckExists("grafana_slo.empty_burn_alert", &slo),
+					testAlertingExists(true, "grafana_slo.empty_burn_alert", &slo),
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by ellisda — APPROVED** [review]
LGTM - We want to be able to GET our PUTs and store the following in tfstate if that's what user sent to SLO API

```
alerting {
		fastburn {}
		slowburn {}
	  }
```

We'll fix SLO API in the near future for the edge case where users tried to PUT only the slow burn (note: it's not supported to disable the fast burn alert by omitting this field, so the client-side fix of sending both fast and slow fields should be good enough for now)
```
alerting {
		slowburn {}
	  }
```

ellisda:
@elainevuong, I think I misunderstood some things - I've updated https://github.com/grafana/slo/issues/4284#issuecomment-4339268416

I think our API was already supporting "Get your PUTs" with respect to `fastBurn` and `slowBurn` fields. if you omit the `fastBurn` entirely in a PUT, it is omitted from the GET response. 

Since we have 3+ layers in the round trip here - terraform -> api model -> api resource-> k8s_storage I want to make sure we're fixing the right parts. I appreciate that your PR removes the misleading `The SLO API always returns non-nil fastburn/slowburn objects even when the user's config did not specify them` comment :+1: so maybe PR is good in current state and needs no edits.

I think I need to understand more fully how:
* (request) terraform converts from resource (HCL) into the PUT payload
* (response) how the GET response gets converted into whatever gets compared to the tfstate

I'll take no offense if you go around me and get approval from someone else. I'll dig back into this first thing tomorrow, as I recognize the urgency. I just don't want to play whack-a-mole with things I don't understand.

ellisda:
cloud acceptance tests - see manual invocation at https://github.com/grafana/terraform-provider-grafana/actions/runs/25089343216

ellisda (approved):
this time I tested it with local build of terraform provider, but worth a look at failed https://github.com/grafana/terraform-provider-grafana/actions/runs/25089343216

repro: deploy a SLO with `fastBurn: {}` then apply another change that removes `fastBurn` - see the deletion in image and then a clean plan

Image

elainevuong:
I looked through the cloud acceptance tests, but I think they're failing due to a Grafana issue - not something related to our specific SLO tests.

I think your repro as well validates that this PR is working as expected. I tested with different variations - no alerting blocks, only fastburn, only slowburn, fastburn and slowburn blocks defined - everything appears to be working as expected with no issues.

If you're comfortable, I'll merge this and I can kick off a release that also includes https://github.com/grafana/terraform-provider-grafana/pull/2744

PR #2744, opened 2026-04-29
fix(slo): empty optional attributes cause inconsistent-result-after-apply

### LINKED ISSUE #2743: SLO ratio queries with group_by_labels = [] cause drift and inconsistent-result-after-apply
v4.32.0 and later

### Affected Resource(s)
`grafana_slo` (ratio query type)

### Description
Since v4.32.0 the `grafana_slo` resource was migrated to the Terraform Plugin Framework (#2527). The new code path round-trips an empty `group_by_labels` list inconsistently:

- Refresh / drift detection adds `+ group_by_labels = []` on every plan for SLOs whose API representation has no labels.
- `terraform apply` fails with:

  ```

  When applying changes to <addr>, provider produced an unexpected
  new value: .query[0].ratio[0].group_by_labels:
  was cty.ListValEmpty(cty.String), but now null.
  ```

The PUT itself succeeds — the SLO is created/updated — but every subsequent plan/apply re-triggers the same error, so the resource is effectively wedged for any future change.

### Root Cause
The SLO API marshals `groupByLabels` with Go's `encoding/json` `omitempty`, so a PUT body containing `"groupByLabels": []` round-trips to a GET response with the field absent. The OpenAPI client decodes that absence as a nil slice, which `types.ListValueFrom(ctx, StringType, nil)` converts to a *null* `types.List`. Meanwhile a HCL value of `group_by_labels = []` plans as an *empty list*. Terraform Core then raises the inconsistency error.

### Reproduction
```hcl
resource "grafana_slo" "example" {
  name        = "example"
  description = "example"
  query {
    type = "ratio"
    ratio {
      success_metric  = "success_total"
      total_metric    = "total"
      group_by_labels = []   # <-- triggers the failure
    }
  }
  objectives { value = 0.99  window = "30d" }
  destination_datasource { uid = "..." }
}
```

### Workarounds (until the fix ships)
1. **Drop `group_by_labels = []`** from HCL — omit the attribute entirely. The API treats absent and explicit-empty identically, so plan and state both become null and the error goes away.
2. **Pin `grafana/grafana` to `~> 4.31`** in your provider requirements until the fix releases. The pre-migration code path round-trips `group_by_labels = []` cleanly.
3. *(Last resort)* `lifecycle { ignore_changes = [query[0].ratio[0].group_by_labels] }` on affected resources — clears the error but suppresses drift detection for that field permanently.

### Related
- Sibling regression for empty `fastburn {}` / `slowburn {}` blocks: #2736.
- Long-term cure is strict GET-your-PUTs on the SLO API.

ellisda:
## Summary

Three attribute-level fixes for post-#2527 (Plugin Framework migration) regressions where Optional values silently lose shape between HCL config and API state, leaving the resource wedged with `Provider produced inconsistent result after apply` errors.

| Attribute | Failure | Fix mechanism |
|---|---|---|
| `query.ratio.group_by_labels` (`[]string`) | `omitempty` collapses nil and empty list identically on the wire — `[]` PUT round-trips to GET-omitted, read returns null, plan said `[]` → mismatch | Converge on non-null empty list: `EmptyListForNullConfig` plan modifier + backward promotion of nil API responses to `[]string{}` |
| `folder_uid` (`*string`) | Pack drops `""` silently, server-side `fromFolderSpec` strips `*Uid == ""` regardless — there's no state shape that round-trips | Reject `""` at validate time with `stringvalidator.LengthAtLeast(1)`; tells user to omit the attribute |
| `search_expression` (`*string`) | Same shape as `folder_uid` | Same validator |

`uuid` was inspected and intentionally left alone — it's `Optional+Computed` with API auto-generation semantics, where empty means "let the API pick one" by design.

Closes #2743.

## Why two different mechanisms

The wire-layer Go types are different:

- `[]string` with `omitempty`: nil and `[]string{}` both serialize as omitted, indistinguishable on the wire. Can't fix from the API side without pointer types or a presence wrapper (tracked in grafana/slo#4284). Has to be normalized on the provider side.
- `*string` with `omitempty`: only nil omits; `&""` would emit. But the SLO API server's `fromFolderSpec` (and friends) actively strip `*Uid == ""` server-side, so even faithful Go marshalling can't round-trip an empty UID. Provider can't normalize toward a state shape that survives the round-trip — the only correct posture is to refuse the input upfront.

The earlier design pass tried "collapse `[]` config to null plan" via plan modifier + `Computed: true`. Terraform Core's plan-validity rule (`plan == config or unknown` for any non-null config value, even on Computed attributes) rejects that. Final design inverts: collapse null config to empty-list plan, and read promotes nil API responses to `[]`. Plan, request body, response body, and state all converge on a single non-null shape. See commit `6b812ac0` for the full discovery.

## Behavior across HCL shapes (`group_by_labels`)

| User HCL | Plan | API effect | State |
|---|---|---|---|
| `["a"]` | `["a"]` | PUT `["a"]` | `["a"]` |
| `[]` | `[]` (no-op for non-null config) | PUT omits via `omitempty` | `[]` (promoted from nil) |
| omitted | `[]` (modifier rewrites null→`[]`) | PUT omits | `[]` |
| `["a"]` then user removes line | `[]` (modifier rewrites null→`[]`) | PUT omits, **clears server-side** | `[]` |

Removing a populated `group_by_labels` from HCL still clears the value server-side — user intent flows through end-to-end.

## Related

- Sibling regression for empty `fastburn {}` / `slowburn {}` blocks shipping in #2736 — same release recommended.
- Long-term cure on the API side (strict GET-your-PUTs via pointer types) tracked in grafana/slo#4284. Once that lands, the `group_by_labels` plan modifier and backward promotion can be deleted; the validators on `folder_uid` / `search_expression` would still be the right answer because they reject objectively invalid input.

## Test plan

- [x] `go test ./internal/resources/slo/ -run TestUnit -v` — all 8 cases pass (3 convertQueryToModel cases, 3 plan-modifier cases, 2 round-trip cases)
- [x] `go vet ./...` clean
- [x] Local `terraform plan` with `group_by_labels = []` succeeds (previously errored with the inconsistency)
- [x] Local `terraform plan` with `folder_uid = ""` rejected at validate time with `Attribute folder_uid string length must be at least 1, got: 0` (previously errored after apply with `was cty.StringVal(""), but now null`)
- [ ] Follow-up: add an `_acc_*.tf` example + acceptance step in `TestAccResourceSlo` exercising both `group_by_labels = []` and `folder_uid` validator paths against a live SLO API. Suggest landing in a follow-up so this PR can ship for the open customer escalation.


internal/resources/slo/data_source_slos.go
```diff
@@ -422,7 +422,17 @@ func convertQueryToModel(ctx context.Context, apiQuery slo.SloV00Query) ([]query
...
+			// list and a PUT with the field absent are indistinguishable on
+			// GET (both arrive here as a nil slice). Promote nil to an empty
+			// slice so the state shape matches the EmptyListForNullConfig plan
+			// modifier on the schema, which converges null/empty/populated
+			// config values onto a non-null plan value.
+			labels := apiQuery.Ratio.GroupByLabels
+			if labels == nil {
+				labels = []string{}
+			}
+			groupByLabels, d := types.ListValueFrom(ctx, types.StringType, labels)
```
joeblubaugh, line 437: Is it better/worse/the same to demote an empty list to `nil`?

internal/resources/slo/plan_modifiers.go
```diff
@@ -0,0 +1,52 @@
+package slo
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// emptyListForNullConfig coerces a null config value to an empty list in the
```
joeblubaugh: s/null/nil throughout for Go code

joeblubaugh (commented):
Need to regenerate the docs, and I think all the Nulls in comments should be nils.

internal/resources/slo/data_source_slos.go
ellisda: Worse — would re-introduce the original bug. If we demote empty → nil before `ListValueFrom`, the result is a nil `types.List` in state. The plan modifier on the schema converges nil/empty/populated config onto a non-nil empty-list plan, so state-nil would mismatch plan-empty-list and fire the post-apply consistency error all over again.

Going the other direction (promoting nil → empty slice) is the only way to keep the read-side state shape aligned with the plan-side modifier. Expanded the comment to spell that out in d8025c4b.

<sub>🤖 Drafted with [Claude Code](https://claude.com/claude-code)</sub>

internal/resources/slo/plan_modifiers.go
ellisda: Done in d8025c4b — converted prose `null` → `nil` across `plan_modifiers.go`, `data_source_slos.go`, and the test file. Kept `null` only where it quotes terraform Core's wire-level error text (`cty.NullVal(...)`, "but now null") since that's not Go-side language.

<sub>🤖 Drafted with [Claude Code](https://claude.com/claude-code)</sub>

internal/resources/slo/plan_modifiers.go
```diff
@@ -0,0 +1,52 @@
...
+// (even with `Computed: true`). The only direction that satisfies the
+// rule for both shapes is to make every config — nil, empty, populated —
+// converge on a non-nil plan value, which is what this modifier does for
+// the nil case. The backward (read) path mirrors by promoting nil API
+// responses to an empty list, keeping the state shape stable.
+//
+// The schema must declare the attribute `Computed: true` so the framework
+// permits the modifier to fill in a value when config is nil. When config
+// is `[]` or populated, terraform-Core's rule (`plan == config`) holds
+// trivially and the modifier is a no-op.
```
elainevuong, line 30: praise: inventive solution to force a nil config to an empty list at the plan stage as well here.

internal/resources/slo/resource_slo.go
```diff
@@ -123,11 +123,17 @@ Resource manages Grafana SLOs (Service Level Objectives).
 			},
 			"folder_uid": schema.StringAttribute{
```
elainevuong, line 124: I tested this. When the `stringvalidator.LengthAtLeast(1)` runs, the corresponding error message does not inform the user that they can omit the attribute entirely to associate the SLO to the default Grafana SLO folder.

Other note is that - perhaps "leave the SLO unfoldered" could be worded a bit more explicitly with "omit the folder_uid attribute entirely to insert the SLO into the default Grafana SLO folder"

The same error message (as below) appears if `search_expression` is a `""` - no mention in the error message about how the field could also be omitted entirely at all. 

image

---

Suggestion: We can use the existing `nonEmptyStringValidator` function that already exists, and augment it to accept an additional `message` field so that we can provide custom error messaging if the field is empty. Example below: 


```
			"folder_uid": schema.StringAttribute{
				Optional:    true,
				Description: "UID for the SLO folder. Must be non-empty if set; omit the attribute entirely to associate SLO with the default Grafana SLO folder.",
				Validators: []validator.String{
					nonEmptyStringValidator{
						fieldName: "folder_uid",
						message:   "folder_uid must be non-empty if set; omit the attribute entirely to associate SLO with the default Grafana SLO folder",
					},
				},
			},
			"search_expression": schema.StringAttribute{
				Optional:    true,
				Description: "The name of a search expression in Grafana Asserts. Must be non-empty if set; omit the attribute entirely to leave it unset. This is used in the SLO UI to open the Asserts RCA workbench and in alerts to link to the RCA workbench.",
				Validators: []validator.String{
					nonEmptyStringValidator{
						fieldName: "search_expression",
						message:   "search_expression must be non-empty if set; omit the attribute entirely to leave it unset",
					},
				},
```

L584-L618:
```
// nonEmptyStringValidator rejects empty strings. By default it produces the
// SDKv2-era message "<fieldName> must be a non-empty string" so existing tests
// pass; callers can override the detail by setting message.
type nonEmptyStringValidator struct {
	fieldName string
	message   string
}

func (v nonEmptyStringValidator) detail() string {
	if v.message != "" {
		return v.message
	}
	return v.fieldName + " must be a non-empty string"
}

func (v nonEmptyStringValidator) Description(_ context.Context) string {
	return v.detail()
}

func (v nonEmptyStringValidator) MarkdownDescription(_ context.Context) string {
	return v.detail()
}

func (v nonEmptyStringValidator) ValidateString(_ context.Context, req validator.StringRequest, resp *validator.StringResponse) {
	if req.ConfigValue.IsNull() || req.ConfigValue.IsUnknown() {
		return
	}
	if req.ConfigValue.ValueString() == "" {
		resp.Diagnostics.AddAttributeError(
			req.Path,
			"Invalid Attribute Value",
			v.detail(),
		)
	}
}
```

---

Error message would now appear as follows:

Image

internal/resources/slo/resource_slo_internal_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+// TestUnit_groupByLabels_roundTripIsConsistent ties the two halves together.
+// All three HCL forms — populated, `[]`, omitted — must produce a non-nil
+// empty-or-populated state value that compares equal to the resolved plan,
+// otherwise terraform raises the inconsistent-result error. The forward
+// `pack` path drops empty lists at omitempty serialization time, so the API
+// stores nothing and GET returns nil; the backward `convertQueryToModel`
+// promotes that nil to `[]` and the schema's plan modifier mirrors it for
+// nil config.
+func TestUnit_groupByLabels_roundTripIsConsistent(t *testing.T) {
```
elainevuong, line 128: praise: I appreciate the thorough test cases here to ensure this doesn't regress. Can we add a similar testing for the `folder_uid` and `search_expression` as well?

elainevuong (approved):
Thanks for hopping in on this fix, much appreciated. Tested with all scenarios for the `groupByLabels` (omitted, with an empty list, and with an explicit null) and things work as expected. Left a comment on how we can make the error messaging more explicit with regards to the `folder_uid` attribute.

internal/resources/slo/plan_modifiers.go
ellisda: Thanks! The trick was realizing the modifier needs to read `ConfigValue` not `PlanValue` — the latter falls back to state for Computed attributes when config is null, which masked the user's true intent. Spelled out the design discovery in commit 6b812ac0 if you want the longer story.

<sub>🤖 Drafted with [Claude Code](https://claude.com/claude-code)</sub>

internal/resources/slo/resource_slo.go
ellisda: Good catch — applied your suggested approach in e040cca6:
- Augmented `nonEmptyStringValidator` with an optional `message` field (default behavior unchanged for `destination_datasource.uid`).
- Switched `folder_uid` and `search_expression` to the augmented validator with the custom messages you proposed.
- Updated the `folder_uid` description to "associate the SLO with the default Grafana SLO folder" rather than the vaguer "leave the SLO unfoldered."

<sub>🤖 Drafted with [Claude Code](https://claude.com/claude-code)</sub>

internal/resources/slo/resource_slo_internal_test.go
ellisda: Added in e040cca6 — `TestUnit_nonEmptyStringValidator` covers six cases: empty string with default message, empty string with custom message for `folder_uid` (asserts error mentions "associate the SLO with the default Grafana SLO folder"), empty string with custom message for `search_expression` (asserts error mentions "omit the attribute entirely to leave it unset"), populated string passes, null passes, unknown passes.

<sub>🤖 Drafted with [Claude Code](https://claude.com/claude-code)</sub>

internal/resources/slo/data_source_slos.go
ellisda: > demote empty list to `nil`

Good suggestion. In fact, that is what claude-and-I tried first, but still got errors on the round trip. Something to do with you can't override ("compute") the Plan side when user does explicitly set the `groupBy = []`


Better explanation from claude at https://github.com/grafana/terraform-provider-grafana/blob/e040cca6f464b97c9e0cdbea343a8967e273d0e1/internal/resources/slo/resource_slo_internal_test.go#L68-L71

internal/resources/slo/resource_slo_internal_test.go
ellisda: added `TestUnit_schemaWiresUpEmptyStringValidators` to explicitly check that schema fields have validator and attempts to use a empty-value fail validation

PR #2763, opened 2026-05-08
fix(cloud_stacks): failfast instead of timing out on existing stack error

nachogiljaldo:
Currently, if a stack exists and one tries to re-create it, this times out. This fails fast instead of timing out.

### DIFF
--- a/internal/resources/cloud/resource_cloud_stack.go
+++ b/internal/resources/cloud/resource_cloud_stack.go
@@ -367,6 +367,14 @@ func createStack(ctx context.Context, d *schema.ResourceData, client *gcom.APICl
 				time.Sleep(waitTime)
 				return retry.RetryableError(err)
 			}
+			// Slug is taken by an existing stack — fail outright instead of possibly adopting via GetInstance below.
+			existing, _, getErr := client.InstancesAPI.GetInstance(ctx, stack.Slug).Execute()
+			if getErr == nil && existing != nil && existing.Status != "deleted" {
+				return retry.NonRetryableError(fmt.Errorf(
+					"cannot create Grafana Cloud stack: slug %q is already used by an existing stack (id %v)",
+					stack.Slug, existing.Id,
+				))
+			}
 			return retry.NonRetryableError(err)
 		case err != nil:
 			// If we had an error that isn't a a conflict error (already exists), try to read the stack
--- a/internal/resources/cloud/resource_cloud_stack.go
+++ b/internal/resources/cloud/resource_cloud_stack.go
@@ -348,7 +348,7 @@ func createStack(ctx context.Context, d *schema.ResourceData, client *gcom.APICl
 	}
 	if existing != nil && existing.Status != "deleted" {
 		existingStackError := fmt.Errorf(
-			"cannot create Grafana Cloud stack: slug %q is already used by an existing stack",
+			"That URL has already been taken, please try an alternate URL: %s",
 			stack.Slug,
 		)
 		return apiError(existingStackError)
--- a/internal/resources/cloud/resource_cloud_stack.go
+++ b/internal/resources/cloud/resource_cloud_stack.go
@@ -342,8 +342,8 @@ func createStack(ctx context.Context, d *schema.ResourceData, client *gcom.APICl
 	}
 
 	req := client.InstancesAPI.GetInstance(ctx, stack.Slug)
-	existing, _, getErr := req.Execute()
-	if getErr != nil {
+	existing, httpResp, getErr := req.Execute()
+	if getErr != nil && httpResp != nil && httpResp.StatusCode != http.StatusNotFound {
 		return apiError(getErr)
 	}
 	if existing != nil && existing.Status != "deleted" {
--- a/internal/resources/cloud/resource_cloud_stack.go
+++ b/internal/resources/cloud/resource_cloud_stack.go
@@ -342,6 +342,9 @@ func createStack(ctx context.Context, d *schema.ResourceData, client *gcom.APICl
 
 	req := client.InstancesAPI.GetInstance(ctx, stack.Slug)
 	existing, _, getErr := req.Execute()
+	if getErr != nil {
+		return apiError(getErr)
+	}
 	if getErr == nil && existing != nil && existing.Status != "deleted" {
 		existingStackError := fmt.Errorf(
 			"cannot create Grafana Cloud stack: slug %q is already used by an existing stack",
--- a/internal/resources/cloud/resource_cloud_stack.go
+++ b/internal/resources/cloud/resource_cloud_stack.go
@@ -348,7 +348,7 @@ func createStack(ctx context.Context, d *schema.ResourceData, client *gcom.APICl
 	}
 	if existing != nil && existing.Status != "deleted" {
 		existingStackError := fmt.Errorf(
-			"That URL has already been taken, please try an alternate URL: %s",
+			"could not create stack: That URL has already been taken, please try an alternate URL: %s",
 			stack.Slug,
 		)
 		return apiError(existingStackError)
--- a/internal/resources/cloud/resource_cloud_stack.go
+++ b/internal/resources/cloud/resource_cloud_stack.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
 	"net/http"
 	"net/url"
 	"regexp"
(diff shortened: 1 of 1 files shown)


internal/resources/cloud/resource_cloud_stack.go
```diff
@@ -340,6 +340,16 @@ func createStack(ctx context.Context, d *schema.ResourceData, client *gcom.APICl
 		DeleteProtection: *gcom.NewNullableBool(common.Ref(d.Get("delete_protection").(bool))),
 	}
 
+	req := client.InstancesAPI.GetInstance(ctx, stack.Slug)
+	existing, _, getErr := req.Execute()
+	if getErr == nil && existing != nil && existing.Status != "deleted" {
```
TrueWill: What if `getErr != nil`? Does this assume that means the instance does not exist?

nachogiljaldo: In that case we would end up with the same result because it would try to create it, get a 409 and exit on line 384, but good catch none the less, I will handle this here

nachogiljaldo: addressed, thanks!

internal/resources/cloud/resource_cloud_stack.go
```diff
@@ -340,6 +340,16 @@ func createStack(ctx context.Context, d *schema.ResourceData, client *gcom.APICl
 		DeleteProtection: *gcom.NewNullableBool(common.Ref(d.Get("delete_protection").(bool))),
 	}
 
+	req := client.InstancesAPI.GetInstance(ctx, stack.Slug)
```
TrueWill, line 344: **[MEDIUM · 77]** Pre-flight check is redundant with the existing in-loop 409 handling

<details>
<summary>Details</summary>

There is an inherent TOCTOU race: between `GetInstance` returning "not found" and `CreateStackV1` being issued, another actor can create a stack with the same slug. The pre-flight does not eliminate the need for the in-loop 409 handling, which already handles this correctly.

In practice the pre-flight:
- Adds an extra API round-trip on **every** successful create
- Surfaces a slightly earlier error message when the slug is provably taken before the call
- Provides no correctness guarantee beyond what the in-loop 409 + `GetInstance` check at lines 380–387 already offers

If the pre-flight `GetInstance` call itself fails transiently (500, network error), the check is silently skipped and execution proceeds — so even the "fast fail" property is conditional.

**Suggestion:** Consider dropping the pre-flight block and relying solely on the in-loop 409 + `GetInstance` check, which is the correct place for this logic. If a pre-flight check is retained for UX reasons, document the TOCTOU limitation in a comment.

_(Claude 🤖)_

</details>

internal/resources/cloud/resource_cloud_stack.go
```diff
@@ -340,6 +340,16 @@ func createStack(ctx context.Context, d *schema.ResourceData, client *gcom.APICl
...
+	existing, _, getErr := req.Execute()
```
TrueWill: **[HIGH · 88]** Pre-flight `GetInstance` silently discards non-404 errors

<details>
<summary>Details</summary>

The pre-flight check uses `if getErr == nil && existing != nil` to decide whether a conflict exists. If `GetInstance` returns a transient error (500, network timeout) or an authorization error (403), `getErr` is non-nil and the check is **silently skipped** — the function proceeds to `CreateStackV1` as if the slug were free, with no indication that the safety check itself failed.

```go
existing, _, getErr := req.Execute()
if getErr == nil && existing != nil && existing.Status != "deleted" {
    // 500 / 403 / timeout — silently falls through
```

**Suggestion:** Distinguish 404 ("slug is free") from other errors ("cannot confirm slug is free"). Propagate anything that is not 404:

```go
existing, httpResp, getErr := client.InstancesAPI.GetInstance(ctx, stack.Slug).Execute()
if getErr != nil && (httpResp == nil || httpResp.StatusCode != http.StatusNotFound) {
    return apiError(fmt.Errorf("pre-create slug availability check failed: %w", getErr))
}
if getErr == nil && existing != nil &&
    existing.Status != "deleted" && existing.Status != "deleting" {
    return apiError(fmt.Errorf(
        "cannot create Grafana Cloud stack: slug %q is already used by an existing stack (id %v)",
        stack.Slug, existing.Id,
    ))
}
```

_(Claude 🤖)_

</details>

internal/resources/cloud/resource_cloud_stack.go
TrueWill: **[HIGH · 85]** `"deleting"` status treated as "in use", causing spurious errors during active stack teardown

<details>
<summary>Details</summary>

The condition `existing.Status != "deleted"` treats a stack in `"deleting"` state as still in use and returns an immediate error. But `"deleting"` means the stack is actively being torn down and its slug will be free shortly. The existing `testAccStackCheckDestroy` helper already uses `stack.Status != "deleting"` as the "still alive" predicate — confirming the project treats `"deleting"` as equivalent to `"deleted"` for resource availability purposes.

Consequence: a user who destroys a stack outside Terraform and immediately runs `terraform apply` on a new stack with the same slug receives `"slug already used"` even though the slot is being vacated.

This applies to **both** new checks — the pre-flight here (line 345) and the in-loop check at line 382.

**Suggestion:** Exclude `"deleting"` alongside `"deleted"` in both locations:

```go
if getErr == nil && existing != nil &&
    existing.Status != "deleted" && existing.Status != "deleting" {
```

_(Claude 🤖)_

</details>

internal/resources/cloud/resource_cloud_stack.go
```diff
@@ -340,6 +340,16 @@ func createStack(ctx context.Context, d *schema.ResourceData, client *gcom.APICl
...
+	if getErr == nil && existing != nil && existing.Status != "deleted" {
+		existingStackError := fmt.Errorf(
+			"cannot create Grafana Cloud stack: slug %q is already used by an existing stack",
```
TrueWill: **[MEDIUM · 65]** Pre-flight error omits the existing stack ID that the in-loop error includes

<details>
<summary>Details</summary>

The pre-flight error at this line:
```
"cannot create Grafana Cloud stack: slug %q is already used by an existing stack"
```

The in-loop error at line 384:
```
"cannot create Grafana Cloud stack: slug %q is already used by an existing stack (id %v)"
```

The `(id %v)` is operationally useful — it lets the operator identify exactly which stack owns the slug. The pre-flight omits it even though `existing.Id` is available at that point.

**Suggestion:**

```go
existingStackError := fmt.Errorf(
    "cannot create Grafana Cloud stack: slug %q is already used by an existing stack (id %v)",
    stack.Slug, existing.Id,
)
```

_(Claude 🤖)_

</details>

TrueWill (commented):
## Claude Code Review — #2763

6 findings (2 high, 4 medium) across 1 file | Agents: go-microservices-engineer, general-purpose

> Automated review by [Claude Code](https://claude.ai/code) (`claude-sonnet-4-6[1m]`). Validate all findings before acting on them.

<details>
<summary>General findings (2)</summary>

- **[MEDIUM · 70]** No tests for new fail-fast behavior — acceptance tests require live API access; a unit/mock-level test covering the pre-flight + in-loop paths would help catch regressions (e.g., verify that a 504-recovered create is still adopted correctly by the `case err != nil` path, and that a stack in `"deleting"` state is not incorrectly treated as a blocker).
- **[MEDIUM · 65]** Pre-existing nil-pointer adjacent to this PR — in the `case err != nil` branch (504 recovery), a successful `GetInstance` can set `d.SetId(...)` and return `nil` from the retry closure without ever assigning `stackCreationResponse`. `waitUntilReady` is then called unconditionally and panics at `stack.Id`. A nil guard (`if stackCreationResponse != nil`) before the `waitUntilReady` call would be defensive and correct.

</details>

---

_🤖 [Claude Code](https://claude.ai/code) · `claude-sonnet-4-6[1m]` · [Report issue](https://github.com/anthropics/claude-code/issues)_

TrueWill (dismissed):
Some Claude Code concerns (including general findings as well as line comments) but nothing seems like a hard blocker.

Approving due to the critical nature of the incident.

internal/resources/cloud/resource_cloud_stack.go
nachogiljaldo: I actually want this, because it is not the only potential source of stacks.

internal/resources/cloud/resource_cloud_stack.go
nachogiljaldo: deleted is not treated as in use, it actually continues with the creation does it not?

nachogiljaldo:
> Some Claude Code concerns (including general findings as well as line comments) but nothing seems like a hard blocker.

the 404 was critical, but I tested and I had found it too the others I think they're fine, but there's one I don't understand: https://github.com/grafana/terraform-provider-grafana/pull/2763#discussion_r3210970545 I feel the comment is just wrong but keep me honest here

(also, I will only merge and release it on Monday, definitively not doing it now, but :bow: for the expedite review.)

KucicM (approved):

PR #2764, opened 2026-05-08
feat(grafana): add grafana_teams data source

Duologic:
## Summary

- Add `grafana_teams` (plural) data source to search/list Grafana teams by keyword
- Uses the Plugin Framework SDK with the `SearchTeams` API (`Query` parameter for substring matching)
- Returns team `id`, `name`, `uid`, `email`, `member_count`, and `org_id` for each match
- Supports optional `org_id` for organization-scoped lookups
- Paginates through all results automatically using `TotalCount` from the API response

## Use case

Teams provisioned externally (e.g. via SCIM) are not managed in Terraform, but you may still need to reference them, for example to assign roles or permissions. This data source lets you look up a subset of teams by keyword so you can wire them into your Terraform config without hardcoding IDs.

## Usage

```hcl
# Find all teams matching "platform"
data "grafana_teams" "platform" {
  query = "platform"
}

# Assign a role to each matched team
resource "grafana_role_assignment" "platform_viewer" {
  role_uid = grafana_role.viewer.uid

  dynamic "team" {
    for_each = data.grafana_teams.platform.teams
    content {
      id = team.value.id
    }
  }
}
```


bergquist (approved):

docs/data-sources/teams.md
```diff
@@ -0,0 +1,50 @@
...
+- `id` (String) The ID of this resource.
+- `teams` (List of Object) The list of matching Grafana teams. (see [below for nested schema](#nestedatt--teams))
+
+<a id="nestedatt--teams"></a>
+### Nested Schema for `teams`
+
+Read-Only:
+
+- `email` (String)
+- `id` (Number)
```
mgyongyosi, line 46: Do we still need the `id`? Considering that the legacy ID should be deprecated soon and the APIs (should be) working with the UID out-of-the-box.

Duologic: https://registry.terraform.io/providers/grafana/grafana/latest/docs/resources/role_assignment_item still expects an `team_id`

mgyongyosi (approved):

PR #2765, opened 2026-05-08
fix(cloud_stacks): transform waiting until stack is ready on a best effort if wait_for_readiness is set to false

nachogiljaldo:
This addresses the following situation:
- stack is created successfully
- but it never becomes ready (or something else happens, like rate limiting, etc.)

In that scenario the stack does not end up in the state, so it has to be imported manually.


TrueWill (dismissed):
LGTM

KucicM (approved):

PR #2766, opened 2026-05-08
fix(oncall): guard against nil elements in labelsSetEqual

jwelch92:
## Summary
- `labelsSetEqual` panics with `interface conversion: interface {} is nil, not map[string]interface {}`
  when a `labels` or `dynamic_labels` list element is nil at plan time
- The SDK's `readListField` can produce nil elements when `ReadField` returns `Exists: false`
  for a list index during diff computation
- Use comma-ok type assertion on both loops (lines 897 and 904) so a nil element returns `false`
  (don't suppress the diff) instead of crashing the plugin
- Bug was introduced in #2536 (v4.29.2) and is present through current main

## Test plan
- [ ] `go test ./internal/resources/oncall/... -run TestLabelsSetEqual -v` passes, including
  three new cases: nil in first list, nil in second list, both lists all-nil
- [ ] `go build ./...` compiles cleanly

### DIFF
--- a/internal/resources/oncall/resource_integration.go
+++ b/internal/resources/oncall/resource_integration.go
@@ -887,6 +887,7 @@ func labelsDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
 }
 
 // labelsSetEqual compares two label lists as sets of {key, value} pairs, ignoring order and the "id" field.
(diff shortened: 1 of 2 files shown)


olegbespalov (approved):
LGTM 👍

jwelch92:
Please hold, ran into some issues trying to sign commits made before verified commits were made required.

PR #2778, opened 2026-05-14
chore: upgrade go directive to 1.26.3

Duologic:
## Summary
- Bump the `go` directive in `go.mod` from 1.25.9 to 1.26.3
- Bump golangci-lint from v2.5.0 to v2.12.2 (required for Go 1.26 compatibility)
- Disable `goconst` linter (50 noisy findings, low value)
- Fix lint findings surfaced by the new linter version:
  - Use `reflect.Pointer` instead of deprecated `reflect.Ptr`
  - Use `fmt.Fprintf` instead of `WriteString(fmt.Sprintf(...))`
  - Replace deprecated `parser.ParseDir` with `parser.ParseFile`
  - Use `os.Root` for scoped filesystem ops in genreferences tool
  - Suppress gosec false positives in tests and build tools

Closes https://github.com/grafana/deployment_tools/issues/588563
Tracking issue: https://github.com/grafana/deployment_tools/issues/588539

### DIFF
--- a/.golangci.toml
+++ b/.golangci.toml
@@ -5,7 +5,7 @@ default = 'none'
 enable = [
   'copyloopvar',
   'dogsled',
-  'goconst',
+  # 'goconst',
   'gocritic',
   'gocyclo',
   'goprintffuncname',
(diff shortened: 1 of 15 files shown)


.golangci.toml
```diff
@@ -5,7 +5,7 @@ default = 'none'
 enable = [
   'copyloopvar',
   'dogsled',
-  'goconst',
+  # 'goconst',
```
suntala: Why not just remove it?

Duologic:
```suggestion
```

suntala (approved):

PR #2807, opened 2026-06-02
feat: add k6_version field to loadtest resource

cristianoventura:
This PR updates the `loadtest` resource to support the `k6_version` field, which can now be set to a specific major version.

The [k6-cloud-openapi-client-go](https://github.com/grafana/k6-cloud-openapi-client-go) API spec has been generated to v1.12.0.


internal/resources/k6/common.go
```diff
@@ -85,7 +85,7 @@ func (d *basePluginFrameworkDataSource) Configure(_ context.Context, req datasou
 
 // getProjectAllowedLoadZones retrieves the allowed load zones for a project
 // Returns k6_load_zone_ids directly from the API response
-func getProjectAllowedLoadZones(ctx context.Context, client *k6.APIClient, config *k6providerapi.K6APIConfig, projectID int32) ([]string, error) {
+func getProjectAllowedLoadZones(ctx context.Context, client *k6.APIClient, config *k6providerapi.K6APIConfig, projectID int64) ([]string, error) {
```
cristianoventura, line 88: There is a change in the new version of [k6-cloud-openapi-client-go](https://github.com/grafana/k6-cloud-openapi-client-go) that changed the ID attributes from `int32` to `int64` - because the version was bumped in `go.mod`, I'm resolving that conflict in this PR.

internal/resources/k6/resource_loadtest.go
```diff
@@ -101,6 +102,14 @@ func (r *loadTestResource) Schema(_ context.Context, _ resource.SchemaRequest, r
 				Description: "Identifier of a baseline test run used for results comparison.",
 				Optional:    true,
 			},
+			"k6_version": schema.StringAttribute{
+				Description: "Identifier of the k6 version used to run the test. If not set, the k6 API assigns a default.",
```
fornfrey: Would it make sense to say explicitly what would be the default in this case? Also, can you please add an example value in the description? (`v1` or just `1`)

```suggestion
				Description: "Identifier of the k6 version used to run the test. If not set, use the latest major version.",
```

internal/resources/k6/resource_loadtest.go
```diff
@@ -323,31 +349,47 @@ func (r *loadTestResource) Update(ctx context.Context, req resource.UpdateReques
...
 			return
 		}
-		toUpdate.SetBaselineTestRunId(int32(intID))
+		toUpdate.SetBaselineTestRunId(intID)
+	}
+	switch {
+	case plan.K6Version.IsUnknown():
+		// Value is preserved from state via UseStateForUnknown; nothing to send.
+	case plan.K6Version.IsNull():
+		toUpdate.SetK6VersionNil()
```
fornfrey: What happens on the backend when the client sends a null for a k6 version? Aren't all tests with a defined script supposed to have a locked version, too, or is it just an "explicit" lock? Meaning, if not set, do we automatically migrate the test to the latest when a new major version is released?

internal/resources/k6/resource_loadtest.go
cristianoventura: I updated the description and added an example as you suggested. However, I'm using "default" instead of "latest" to reflect the actual behaviour determined by the API.

internal/resources/k6/resource_loadtest.go
cristianoventura: > What happens on the backend when the client sends a null for a k6 version?

It depends on whether it's a create or a patch.

- **Create with null**: the backend never stores null. It resolves a concrete major version (whatever the default is set to in Cloud, currently `1`) and persists that integer. The create response echoes back a non-null version. 
- **Patch with null**: This will unpin the test, using whatever the default is set to in Cloud during execution.

> Aren't all tests with a defined script supposed to have a locked version, too, or is it just an "explicit" lock? Meaning, if not set, do we automatically migrate the test to the latest when a new major version is released?

It's an explicit lock. If no pin is found, the test will resolve to a default major version that is set in Cloud. If the default changes, tests that are not pinned will also change.

fornfrey: > It's an explicit lock. If no pin is found, the test will resolve to a default major version that is set in Cloud. If the default changes, tests that are not pinned will also change.

So, it sounds like it is possible to have an unpinned test definition that is updated to the latest major version on every release. Is this an expected behavior for Terraform-defined tests? IIUC, creating a test in the UI works differently: a test is always pinned at creation and will not be moved to a newer version without user input.

cristianoventura: This is now resolved on the backend. The API no longer accepts a null k6 version on create or update, so it's no longer possible to have an unpinned/floating test definition. Behavior now matches the UI:

- Create without `k6_version`: the backend resolves the current default major version and persists it as a pin. Terraform stores that value in state, so the test stays pinned and won't move to a newer major version on new releases.
- Update: the provider no longer sends null. An omitted k6_version keeps the currently pinned version.

I've updated the provider to drop the unpin path and clarified the attribute description accordingly.

internal/resources/k6/resource_loadtest.go
```diff
@@ -324,19 +350,45 @@ func (r *loadTestResource) Update(ctx context.Context, req resource.UpdateReques
...
 		)
 		return
 	}
-	loadTestID := int32(intID)
+	loadTestID := intID
 
 	// Generate API request body from plan
 	toUpdate := k6.NewPatchLoadTestApiModel()
 	toUpdate.SetName(plan.Name.ValueString())
+	if plan.BaselineTestRunID.IsNull() {
```
fornfrey: Baseline test runs have been deprecated. Looks like the rebase added it back by accident.

fornfrey (approved):
Well done 🙌

PR #2808, opened 2026-06-03
feat: add k6 channel support

VikaCep:
## Summary

- **Bump `synthetic-monitoring-api-go-client` to v0.20.0** 
- **Add `channels` attribute** to `grafana_synthetic_monitoring_check`, allowing users to assign a k6 channel (version) to scripted/browser checks via `channels { k6 { id = "..." } }`
- **Mark `channels` as Optional+Computed** so the API-assigned default channel is preserved when not explicitly set, avoiding drift on scripted/browser checks
- Add acceptance test for the new `channels` attribute

### DIFF
--- a/docs/resources/synthetic_monitoring_check.md
+++ b/docs/resources/synthetic_monitoring_check.md
@@ -430,7 +430,7 @@ resource "grafana_synthetic_monitoring_check" "traceroute" {
 
 - `alert_sensitivity` (String) Can be set to `none`, `low`, `medium`, or `high` to correspond to the check [alert levels](https://grafana.com/docs/grafana-cloud/testing/synthetic-monitoring/configure-alerts/synthetic-monitoring-alerting/). Defaults to `none`.
 - `basic_metrics_only` (Boolean) Metrics are reduced by default. Set this to `false` if you'd like to publish all metrics. We maintain a [full list of metrics](https://github.com/grafana/synthetic-monitoring-agent/tree/main/internal/scraper/testdata) collected for each. Defaults to `true`.
-- `channels` (Block List, Max: 1) Channels to assign the check to. If not specified for scripted/browser checks, the API assigns a default k6 channel. (see [below for nested schema](#nestedblock--channels))
+- `channels` (Block List, Max: 1) Channels to assign the check to. See [Manage k6 versions](https://grafana.com/docs/grafana-cloud/testing/synthetic-monitoring/create-checks/manage-k6-versions/) for details. If not specified for scripted/browser checks, the API assigns a default k6 channel. (see [below for nested schema](#nestedblock--channels))
 - `enabled` (Boolean) Whether to enable the check. Defaults to `true`.
 - `folder_uid` (String) The UID of the Grafana folder to associate the check with.
 - `frequency` (Number) How often the check runs in milliseconds (the value is not truly a "frequency" but a "period"). The minimum acceptable value is 1 second (1000 ms), and the maximum is 1 hour (3600000 ms). Defaults to `60000`.
--- a/docs/resources/synthetic_monitoring_check.md
+++ b/docs/resources/synthetic_monitoring_check.md
@@ -430,6 +430,7 @@ resource "grafana_synthetic_monitoring_check" "traceroute" {
 
 - `alert_sensitivity` (String) Can be set to `none`, `low`, `medium`, or `high` to correspond to the check [alert levels](https://grafana.com/docs/grafana-cloud/testing/synthetic-monitoring/configure-alerts/synthetic-monitoring-alerting/). Defaults to `none`.
 - `basic_metrics_only` (Boolean) Metrics are reduced by default. Set this to `false` if you'd like to publish all metrics. We maintain a [full list of metrics](https://github.com/grafana/synthetic-monitoring-agent/tree/main/internal/scraper/testdata) collected for each. Defaults to `true`.
+- `channels` (Block List, Max: 1) Channels to assign the check to. (see [below for nested schema](#nestedblock--channels))
 - `enabled` (Boolean) Whether to enable the check. Defaults to `true`.
 - `folder_uid` (String) The UID of the Grafana folder to associate the check with.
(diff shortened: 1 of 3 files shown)


docs/resources/synthetic_monitoring_check.md
```diff
@@ -430,6 +430,7 @@ resource "grafana_synthetic_monitoring_check" "traceroute" {
 
 - `alert_sensitivity` (String) Can be set to `none`, `low`, `medium`, or `high` to correspond to the check [alert levels](https://grafana.com/docs/grafana-cloud/testing/synthetic-monitoring/configure-alerts/synthetic-monitoring-alerting/). Defaults to `none`.
 - `basic_metrics_only` (Boolean) Metrics are reduced by default. Set this to `false` if you'd like to publish all metrics. We maintain a [full list of metrics](https://github.com/grafana/synthetic-monitoring-agent/tree/main/internal/scraper/testdata) collected for each. Defaults to `true`.
+- `channels` (Block List, Max: 1) Channels to assign the check to. If not specified for scripted/browser checks, the API assigns a default channel. (see [below for nested schema](#nestedblock--channels))
```
ka3de: I don't think we should state `Max: 1` here. Right now we only support one channel for k6, but the _channels_ field was defined this way so a check could be assigned to other channels too, and these could arguably be unrelated to k6, therefore supporting for more than one.

Should we link to some documentation here so users can understand better what a channel/k6 channel represents?

internal/resources/syntheticmonitoring/resource_check.go
```diff
@@ -789,6 +789,34 @@ multiple checks for a single endpoint to check different capabilities.
 				Type:        schema.TypeString,
 				Optional:    true,
 			},
+			"channels": {
+				Description: "Channels to assign the check to. If not specified for scripted/browser checks, the API assigns a default channel.",
```
ka3de: Similar to my comment above, when I think about channels, and as they were defined, these could be defined for multiple things, not necessarily only for k6. Since right now we only support the k6 channel I would specify that here.
```suggestion
				Description: "Channels to assign the check to. If not specified for scripted/browser checks, the API assigns a default k6 channel.",
```

internal/resources/syntheticmonitoring/resource_check.go
```diff
@@ -789,6 +789,34 @@ multiple checks for a single endpoint to check different capabilities.
...
+				Type:        schema.TypeList,
+				Optional:    true,
+				Computed:    true,
+				MaxItems:    1,
```
ka3de: I wouldn't say this is true. Right now it is because only the k6 channel is supported, but this might not be the case in the future. Should we keep this as `1` by now and update it once more channel types are supported? Or keep this as `1` by now to avoid users specifying multiple k6 channels (although I don't think that's a very probable mistake).

internal/resources/syntheticmonitoring/resource_check.go
```diff
@@ -789,6 +789,34 @@ multiple checks for a single endpoint to check different capabilities.
...
+					Schema: map[string]*schema.Schema{
+						"k6": {
+							Description: "K6 channel configuration.",
+							Type:        schema.TypeList,
+							Optional:    true,
+							Computed:    true,
+							MaxItems:    1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"id": {
```
ka3de, line 810: A channel definition, also as included in the check definition has more fields. E.g.:
```json
"channels": {
		"k6": {
			"id": "v2",
			"name": "v2.x",
			"default": false,
			"manifest": "^2.0.0",
			"deprecatedAfter": "2027-06-01"
		}
	},
```
However check requests for create/update only support the `id` field as it identifies the k6 channel defined.
How should that be represented here? Maybe that's not relevant for Terraform resource definition as the only property that the user has action on is the channel ID?

ka3de (commented):
Not super familiar with the particularities of the Terraform provider, but made a few comments.

docs/resources/synthetic_monitoring_check.md
VikaCep: `Max: 1` here is auto-generated from `MaxItems: 1` in the schema, which is correct. In the protobuf, `Check.Channels` is a single struct (*Channels), not a repeated field, so a check has exactly one `channels` block. If new channel types are added in the future, they'd be new fields inside the Channels struct (e.g. K6 *K6Channel, NewType *NewTypeChannel), which maps to new sibling fields inside the same channels block in Terraform. `MaxItems: 1` won't need changing when that happens.

Added a link to the [Manage k6 versions](https://grafana.com/docs/grafana-cloud/testing/synthetic-monitoring/create-checks/manage-k6-versions/) docs.

internal/resources/syntheticmonitoring/resource_check.go
VikaCep: Only `id` is exposed because it's the only user-actionable field. The other fields (name, manifest, default, deprecatedAfter) are API-enriched metadata returned on read but not settable by users . This matches how the frontend handles it too (it only sends `{ k6: { id } }`).

internal/resources/syntheticmonitoring/resource_check.go
VikaCep: It doesn't limit the number of channel types, it limits the number of channels blocks in the Terraform config:

```
# This would be wrong — two channels blocks on one check
channels {
  k6 { id = "v1" }
}
channels {
  new_type { id = "..." }
}


# This is correct — one channels block with multiple fields inside
channels {
  k6 { id = "v1" }
  new_type { id = "..." }
}
```
New channel types are new fields inside the single channels block, not additional channels blocks. So `MaxItems: 1` is correct and won't change.

docs/resources/synthetic_monitoring_check.md
ka3de: Ah ok, I misunderstood what was that implying.

internal/resources/syntheticmonitoring/resource_check.go
ka3de: Got it, yep, I misunderstood how it worked.

ka3de (approved):

VikaCep:
Merging after `cloud acceptance tests` start passing again: https://github.com/grafana/terraform-provider-grafana/actions/workflows/cloud-acc-tests.yml (unrelated to these changes, caused by this incident https://grafanalabs.enterprise.slack.com/archives/C0B8BEJK10S)

Issue #2253, opened 2025-07-14
Grafana Oncally Integration Dynamic Label creation error


DownWithThisSortOfThing:
### Terraform Version

1.11

### Terraform Grafana Provider Version

3.25.2 -> 3.25.7

### Grafana Version

Grafana Cloud IRM Version 0.37.2

### Terraform Configuration Files

```hcl
resource "grafana_oncall_integration" "alertmanager" {
  name = "alertmanager"

  type    = "alertmanager"
  team_id = data.grafana_oncall_team.on_call.id

  dynamic_labels = [data.grafana_oncall_label.alertmanager_customer]

  templates {
    grouping_key = "{{ payload.customer }}"
  }

  default_route {
    escalation_chain_id = grafana_oncall_escalation_chain.default.id
  }
}

resource "grafana_team" "on_call_team" {
  name    = var.on_call_rotation.team_name
}

data "grafana_oncall_team" "on_call" {
  name = grafana_team.on_call_team.name
}

data "grafana_oncall_label" "alertmanager_customer" {
  key   = "customer"
  value = "{{ payload.get('customer', 'unknown') }}"
}

resource "grafana_oncall_escalation_chain" "default" {
  name = var.schedule_name

  team_id = data.grafana_oncall_team.on_call.id
}
```

### Expected Behavior

The dynamic labels will be created in accordance with the plan (Example plan provided). ID's have been X'd out just to not leak information 😅 
```shell
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # module.an_other_schedule.grafana_oncall_integration.alertmanager will be updated in-place
  ~ resource "grafana_oncall_integration" "alertmanager" {
      ~ dynamic_labels = [
          + {
              + "id"    = "customer"
              + "key"   = "customer"
              + "value" = "{{ payload.get('customer', 'unknown') }}"
            },
        ]
        id             = "XXXXXXXXX"
        name           = "pgt-cmp-schedule-alertmanager"
        # (4 unchanged attributes hidden)

        # (2 unchanged blocks hidden)
    }

  # module.schedule.grafana_oncall_integration.alertmanager will be updated in-place
  ~ resource "grafana_oncall_integration" "alertmanager" {
      ~ dynamic_labels = [
          + {
              + "id"    = "customer"
              + "key"   = "customer"
              + "value" = "{{ payload.get('customer', 'unknown') }}"
            },
[...]

PR #2818, opened 2026-06-09
fix: oncall integrations allow nil labels

willgallego-grafana:
For OnCall integrations, excluding the `labels` or `dynamic_labels` fields on an update should retain existing labels/dynamic_labels respectively, keep in contract with the public API.


roberthorn (approved):

Duologic:
Would this PR also resolve #2253?

internal/resources/oncall/resource_integration.go
```diff
@@ -349,16 +349,22 @@ func resourceIntegrationUpdate(ctx context.Context, d *schema.ResourceData, clie
...
-		Templates:     expandTemplates(templateData),
-		DefaultRoute:  expandDefaultRoute(defaultRouteData),
-		Labels:        expandLabels(labelsData),
-		DynamicLabels: expandLabels(dynamicLabelsData),
+		Name:         nameData,
+		TeamId:       teamIDData,
+		Templates:    expandTemplates(templateData),
+		DefaultRoute: expandDefaultRoute(defaultRouteData),
+	}
+	if labelsSet {
```
Duologic: If a user wants to remove the all labels on this resource, then this will prevent it ever being executed. There is a [d.HasChanges()](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#ResourceData.HasChanges) that might do what you are trying to accomplish.

You might want  to unit test this behavior.

willgallego-grafana:
> Would this PR also resolve #2253?

Unsure. That's from almost a year ago so things have changed a bit since. I'll verify if it's still a issue against the currently PR and either it's fixed by or before it (where we can close that other issue) or it's still an ongoing problem and we can take a look

internal/resources/oncall/resource_integration.go
willgallego-grafana: Hey, that's the intended flow (see the description above).

Originally, there was the public API which you could create integrations and corresponding terraform that uses that. labels/dynamic_labels were added later. We didn't want people who had manually added labels to lose them (because there would be a diff) so it was decided to allow if the TF (and subsequent API request) to not have labels that it wouldn't try to remove them but skip over.

willgallego-grafana:
> > Would this PR also resolve #2253?
> 
> Unsure. That's from almost a year ago so things have changed a bit since. I'll verify if it's still a issue against the currently PR and either it's fixed by or before it (where we can close that other issue) or it's still an ongoing problem and we can take a look

I've just run the terraform from the bug on main (for this and axmir-api-go-client) and it seems to build just fine for me. I'm guessing it's been fixed since, will close it up.

willgallego-grafana:
@Duologic Added some additional unit testing that should solidify the intended behavior

roberthorn (approved):

PR #2819, opened 2026-06-10
test(equivalence): install cli to repo-local .cache/bin

suntala:
For https://github.com/grafana/deployment_tools/issues/585771 (the second bullet point).


Duologic (approved):
It might be useful to share the context in the PR description instead of referring to a private repo, OSS contributors can't access that.

PR #2823, opened 2026-06-10
feat(alerting): add routingtree app platform resource

yuri-tceretian:
## What

Adds `grafana_apps_notifications_routingtree_v1beta1`, a Plugin Framework AppPlatform resource backed by the `notifications.alerting.grafana.app` Kubernetes API. It manages the notification routing tree — the same object as the legacy `grafana_notification_policy`, but via the new app platform API, and supports multiple **named** routing trees (not just the default `user-defined` tree).

## Why

The routing tree is moving to the app-platform K8s API. This gives Terraform users a first-class resource on the new API and a migration path off the legacy provisioning-based `grafana_notification_policy`.

## Notable design points

- **Recursive routes**: modeled as bounded-depth (5) typed blocks, mirroring `grafana_notification_policy`, since Terraform schemas can't be infinitely recursive.
- **Feature flag**: requires Grafana **13.0+** with the `alertingMultiplePolicies` toggle. Added to the docker-compose test instance; the example/acc tests gate at `>=13.0.0`.
- **Provenance**: `disable_provenance` is implemented directly in the routingtree resource (not via a generic `ResourceConfig` field). Setting it to `true` writes the `grafana.app/provenance: none` annotation, allowing the routing tree to be edited from other sources (UI, API). Setting it to `false` (the default) marks it as Terraform-provisioned by writing `grafana.app/provenance: api`.
- **Docs**: the schema section is hand-authored in the doc template so the recursive `routes` block is documented once (instead of ~10 repeated nested sections), plus a guide for migrating from `grafana_notification_policy`.

## Known follow-ups (not in this PR)

- `metadata.folder_uid` is inherited from the generic resource and not applicable here; it's omitted from the docs, but the schema still *accepts* (and ignores) it. Fully removing it needs a generic-resource change.
- The routingtree schema doc is hand-authored, so it won't auto-update on schema changes (flagged in a template comment).


wealthystudent:
hey, we run a grafana cloud 13.1.0 instance with `alertingMultiplePolicies` enabled and have been looking for exactly this. we manage our notification policy via terraform today using `grafana_notification_policy` through a controller that reconciles the full tree, and the main thing we are missing is the ability to let teams configure custom routing (email, slack) outside of the default oncall flow without conflicting with the controller managed tree. named routing trees solves that for us.

happy to test this against our instance once it is ready for feedback. let me know if that would be useful or if there is anything we can help with.

yuri-tceretian:
@wealthystudent thanks for offerring help! It's in draft because we were working on renaming default route name from awkward "user-defined", which used to be meaningful, to "default". We started some work in that direction https://github.com/grafana/grafana/pull/126841. I will test this PR with that change and will make it ready for review.

JacobsonMT (approved):
LGTM 🚀

PR #2824, opened 2026-06-11
fix(ci): replace native concurrency groups with turnstyle and add /run-cloud-tests slash command

Duologic:
## Summary

- Replace native `concurrency` groups on the `cloudinstance` and `cloud` jobs with [`softprops/turnstyle`](https://github.com/softprops/turnstyle) for true FIFO queuing
- Add a `/run-cloud-tests` slash command to trigger cloud acceptance tests from PR comments
- Add workflow-level `cancel-in-progress` to `acc-tests.yml` so new pushes cancel stale runs

## Problem

GitHub Actions concurrency groups only keep **one** pending job per group. When a third run arrives while one is running and one is pending, the pending run is silently cancelled. This causes spurious `cloudinstance` and `cloud` job failures on busy repos.

## Changes

1. **acc-tests.yml**: Remove `concurrency: { group: cloud-instance, cancel-in-progress: false }` from the `cloudinstance` job, add a turnstyle step with `job-to-wait-for: cloudinstance`. Add a workflow-level concurrency group per PR with `cancel-in-progress: true` to cancel stale runs on new pushes.
2. **cloud-acc-tests.yml**: Remove `concurrency: cloud-api` from the `cloud` job, add a turnstyle step with `job-to-wait-for: cloud`.
3. **slash-run-cloud-tests.yml** (new): `issue_comment`-triggered workflow that dispatches cloud acceptance tests when a maintainer comments `/run-cloud-tests` on a PR. Posts back a comment with a link to the triggered run.
4. **comment-on-pr.yml**: Update the info comment to mention the slash command and include a link to the Actions UI.

Both turnstyle steps use a 40-minute `abort-after-seconds` safety net and 30-second poll interval. Since this is a public repo on GitHub-hosted runners, the polling wait has no billing impact.

## Test plan

- [ ] Verify `cloudinstance` job queues correctly when multiple acc-tests runs overlap (turnstyle waits instead of cancelling)
- [ ] Verify `cloud` job queues correctly when multiple cloud-acc-tests runs overlap
- [ ] Verify new pushes to a PR cancel in-flight acc-tests runs
- [ ] Verify `/run-cloud-tests` slash command works after merge (cannot test on this PR — `issue_comment` workflows run from main)
- [ ] Verify PR info comment renders with slash command instructions and Actions UI link

### DIFF
--- a/.github/workflows/comment-on-pr.yml
+++ b/.github/workflows/comment-on-pr.yml
@@ -17,5 +17,5 @@ jobs:
         with:
           message: |
             In order to lower resource usage and have a faster runtime, PRs will not run Cloud tests automatically.
-            To do so, a Grafana Labs employee must trigger the `cloud acceptance tests` workflow manually.
+            To do so, a Grafana Labs employee must [trigger the `cloud acceptance tests` workflow](${{ github.server_url }}/${{ github.repository }}/actions/workflows/cloud-acc-tests.yml) manually for the `${{ github.head_ref }}` branch.
           message-id: pr-cloud-tests-info
--- a/.github/workflows/comment-on-pr.yml
+++ b/.github/workflows/comment-on-pr.yml
@@ -17,5 +17,5 @@ jobs:
         with:
           message: |
             In order to lower resource usage and have a faster runtime, PRs will not run Cloud tests automatically.
-            To do so, a Grafana Labs employee must [trigger the `cloud acceptance tests` workflow](${{ github.server_url }}/${{ github.repository }}/actions/workflows/cloud-acc-tests.yml) manually for the `${{ github.head_ref }}` branch.
+            A Grafana Labs employee can trigger them by commenting `/run-cloud-tests` on this PR, or via the [Actions UI](${{ github.server_url }}/${{ github.repository }}/actions/workflows/cloud-acc-tests.yml) for the `${{ github.head_ref }}` branch.
           message-id: pr-cloud-tests-info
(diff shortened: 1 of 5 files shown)


.github/workflows/comment-on-pr.yml
```diff
@@ -17,5 +17,5 @@ jobs:
         with:
           message: |
             In order to lower resource usage and have a faster runtime, PRs will not run Cloud tests automatically.
-            To do so, a Grafana Labs employee must trigger the `cloud acceptance tests` workflow manually.
+            A Grafana Labs employee can trigger them by commenting `/run-cloud-tests` on this PR, or via the [Actions UI](${{ github.server_url }}/${{ github.repository }}/actions/workflows/cloud-acc-tests.yml) for the `${{ github.head_ref }}` branch.
```
suntala, line 20: Nice!

Duologic:
Note: the `/run-cloud-tests` slash command won't work on this PR yet — `issue_comment`-triggered workflows always run from the default branch (main). It will become functional after this PR is merged.

suntala (dismissed):
Great idea about the comment to trigger the tests 🚀

About turnstyle, it's definitely an improvement. However, given how long it takes for tests to run, it might not be sufficient. For a queue of more than ~2 pending jobs, there will still be jobs which fail (due to ` abort-after-seconds: 2400 # 40 min safety net`). I expect that the long term solution is to make tests run more efficiently.

Duologic:
good call, commit coming in

suntala (approved):

PR #2829, opened 2026-06-12
fix(cloud): honour Retry-After on 429s and retry transient grafana.com API errors in limited call sites

s-gohel:
Part of https://github.com/grafana/grafana-com/issues/18305

Second pass at #2767, which was abandoned because it applied retries to every grafana.com call site at once. This PR takes the same retry machinery but only enables it in **two places** so the behavior can be validated with a small blast radius:

- `grafana_cloud_access_policy` resource (create / read / update / delete)
- `grafana_cloud_organization` data source (read)

## What it does

- Adds `RetryGCOM` (`internal/resources/cloud/gcom_retry.go`), which retries grafana.com API calls on transient errors: 429, 408, 5xx, network errors, and `io.EOF`/`io.ErrUnexpectedEOF`. 409 is deliberately **not** retryable (stack creation keeps its bespoke conflict handling).
- Honours the `Retry-After` header on 429s (both RFC 9110 delta-seconds and HTTP-date formats), capped at 2 minutes. GCOM sends this header from its rate limiters (`lib-api/src/errors.ts`, `lib-go` ratelimiter middleware).
- If `Retry-After` exceeds the remaining retry budget (access-policy write limits use a 1-hour window, so it can be up to ~3600s), the call fails immediately instead of sleeping the cap and timing out anyway.
- `OnlyRetryRateLimited` restricts the create path to 429-only retries: GCOM rate limiting rejects requests **before** proxying to the Auth API, so a 429 is always safe to retry; 5xx is ambiguous for a non-idempotent POST (`X-Request-Id` is CSRF/tracing only, not an idempotency key — verified in grafana-com).
- `TreatNotFoundAsSuccess` makes the access policy delete idempotent: destroying an already-gone policy now succeeds instead of erroring (behavior change).

## Test plan

- [x] Unit tests for Retry-After parsing (delta-seconds, HTTP-date, invalid), 429 retry with wait, 429-only mode, budget-exceeded give-up, 404-as-success, and non-retryable 4xx (`go test ./internal/resources/cloud/ -run TestUnit`)
- [x] Cloud acceptance tests (`TF_ACC_CLOUD_API=true`) for `grafana_cloud_access_policy` and `grafana_cloud_organization`
- [ ] Validate against real rate limiting in a test org before expanding to the remaining call sites

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/resources/cloud/gcom_retry.go
+++ b/internal/resources/cloud/gcom_retry.go
@@ -1,0 +1,199 @@
+package cloud
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"log"
+	"net"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
+)
+
+const (
+	// DefaultGCOMRetryTimeout caps the Grafana Cloud OpenAPI retry loop duration.
+	DefaultGCOMRetryTimeout = 2 * time.Minute
+	gcomRetryAfterCap       = 2 * time.Minute
+)
+
+// GCOMRetryConfig configures RetryGCOM.
+type GCOMRetryConfig struct {
+	Timeout time.Duration
+
+	// TreatNotFoundAsSuccess, when true, treats HTTP 404 responses as success (typically for DELETE),
+	// so destroys are idempotent when the resource is already gone.
+	TreatNotFoundAsSuccess bool
+
+	// OnlyRetryRateLimited, when true, restricts retries to HTTP 429 responses.
+	//
+	// Use this for non-idempotent operations (POST creates): grafana.com rate limiting rejects
+	// requests before they are processed, so a 429 is always safe to retry. 5xx responses and
+	// transport errors are ambiguous — the request may have been processed server-side even though
+	// no response was received — and retrying could create duplicate resources.
+	// When set, DefaultGCOMTransient and OnTransient are not consulted.
+	OnlyRetryRateLimited bool
+
+	// OnTransient, when non-nil, is consulted after DefaultGCOMTransient returns false.
+	OnTransient func(resp *http.Response, err error) bool
+}
+
+func (c GCOMRetryConfig) timeoutDuration() time.Duration {
+	if c.Timeout > 0 {
+		return c.Timeout
+	}
+	return DefaultGCOMRetryTimeout
+}
+
+// RetryGCOM runs op under retry.RetryContext until success or non-retryable error.
+//
+// Follow-up retries:
+//   - For HTTP 429, sleeps according to Retry-After (RFC 9110 delta-seconds or HTTP-date), capped by gcomRetryAfterCap and ctx cancellation.
+//     If the requested wait exceeds the remaining retry budget, the error is returned immediately instead of sleeping and timing out anyway
+//     (grafana.com rate limit windows can be up to an hour, far beyond the retry timeout).
+//   - When TreatNotFoundAsSuccess is set, HTTP 404 is treated as immediate success (common for idempotent DELETE).
+//   - Failed response bodies are drained for connection reuse (success responses are unchanged).
+func RetryGCOM(ctx context.Context, cfg GCOMRetryConfig, op func() (*http.Response, error)) error {
+	timeout := cfg.timeoutDuration()
+	deadline := time.Now().Add(timeout)
+
+	return retry.RetryContext(ctx, timeout, func() *retry.RetryError {
+		resp, err := op()
+		if err == nil {
+			return nil
+		}
+
+		if cfg.TreatNotFoundAsSuccess && resp != nil && resp.StatusCode == http.StatusNotFound {
+			gcomDrainResponse(resp)
+			return nil
+		}
+
+		if transientGCOMError(cfg, resp, err) {
+			logGCOMRetryWarning(resp, err)
+			if resp != nil && resp.StatusCode == http.StatusTooManyRequests {
+				wait, ok := parseRetryAfter(resp.Header.Get("Retry-After"), time.Now())
+				if ok && wait > time.Until(deadline) {
+					gcomDrainResponse(resp)
+					return retry.NonRetryableError(fmt.Errorf("rate limited by the Grafana Cloud API and Retry-After (%s) exceeds the remaining retry budget, giving up: %w", wait, err))
+				}
+				sleepRetryAfter(ctx, wait)
+			}
+			gcomDrainResponse(resp)
+			return retry.RetryableError(err)
+		}
+
+		gcomDrainResponse(resp)
+		return retry.NonRetryableError(err)
+	})
+}
+
+func transientGCOMError(cfg GCOMRetryConfig, resp *http.Response, err error) bool {
+	if err == nil {
+		return false
+	}
+	if cfg.OnlyRetryRateLimited {
+		return resp != nil && resp.StatusCode == http.StatusTooManyRequests
+	}
+	if DefaultGCOMTransient(resp, err) {
+		return true
+	}
+	if cfg.OnTransient != nil && cfg.OnTransient(resp, err) {
+		return true
+	}
+	return false
+}
+
+// DefaultGCOMTransient classifies Grafana Cloud transient HTTP responses and transport errors.
+//
+// Treats typical 429/408/5xx as retryable. Does not treat 409 as retryable — stack creation keeps bespoke conflict handling (see resource_cloud_stack createStack).
+//
+// Errors satisfying [net.Error] (including *[net.OpError]) are retryable unless the chain includes [context.DeadlineExceeded] (explicit client deadline / cancellation).
+func DefaultGCOMTransient(resp *http.Response, err error) bool {
+	if err == nil {
+		return false
+	}
+
+	if resp != nil {
+		switch resp.StatusCode {
+		case http.StatusTooManyRequests, http.StatusRequestTimeout:
+			return true
+		case http.StatusInternalServerError, http.StatusBadGateway, http.StatusServiceUnavailable, http.StatusGatewayTimeout:
+			return true
+		default:
+			return resp.StatusCode >= 500
+		}
+	}
+
+	if errors.Is(err, context.DeadlineExceeded) {
+		return false
+	}
+
+	var netErr net.Error
+	if errors.As(err, &netErr) {
+		return true
+	}
+
+	return errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF)
+}
+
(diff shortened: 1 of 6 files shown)


internal/resources/cloud/gcom_retry.go
```diff
@@ -0,0 +1,199 @@
...
+	gcomRetryAfterCap       = 2 * time.Minute
+)
+
+// GCOMRetryConfig configures RetryGCOM.
+type GCOMRetryConfig struct {
+	Timeout time.Duration
+
+	// TreatNotFoundAsSuccess, when true, treats HTTP 404 responses as success (typically for DELETE),
+	// so destroys are idempotent when the resource is already gone.
+	TreatNotFoundAsSuccess bool
```
nachogiljaldo: Instead of this, I would:
- create a new type ErrorAnalyzer func(error) error
- pass an instance to the retry
- implement one that is AcceptNotFounds

that way, you can compose them as you will need this in other grafana cloud calls that also require retries.

internal/resources/cloud/gcom_retry.go
```diff
@@ -0,0 +1,199 @@
...
+
+	// OnlyRetryRateLimited, when true, restricts retries to HTTP 429 responses.
+	//
+	// Use this for non-idempotent operations (POST creates): grafana.com rate limiting rejects
+	// requests before they are processed, so a 429 is always safe to retry. 5xx responses and
+	// transport errors are ambiguous — the request may have been processed server-side even though
+	// no response was received — and retrying could create duplicate resources.
+	// When set, DefaultGCOMTransient and OnTransient are not consulted.
+	OnlyRetryRateLimited bool
```
nachogiljaldo: same here, instead of hardcoding behaviors, make them configurable and composable.

internal/resources/cloud/gcom_retry.go
```diff
@@ -0,0 +1,199 @@
...
+
+// RetryGCOM runs op under retry.RetryContext until success or non-retryable error.
+//
+// Follow-up retries:
+//   - For HTTP 429, sleeps according to Retry-After (RFC 9110 delta-seconds or HTTP-date), capped by gcomRetryAfterCap and ctx cancellation.
+//     If the requested wait exceeds the remaining retry budget, the error is returned immediately instead of sleeping and timing out anyway
+//     (grafana.com rate limit windows can be up to an hour, far beyond the retry timeout).
+//   - When TreatNotFoundAsSuccess is set, HTTP 404 is treated as immediate success (common for idempotent DELETE).
+//   - Failed response bodies are drained for connection reuse (success responses are unchanged).
+func RetryGCOM(ctx context.Context, cfg GCOMRetryConfig, op func() (*http.Response, error)) error {
```
nachogiljaldo: I would call this RetryHTTPRequest because it does not seem gcom specific at all.

internal/resources/cloud/http_request_retry.go
```diff
@@ -0,0 +1,199 @@
...
+			if resp != nil && resp.StatusCode == http.StatusTooManyRequests {
+				wait, ok := parseRetryAfter(resp.Header.Get("Retry-After"), time.Now())
+				if ok && wait > time.Until(deadline) {
+					gcomDrainResponse(resp)
+					return retry.NonRetryableError(fmt.Errorf("rate limited by the Grafana Cloud API and Retry-After (%s) exceeds the remaining retry budget, giving up: %w", wait, err))
+				}
+				sleepRetryAfter(ctx, wait)
+			}
+			gcomDrainResponse(resp)
+			return retry.RetryableError(err)
```
nachogiljaldo, line 111: there should be a wait after non 429s too and it should be configurable.

internal/resources/cloud/gcom_retry.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+func transientGCOMError(cfg GCOMRetryConfig, resp *http.Response, err error) bool {
+	if err == nil {
+		return false
+	}
+	if cfg.OnlyRetryRateLimited {
+		return resp != nil && resp.StatusCode == http.StatusTooManyRequests
+	}
+	if DefaultGCOMTransient(resp, err) {
```
nachogiljaldo: I think this should be configurable with a default to this behavior.

internal/resources/cloud/gcom_retry.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+func transientHTTPError(cfg HTTPRequestRetryConfig, resp *http.Response, err error) bool {
+	if err == nil {
+		return false
+	}
+
+	analyzers := cfg.TransientErrorAnalyzers
+	if len(analyzers) == 0 {
+		analyzers = []TransientErrorAnalyzer{DefaultGCOMTransient}
```
nachogiljaldo: I wonder if instead of having these default values here, we should just have a DefaultHttpRequestRetryConfig and assume it is there. The advantage of that is that you have 1 place that summarizes what is used and also that it is up to the user to select it, while here it is somehow "hidden".

internal/resources/cloud/resource_cloud_access_policy.go
```diff
@@ -206,8 +207,13 @@ func createCloudAccessPolicy(ctx context.Context, d *schema.ResourceData, client
 			Conditions:  expandCloudAccessPolicyConditions(d.Get("conditions").(*schema.Set).List()),
 		})
 
-	result, _, err := req.Execute()
-	if err != nil {
+	// POST creates are not idempotent: only retry 429s, which are rejected before processing.
```
nachogiljaldo: I don't think this is accurate. Imagine you get a 500, that is perfectly retryable. I _think_ the key here is that before retrying you should read if the policy already exists, or at least abort immediately should you get a 400 / 409 that are indeed not retriable.

nachogiljaldo (commented):
Looking pretty good and glad you started with small examples, I left a few suggestions.
Additionally it would be nice if we could add logs to the tracing machinery.

internal/resources/cloud/gcom_retry.go
s-gohel: Addressed — added DefaultHTTPRequestRetryConfig() as the single explicit Grafana Cloud retry policy and updated callers/tests to opt into it.

internal/resources/cloud/resource_cloud_access_policy.go
s-gohel: Addressed — access policy creates now use the default retry policy, so transient 5xx errors retry while 400/409 client/conflict responses remain non-retryable; added unit coverage for the 400/409 behavior.

internal/resources/cloud/gcom_retry.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+func retryLogOperation(operation string) string {
+	if operation == "" {
+		return "request"
+	}
+	return operation
+}
+
+// AcceptNotFounds treats HTTP 404 responses as success, typically for idempotent DELETE calls.
```
nachogiljaldo: AcceptNotFound instead.

internal/resources/cloud/http_request_retry.go
```diff
@@ -0,0 +1,259 @@
+package cloud
```
nachogiljaldo, line 1: change to http_request_retry.go ?

internal/resources/cloud/http_request_retry.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+// DefaultGCOMTransient classifies Grafana Cloud transient HTTP responses and transport errors.
+//
+// Treats typical 429/408/5xx as retryable. Does not treat 409 as retryable — stack creation keeps bespoke conflict handling (see resource_cloud_stack createStack).
+//
+// Errors satisfying [net.Error] (including *[net.OpError]) are retryable unless the chain includes [context.DeadlineExceeded] (explicit client deadline / cancellation).
+func DefaultGCOMTransient(resp *http.Response, err error) bool {
+	if err == nil {
+		return false
```
nachogiljaldo, line 175: no error should be transient, right?

internal/resources/cloud/gcom_retry.go
```diff
@@ -0,0 +1,259 @@
...
+		case http.StatusTooManyRequests, http.StatusRequestTimeout:
+			return true
+		case http.StatusInternalServerError, http.StatusBadGateway, http.StatusServiceUnavailable, http.StatusGatewayTimeout:
+			return true
+		default:
+			return resp.StatusCode >= 500
+		}
+	}
+
+	if errors.Is(err, context.DeadlineExceeded) {
```
nachogiljaldo: context.Canceled too?

internal/resources/cloud/resource_cloud_access_policy.go
```diff
@@ -206,8 +207,15 @@ func createCloudAccessPolicy(ctx context.Context, d *schema.ResourceData, client
...
 		})
 
-	result, _, err := req.Execute()
-	if err != nil {
+	// Retry transient server and rate-limit failures; 400/409 responses are client/conflict errors and return immediately.
+	var result *gcom.AuthAccessPolicy
+	cfg := DefaultHTTPRequestRetryConfig()
+	cfg.Operation = "create cloud access policy"
+	if err := RetryHTTPRequest(ctx, cfg, func() (*http.Response, error) {
+		r, httpResp, err := req.Execute()
```
nachogiljaldo, line 241: I think this was missed from a previous comment, but here we should (before retrying), check if it exists already or not?

internal/resources/cloud/http_request_retry.go
```diff
@@ -0,0 +1,259 @@
...
+//     If the requested wait exceeds the remaining retry budget, the error is returned immediately instead of sleeping and timing out anyway
+//     (grafana.com rate limit windows can be up to an hour, far beyond the retry timeout).
+//   - ErrorAnalyzer can accept or transform errors before they are classified for retry.
+//   - Failed response bodies are drained for connection reuse (success responses are unchanged).
+func RetryHTTPRequest(ctx context.Context, cfg HTTPRequestRetryConfig, op func() (*http.Response, error)) error {
+	timeout := cfg.Timeout
+	deadline := time.Now().Add(timeout)
+	attempt := 0
+
+	return retry.RetryContext(ctx, timeout, func() *retry.RetryError {
```
nachogiljaldo, line 83: I just thought about this, but doesn't this do its own sleeps?

internal/resources/cloud/resource_cloud_access_policy.go
s-gohel: added checks

internal/resources/cloud/http_request_retry.go
s-gohel: Right, `err == nil` means the call succeeded, so it's not transient and we return `false`. It's mostly defensive.

nachogiljaldo (approved):
LGTM

PR #2838, opened 2026-06-17
fix(appplatform): retry transient referential-integrity and server errors

rknightion:
## Summary

App Platform resources only retried HTTP 409 conflicts on delete, and did not retry reads at all. Against the provisioning (Git Sync) API this surfaces as two intermittent, eventual-consistency failures:

1. **Destroy 422 — cross-resource referential-integrity race.** When a connection and a repository that references it (`spec.connection.name`) are destroyed together, Terraform deletes the repository first. The repository `DELETE` returns `200` and stamps a `deletionTimestamp`, but the object lingers in storage until its finalizers complete.

   The connection delete validator decides at admission via a live `ListByConnection` (field selector `spec.connection.name`). With grafana/grafana#126822 (merged) the validator now ignores repositories that are themselves terminating (`DeletionTimestamp != nil`), which reduces the previously-*persistent* 422 to a narrow **storage read-after-write window**: if the validator's list has not yet observed the repository's freshly-written `deletionTimestamp` (~100ms; a field-selector list may be served from the eventual-consistent index path rather than a quorum read), it still counts the repository as a live reference and returns:

   ```
   Error: invalid value for field "metadata.name":
     * Forbidden: cannot delete connection: referenced by 1 repository(s): [...]
   ```

   This 422 (a `StatusReasonInvalid`, not a 409) was not retried, so the destroy failed even though the window clears moments later. The backend change and this client retry are complementary halves of the fix.

2. **Plan/refresh 5xx.** A transient `500` (or `503`/timeout/`429`) on a `GET` during plan/refresh failed the whole operation with no retry. No server-side change covers this yet, so it stands on its own.

## Changes

- Extract a generic `retryWhile(ctx, attempts, delay, retryable, run)` helper. `retryOnConflict` is now a thin wrapper over it — **update-path behaviour is unchanged** (still 409-only, 5×200ms). A non-positive attempt budget is clamped to a single attempt, so a misconfigured budget can never return `nil` without running.
- **Delete** now retries on conflict (409) **or** a transient referential-integrity error (422 `Invalid` whose message reports it is still "referenced by" another resource), bounded at 6×2s so a genuinely-permanent reference still surfaces its real error promptly rather than hanging.
- **Read** and **ImportState** now retry transient server errors (500/503/server-timeout/504/429) at 5×1s; `NotFound` is unaffected and still removes the resource from state.
- Predicates (`isReferencedDependencyError`, `isRetryableDeleteError`, `isRetryableServerError`) and `retryWhile` are unit-tested, including negative cases (a genuine validation 422 is **not** retried) and the non-positive-attempts guard.

## Scope / blast radius

The retry logic lives in the generic App Platform `Resource[T, L]` base, so it applies to all App Platform resources, but the new behaviour only changes outcomes for the transient error classes above; non-retryable errors fail exactly as before. The referential-integrity 422 is the documented behaviour of the provisioning connection delete validator (`apps/provisioning/pkg/connection/delete_validator.go`).

## Related backend work

- **grafana/grafana#126822 (merged):** delete validator ignores terminating repositories — turns the *persistent* 422 (entire finalizer-completion window, seconds) into a *transient* one (storage read-after-write lag) that this PR absorbs.
- **grafana/grafana#126819 (closed):** a finalizer-based alternative; explored but closed. It doesn't touch the validator's read path, so it would not have removed the need for this retry.

## Notes / follow-ups

- The "referenced by" classification matches the admission controller's message text (gated behind `IsInvalid`); the marker is centralized in a named constant with a source pointer. The 422 is a generic `Invalid` + `Forbidden` on `metadata.name`, so the message text is the only distinguishing signal available client-side — a stable machine-readable marker would be a backend follow-up.
- The update-path's inner `Get` (on conflict re-fetch) is intentionally left as conflict-only retry — pre-existing behaviour, out of scope for this fix.
- The combined connection+repository destroy path is exercised by the existing `TestAccProvisioningRepository_viaConnection` acceptance test.

## Testing

- `go test ./internal/resources/appplatform/...` — pass
- `go vet ./internal/resources/appplatform/...` — clean
- `golangci-lint run ./internal/resources/appplatform/...` — 0 issues
- `gofmt` — clean

No schema/example changes, so no docs regeneration required.

### DIFF
--- a/internal/resources/appplatform/resource.go
+++ b/internal/resources/appplatform/resource.go
@@ -34,6 +34,20 @@ const (
 	conflictRetryAttempts  = 5
 	conflictRetryDelay     = 200 * time.Millisecond
 
+	// Delete retries cover the cross-resource referential-integrity race (e.g. deleting a
+	// connection right after its repository): the reference takes a short time to be
+	// reconciled away, so we wait longer than the optimistic-lock retry. The budget is a
+	// balance — long enough to absorb the reconcile window, but bounded so a genuinely
+	// permanent reference (a connection that is still referenced and not being destroyed)
+	// surfaces its real error reasonably quickly rather than hanging.
+	deleteRetryAttempts = 6
+	deleteRetryDelay    = time.Second
+
+	// Read retries cover transient server-side failures (5xx) so a single backend blip
+	// does not fail an entire plan/refresh.
+	readRetryAttempts = 5
+	readRetryDelay    = 200 * time.Millisecond
+
 	// DefaultManagerIdentity is the static identity stamped on App Platform resources
 	// managed by this Terraform provider. It intentionally excludes version numbers
 	// so that upgrading the provider or Terraform does not change the identity
@@ -392,7 +406,12 @@ func (r *Resource[T, L]) readModel(ctx context.Context, data ResourceModel, resp
 		return
 	}
 
-	res, err := r.client.Get(ctx, obj.GetName())
+	var res T
+	err := retryWhile(ctx, readRetryAttempts, readRetryDelay, isRetryableServerError, func(_ int) error {
+		var gerr error
+		res, gerr = r.client.Get(ctx, obj.GetName())
+		return gerr
+	})
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			resp.State.RemoveResource(ctx)
@@ -680,7 +699,7 @@ func (r *Resource[T, L]) deleteModel(ctx context.Context, data ResourceModel, re
 		return
 	}
 
-	if err := retryOnConflict(ctx, conflictRetryAttempts, conflictRetryDelay, func(_ int) error {
+	if err := retryWhile(ctx, deleteRetryAttempts, deleteRetryDelay, isRetryableDeleteError, func(_ int) error {
 		return r.client.Delete(ctx, obj.GetName(), sdkresource.DeleteOptions{})
 	}); err != nil {
 		if apierrors.IsNotFound(err) {
@@ -707,7 +726,12 @@ func (r *Resource[T, L]) importStateModel(
 	resp *resource.ImportStateResponse,
 	setState func(updated ResourceModel),
 ) {
-	res, err := r.client.Get(ctx, req.ID)
+	var res T
+	err := retryWhile(ctx, readRetryAttempts, readRetryDelay, isRetryableServerError, func(_ int) error {
+		var gerr error
+		res, gerr = r.client.Get(ctx, req.ID)
+		return gerr
+	})
 	if err != nil {
 		resp.Diagnostics.Append(ErrorToDiagnostics(ResourceActionRead, req.ID, r.resourceName, err)...)
 		return
@@ -1743,6 +1767,32 @@ func secureVersionChanged(current, previous types.Int64) bool {
 	}
 }
 
+// retryWhile retries run until it succeeds, the returned error is not retryable
+// (per retryable), or the attempt budget is exhausted. It sleeps delay between
+// attempts and aborts early if ctx is cancelled while waiting.
+func retryWhile(
+	ctx context.Context,
+	attempts int,
+	delay time.Duration,
+	retryable func(error) bool,
+	run func(attempt int) error,
+) error {
+	var err error
+
+	for attempt := 0; attempt < attempts; attempt++ {
+		err = run(attempt)
+		if err == nil || !retryable(err) || attempt == attempts-1 {
+			return err
+		}
+
+		if err := waitForRetry(ctx, delay); err != nil {
+			return err
+		}
+	}
+
+	return err
+}
+
 // retryOnConflict smooths over optimistic-lock races with Grafana's unified storage.
 // Provisioning resources can be reconciled asynchronously by Grafana controllers, so
 // the ResourceVersion held by Terraform can go stale between plan/apply and the API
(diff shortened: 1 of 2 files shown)


MissingRoberto:
I think this should be solved in Grafana directly, and not the provider. If you received a 200 for the deletion of the repository, connection should not fail.

MissingRoberto:
Backend context for this PR, now that the server-side change has landed.

**The 422-on-delete retry here is a required half of the fix — please keep it.**

On the backend, grafana/grafana#126822 changed the connection delete validator to ignore repositories that are themselves terminating (`DeletionTimestamp != nil`). That addresses the root cause of the `terraform destroy` failure:

```
422 Forbidden: cannot delete connection: referenced by N repository(s)
```

But it does **not** fully eliminate the 422 — it shrinks it to a **storage read-after-write window**. The validator decides at admission by doing a live `ListByConnection` (field selector `spec.connection.name`):

1. repository `DELETE` → `200` (deletionTimestamp committed)
2. ~100ms later, connection `DELETE` → validator lists repositories

If that list has not yet observed the repository's freshly-written `deletionTimestamp` (plausible — a field-selector list is likely served from the eventual-consistent search/index path rather than a quorum read), the validator still counts the repository as live and returns the same `422`. Without a client retry, that transient `422` fails the run.

So the two changes are complementary and both are needed:
- **Backend (#126822):** turns a *persistent* 422 (entire finalizer-completion window, seconds) into a *transient* one (storage-lag window).
- **This PR:** retries that transient 422 so the module succeeds.

Separately, the **read-retry for transient 5xx (the intermittent 500-on-GET during plan/refresh)** in this PR addresses a distinct backend issue that no server-side change covers yet, so that part stands on its own.

For the record: a backend finalizer approach was explored (grafana/grafana#126819) but closed — it doesn't touch the validator's read path, so it would not have removed the need for this retry.

internal/resources/appplatform/resource.go
```diff
@@ -34,6 +34,20 @@ const (
...
+	// balance — long enough to absorb the reconcile window, but bounded so a genuinely
+	// permanent reference (a connection that is still referenced and not being destroyed)
+	// surfaces its real error reasonably quickly rather than hanging.
+	deleteRetryAttempts = 6
+	deleteRetryDelay    = time.Second
+
+	// Read retries cover transient server-side failures (5xx) so a single backend blip
+	// does not fail an entire plan/refresh.
+	readRetryAttempts = 5
+	readRetryDelay    = 200 * time.Millisecond
```
MissingRoberto: isn't this one too short?

internal/resources/appplatform/resource.go
```diff
@@ -34,6 +34,20 @@ const (
...
 	conflictRetryDelay     = 200 * time.Millisecond
 
+	// Delete retries cover the cross-resource referential-integrity race (e.g. deleting a
+	// connection right after its repository): the reference takes a short time to be
+	// reconciled away, so we wait longer than the optimistic-lock retry. The budget is a
+	// balance — long enough to absorb the reconcile window, but bounded so a genuinely
+	// permanent reference (a connection that is still referenced and not being destroyed)
+	// surfaces its real error reasonably quickly rather than hanging.
+	deleteRetryAttempts = 6
+	deleteRetryDelay    = time.Second
```
MissingRoberto: maybe 2 seconds?

internal/resources/appplatform/resource.go
```diff
@@ -1775,6 +1796,61 @@ func retryOnConflict(
...
+	run func(attempt int) error,
+) error {
+	return retryWhile(ctx, attempts, delay, apierrors.IsConflict, run)
+}
+
+// referencedDependencyMarker is the substring Grafana's provisioning admission controller
+// uses when a resource cannot be deleted because another resource still references it
+// (e.g. deleting a connection while a repository still points at it via spec.connection.name).
+// Source: apps/provisioning/pkg/connection/delete_validator.go
+const referencedDependencyMarker = "referenced by"
```
MissingRoberto, line 1829: I don't like this, but I don't see a better option currently

MissingRoberto (changes requested):
a couple minor comments about the delay values

ferruvich (commented):
Overall LGTM - I agree on @MissingRoberto's comment about durations we set for delays

internal/resources/appplatform/resource.go
```diff
@@ -1743,27 +1767,21 @@ func secureVersionChanged(current, previous types.Int64) bool {
...
+// (per retryable), or the attempt budget is exhausted. It sleeps delay between
+// attempts and aborts early if ctx is cancelled while waiting.
+func retryWhile(
 	ctx context.Context,
 	attempts int,
 	delay time.Duration,
+	retryable func(error) bool,
 	run func(attempt int) error,
 ) error {
 	var err error
```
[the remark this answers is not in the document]
rknightion: Fixed in ae6dc3ae - `retryWhile` clamps a non-positive `attempts` to one, so it always runs once and returns the real error instead of nil. Added a `TestRetryWhile` case for `attempts` 0 and -3.

internal/resources/appplatform/resource.go
rknightion: Yep, too short - bumped to 1s (5 attempts, ~4s total). Only sleeps on an actual error, so no cost on the happy path. Shout if you want it longer.

internal/resources/appplatform/resource.go
rknightion: Done - 2s now (6 attempts, ~10s total).

internal/resources/appplatform/resource.go
rknightion: Agreed - keeping it for now, easy to swap out if the backend ever exposes a cleaner signal.

rknightion:
Keeping the 422 retry. Updated the `deleteRetry*` / `isReferencedDependencyError` comments and the PR description to the post-#126822 model (residual 422 = storage read-after-write window on the validator's live `ListByConnection`, not a finalizer wait), and linked #126822 / #126819. 5xx read-retry stands on its own as you said.

MissingRoberto:
Tuned the retry counts to match the new delays (pushed in 387b46fc): delete is now `4 × 2s` (~6s wall) and read `3 × 1s` (~2s wall), keeping the total wait budgets bounded so a genuinely-permanent reference still surfaces its real error promptly.

Follow-up: these flat `attempts × delay` budgets are a reasonable first cut, but they're still a guess about the reconcile / read-after-write window. We'll open a follow-up to move this to **exponential backoff** (catch the common sub-second clear fast, grow the delay to absorb a slow window, ideally bounded by elapsed wall-time rather than a fixed attempt count). Keeping it flat here to keep this fix focused.

MissingRoberto (approved):

ferruvich (approved):

MissingRoberto:
@rknightion  you commits must be verified :(

rknightion:
> @rknightion you commits must be verified :(

sorry (first time comiting to a grafana repo). Signed now and rebased (worth noting for https://github.com/grafana/terraform-provider-grafana/pull/2844 so you can rebase and avoid reintroducing my unsigned commits)

PR #2851, opened 2026-06-25
fix(private_data_source_connect_networks): fixed retrieval of pdc networks

### LINKED ISSUE #2850: grafana_cloud_private_data_source_connect_networks returns no results
when using the [grafana_cloud_private_data_source_connect_networks](https://registry.terraform.io/providers/grafana/grafana/latest/docs/data-sources/cloud_private_data_source_connect_networks) data source in grafana cloud,

it returns no results.

if i enable debug logs with `TF_LOG_PROVIDER=debug`, i see this in the logs:

```
[DEBUG] provider.terraform-provider-grafana_v4.39.0: 2026/06/19 16:18:25 [DEBUG] GET https://grafana.com/api/v1/accesspolicies?region=...
.
.
.
data.grafana_cloud_private_data_source_connect_networks.test: Read complete after 0s [id=....
2026-06-19T16:18:25.970+0200 [DEBUG] provider.stdio: received EOF, stopping recv loop: err="rpc error: code = Unavailable desc = error reading from server: EOF"
```

if i run the http-request that it does manually, using `curl`, that request succeeds.

aangelisc:
The condition that checked if a PDC network existed would only succeed if both the legacy and current access policy scope were set. This change fixes that and adds a test suite to validate.

Fixes #2850

### DIFF
--- a/internal/resources/cloud/data_source_private_data_source_connect_networks.go
+++ b/internal/resources/cloud/data_source_private_data_source_connect_networks.go
@@ -88,6 +88,14 @@ type PDCNetworksDataSourceModel struct {
 	PrivateDataSourceNetworks []PDCNetworksDataSourcePolicyModel `tfsdk:"private_data_source_connect_networks"`
 }
 
+// isPDCSigningPolicy reports whether an access policy belongs to a Private Data
+// source Connect network. A policy qualifies if it has either the newer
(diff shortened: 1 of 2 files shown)


asimpson (approved):
Thanks!

PR #2853, opened 2026-06-26
fix(provider): forward http_headers to the app platform client

KennyChenFight:
## Problem

The App Platform (Kubernetes API) client (`createGrafanaAppPlatformClient`) sets only auth on its `rest.Config` and never injects the provider's `http_headers`, unlike the REST clients (which forward them via `grafanaHTTPRoundTripper`). So behind a header-based auth proxy — e.g. Google IAP, which needs a `Proxy-Authorization` header set via `http_headers` / `GRAFANA_HTTP_HEADERS` — every App Platform resource gets HTTP 401 while REST resources work.

## Fix

Wrap the `rest.Config` transport to set the configured `http_headers` on each request, skipping any header already set so configured auth still wins — matching the REST client.

## Testing

A unit test covering both behaviours (configured headers are sent; an existing `Authorization` is preserved). No schema change; not breaking.

### DIFF
--- a/pkg/provider/configure_clients.go
+++ b/pkg/provider/configure_clients.go
@@ -269,6 +269,16 @@ func createGrafanaAppPlatformClient(client *common.Client, cfg ProviderConfig) e
 		}
 	}
 
+	headers, err := getHTTPHeadersMap(cfg)
+	if err != nil {
+		return err
+	}
+	// The Kubernetes rest client does not forward the provider's http_headers like the REST client does.
+	// Inject them on every request so deployments behind a header-based auth proxy (e.g. Google IAP) work.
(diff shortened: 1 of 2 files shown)


Duologic (approved):

Duologic:
This'll need a rebase.

KennyChenFight:
> This'll need a rebase.

@Duologic 
Rebased.

dongho-jung:
One small thing I noticed while comparing this with my duplicate PR (#2919): `RoundTrip` shouldn't modify the incoming request according to the `http.RoundTripper` contract.

It might be safer to clone the request before adding the headers, e.g. `clone := req.Clone(req.Context())`, and pass the clone to the underlying transport.

I used that approach in #2919 and added a test to make sure the original request isn't mutated, in case that's useful as a reference.

PR #2854, opened 2026-06-26
fix(asserts): omit empty match values for null-check operators

vpadi:
## Problem

Knowledge Graph drilldown configs (`grafana_asserts_profile_config`, `_trace_config`, `_log_config`) failed to create whenever a `match` rule used the `IS NOT NULL` / `IS NULL` operator. The apply errored with the misleading message:

```
failed to create profile configuration: data failed to match schemas in oneOf(ApiErrorStatus)
```

This surfaced from `TestAccAssertsProfileConfig_update`.

## Root cause

`buildMatchRules` always called `SetValues` on the DTO, even for an empty list, producing a non-nil empty `[]string`. The generated client serializes via `IsNil` (not `omitempty`), so a non-nil empty slice was emitted as `"values":[]`. The server rejects `values` on null-check operators with a **422**, and that 422 body didn't match the client's `ApiErrorStatus` oneOf — so the real error was masked behind a decode failure.

This could be due to recent changes in the asserts Go client library.

## Changes

- **`buildMatchRules`** only sets `Values` when at least one value is present, so null-check operators no longer send a `values` field.
- The shared match **`values`** attribute is now `Optional`, with a `CustomizeDiff` (`validateMatchRulesDiff`) enforcing the conditional rule at **plan time**: values are required for all operators except `IS NULL` / `IS NOT NULL`, which must have none. Wired into the profile, trace, and log config resources. This follows the provider's idiom for cross-field validation inside list blocks (cf. `resourceCheckCustomizeDiff` in synthetic monitoring) since schema-level `RequiredWith`/`ConflictsWith` don't work inside nested blocks.
- Profile config Create/Update/Delete now use `formatAPIError` (matching trace/log configs) so future API errors surface the raw response body instead of the oneOf decode error.

## Testing

- `go test ./internal/resources/asserts/ -run TestUnitValidateMatchRules` — passes (11 cases).
- `go build ./...` / `go vet` clean.

### DIFF
--- a/docs/resources/asserts_log_config.md
+++ b/docs/resources/asserts_log_config.md
@@ -136,7 +136,10 @@ Required:
 
 - `op` (String) Operation to use for matching. One of: =, <>, <, >, <=, >=, IS NULL, IS NOT NULL, STARTS WITH, CONTAINS.
 - `property` (String) Entity property to match.
-- `values` (List of String) Values to match against.
+
+Optional:
+
+- `values` (List of String) Values to match against. Required for all operators except "IS NULL" and "IS NOT NULL", which must not have any values.
 
 
 <a id="nestedblock--timeouts"></a>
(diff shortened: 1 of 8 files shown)


krritik (approved):
Looks good.

PR #2865, opened 2026-07-07
fix(cloud): apply grafana.com retries across remaining call sites

s-gohel:
## Summary

Continues the work from #2829 for grafana/grafana-com#18305. That PR introduced the transient-error retry machinery (honour `Retry-After` on 429s, retry 408/5xx/network errors, give up when the wait exceeds the retry budget) but deliberately enabled it in only two places to keep the blast radius small: the `grafana_cloud_access_policy` resource and the `grafana_cloud_organization` data source.

This PR rolls it out to the rest of the grafana.com API call sites.

- **Promotes the retry helper** from `internal/resources/cloud` to `internal/common` so it can be shared repo-wide without coupling packages, and adds a `RetryGCOMRequest` convenience wrapper over the default policy. (`git mv`, so history is preserved.)
- **Wraps the remaining grafana.com CRUD/list calls** across the `cloud` package — stacks, stack service accounts + tokens, org members, plugin installation, access-policy tokens, PDC networks + tokens, SM installation, the listers, and the framework data sources — plus `pkg/generate` and `frontendo11y`. Each site keeps its existing 404 / `CheckReadError` semantics; the change is purely additive retries.
- **Leaves existing bespoke retry loops untouched** (stack-create conflict handling, stack connections, plugin install, SA-token create), since the default policy intentionally does not retry 409.
- `frontendo11y` now calls `common.RetryGCOMRequest` directly instead of importing the `cloud` resource package (removes the cross-package coupling).

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/common/... ./internal/resources/cloud/... ./internal/resources/frontendo11y/... ./pkg/generate/... -run TestUnit` (incl. relocated retry unit tests + new `RetryGCOMRequest` tests)
- [x] `gofmt` / `go vet` clean
- [x] Cloud acceptance tests (`TF_ACC_CLOUD_API=true` / `TF_ACC_CLOUD_INSTANCE=true`)

Made with [Cursor](https://cursor.com)


internal/resources/cloud/resource_cloud_access_policy.go
```diff
@@ -307,9 +315,7 @@ func readCloudAccessPolicy(ctx context.Context, d *schema.ResourceData, client *
 	region, id := split[0], split[1]
 
 	var result *gcom.AuthAccessPolicy
-	cfg := DefaultHTTPRequestRetryConfig()
-	cfg.Operation = "read cloud access policy"
-	getErr := RetryHTTPRequest(ctx, cfg, func() (*http.Response, error) {
+	getErr := common.RetryGCOMRequest(ctx, "read cloud access policy", func() (*http.Response, error) {
```
nachogiljaldo: I would not create an alias and just keep it explicit, but if you create the alias, I would remove the GCOM as the default values are reasonable as retries in general.

internal/resources/cloud/resource_cloud_access_policy_token.go
```diff
@@ -180,7 +193,10 @@ func deleteTokenHelper(ctx context.Context, d *schema.ResourceData, client *gcom
 	}
 	region, id := split[0], split[1]
 
-	_, _, err = client.TokensAPI.DeleteToken(ctx, id.(string)).Region(region.(string)).XRequestId(ClientRequestID()).Execute()
+	err = common.RetryGCOMRequest(ctx, "delete policy token", func() (*http.Response, error) {
```
nachogiljaldo: Should we not accept 404s here?

internal/resources/cloud/resource_cloud_org_member.go
```diff
@@ -143,8 +147,10 @@ func (r *orgMemberResource) Create(ctx context.Context, req resource.CreateReque
 	postReq := gcom.NewPostOrgMembersRequest(data.User.ValueString())
 	postReq.SetBilling(billing)
 	postReq.SetRole(data.Role.ValueString())
-	_, _, err := r.client.OrgsAPI.PostOrgMembers(ctx, data.Org.ValueString()).PostOrgMembersRequest(*postReq).XRequestId(ClientRequestID()).Execute()
-	if err != nil {
+	if err := common.RetryGCOMRequest(ctx, "create org member", func() (*http.Response, error) {
```
nachogiljaldo: How does this behave on 409s? Do we just not retry anymore? Should we check that the member does not exist before?

internal/resources/cloud/resource_cloud_org_member.go
```diff
@@ -204,7 +210,10 @@ func (r *orgMemberResource) Update(ctx context.Context, req resource.UpdateReque
 	postReq := gcom.NewPostOrgMemberRequest()
 	postReq.SetBilling(billing)
 	postReq.SetRole(data.Role.ValueString())
-	if _, _, err := r.client.OrgsAPI.PostOrgMember(ctx, data.Org.ValueString(), data.User.ValueString()).XRequestId(ClientRequestID()).PostOrgMemberRequest(*postReq).Execute(); err != nil {
+	if err := common.RetryGCOMRequest(ctx, "update org member", func() (*http.Response, error) {
+		_, httpResp, err := r.client.OrgsAPI.PostOrgMember(ctx, data.Org.ValueString(), data.User.ValueString()).XRequestId(ClientRequestID()).PostOrgMemberRequest(*postReq).Execute()
```
nachogiljaldo: Just a thought, but should this recover if the member is deleted under the hood?

internal/resources/cloud/resource_cloud_org_member.go
```diff
@@ -240,7 +249,9 @@ func (r *orgMemberResource) Delete(ctx context.Context, req resource.DeleteReque
 	org, user := split[0].(string), split[1].(string)
 
 	// DELETE
-	if _, err := r.client.OrgsAPI.DeleteOrgMember(ctx, org, user).XRequestId(ClientRequestID()).Execute(); err != nil {
+	if err := common.RetryGCOMRequest(ctx, "delete org member", func() (*http.Response, error) {
```
nachogiljaldo: Should we consider 404 success here?

nachogiljaldo (commented):
Left a few comments, but I think we should analyze the error handling on each one of the endpoints we're modifying here.

Also it would be nice if we could add some tests where we would see for each one of the calls, how it would behave with various status codes.

internal/resources/cloud/resource_cloud_access_policy.go
s-gohel: Renamed

internal/resources/cloud/resource_cloud_access_policy_token.go
s-gohel: The token delete now treats 404 as success via `AcceptNotFound`, so destroying an already-removed token is idempotent.

internal/resources/cloud/resource_cloud_org_member.go
s-gohel: Addressed - create is now idempotent. We still don't retry 409 (a POST isn't safe to blindly retry), but on a 409 (user already a member) or on a retry after a lost response, the ErrorAnalyzer looks the member up and adopts it instead of failing; the follow-up read reconciles role/billing.

internal/resources/cloud/resource_cloud_org_member.go
s-gohel: Addressed - update now recovers if the membership was deleted under the hood: a 404 from `PostOrgMember` falls back to re-adding the member via `PostOrgMembers` so the apply converges instead of erroring.

internal/resources/cloud/resource_cloud_org_member.go
s-gohel: Addressed - org member delete now treats 404 as success via `AcceptNotFound`, so destroying an already-removed member is idempotent.

internal/resources/cloud/retry_behavior_test.go
```diff
@@ -0,0 +1,411 @@
...
+				t.Fatalf("resource removed = %v, want %v", gotMissing, tt.wantMissing)
+			}
+			if route.count != tt.wantAttempts {
+				t.Fatalf("attempts = %d, want %d", route.count, tt.wantAttempts)
+			}
+		})
+	}
+}
+
+func TestUnitDeleteCloudAccessPolicy_StatusCodes(t *testing.T) {
```
nachogiljaldo: Can we have this on their individual test files with the same name as the file the functionality is defined in?

internal/resources/cloud/retry_behavior_test.go
```diff
@@ -0,0 +1,411 @@
...
+
+func TestUnitReadCloudAccessPolicy_StatusCodes(t *testing.T) {
+	tests := []struct {
+		name         string
+		script       []stubResponse
+		wantErr      bool
+		wantMissing  bool
+		wantAttempts int
+	}{
+		{"200 ok", []stubResponse{{status: 200, body: accessPolicyBody}}, false, false, 1},
```
nachogiljaldo: nit: if we could make these attributes not anonymous (i.e. wantErr: false), it would make reading the test easier.

internal/resources/cloud/retry_behavior_test.go
```diff
@@ -0,0 +1,411 @@
...
+	}
+}
+
+const accessPolicyBody = `{"id":"policy-1","name":"test","realms":[],"createdAt":"2024-01-01T00:00:00Z"}`
+
+func TestUnitReadCloudAccessPolicy_StatusCodes(t *testing.T) {
+	tests := []struct {
+		name         string
+		script       []stubResponse
+		wantErr      bool
```
nachogiljaldo: instead of a boolean, should this be the error we actually want?

internal/resources/cloud/retry_behavior_test.go
```diff
@@ -0,0 +1,411 @@
...
+		wantMissing  bool
+		wantAttempts int
+	}{
+		{"200 ok", []stubResponse{{status: 200, body: accessPolicyBody}}, false, false, 1},
+		{"429 then 200 (retried)", []stubResponse{retryAfterZero(), {status: 200, body: accessPolicyBody}}, false, false, 2},
+		{"503 then 200 (retried)", []stubResponse{{status: 503}, {status: 200, body: accessPolicyBody}}, false, false, 2},
```
nachogiljaldo: I would add a few 50x more (500, 504)

internal/resources/cloud/retry_behavior_test.go
```diff
@@ -0,0 +1,411 @@
...
+func TestUnitDeleteCloudAccessPolicy_StatusCodes(t *testing.T) {
+	// The access policy delete is idempotent: a 404 is treated as success.
+	tests := []struct {
+		name         string
+		script       []stubResponse
+		wantErr      bool
+		wantAttempts int
+	}{
+		{"200 ok", codes(http.StatusOK), false, 1},
+		{"404 idempotent success", codes(http.StatusNotFound), false, 1},
```
nachogiljaldo: Can we add the 5xx here too? why did we not add 400 here?

internal/resources/cloud/retry_behavior_test.go
```diff
@@ -0,0 +1,411 @@
...
+		wantErr      bool
+		wantMissing  bool
+		wantAttempts int
+	}{
+		{"200 ok", []stubResponse{{status: 200, body: accessPolicyBody}}, false, false, 1},
+		{"429 then 200 (retried)", []stubResponse{retryAfterZero(), {status: 200, body: accessPolicyBody}}, false, false, 2},
+		{"503 then 200 (retried)", []stubResponse{{status: 503}, {status: 200, body: accessPolicyBody}}, false, false, 2},
+		{"404 removed from state", codes(http.StatusNotFound), false, true, 1},
+		{"403 terminal error", codes(http.StatusForbidden), true, false, 1},
+		{"400 terminal error", codes(http.StatusBadRequest), true, false, 1},
```
nachogiljaldo: Our APIs often reply with 409s that should be not retried (for the most part), can we add that one?

internal/resources/cloud/retry_behavior_test.go
```diff
@@ -0,0 +1,411 @@
...
+	// contrast to the token / org-member deletes above.
+	tests := []struct {
+		name    string
+		script  []stubResponse
+		wantErr bool
+	}{
+		{"200 ok", codes(http.StatusOK), false},
+		{"404 terminal error (not idempotent)", codes(http.StatusNotFound), true},
+		{"503 then 200 (retried)", []stubResponse{{status: 503}, {status: 200}}, false},
+		{"403 terminal error", codes(http.StatusForbidden), true},
```
nachogiljaldo: What about 409 / 500 ?

nachogiljaldo (commented):
Seems good, but can we make the test tables consistent and complete them with some real status code that can actually happen? (409, 500, 502).

internal/resources/cloud/retry_behavior_test.go
s-gohel: Addressed — split the single retry_behavior_test.go into per-resource files matching where the functionality lives

internal/resources/cloud/retry_behavior_test.go
s-gohel: Addressed — all table cases now use named struct fields (e.g. name:, script:, wantErr:).

internal/resources/cloud/retry_behavior_test.go
s-gohel: Addressed — wantErr is now a string asserted against the actual diagnostic (e.g. "403 Forbidden", "400 Bad Request", "409 Conflict"); empty means no error.

internal/resources/cloud/retry_behavior_test.go
s-gohel: Addressed — added 500 and 504 retried cases to the read test.

internal/resources/cloud/retry_behavior_test.go
s-gohel: Addressed — added 500 and 503 retried cases plus a 400 terminal case to the delete test.

internal/resources/cloud/retry_behavior_test.go
s-gohel: Addressed — added a 409 terminal (not retried) case to the read test, asserting attempts=1.

internal/resources/cloud/retry_behavior_test.go
s-gohel: Addressed — added 409 (terminal, not retried) and 500 (retried) cases to the stack delete test, with attempt-count assertions.

internal/resources/cloud/data_source_cloud_organization.go
```diff
@@ -100,9 +100,7 @@ func (r *CloudOrganizationDataSource) Read(ctx context.Context, req datasource.R
 
 	// Fetch organization from API
 	var org *gcom.FormattedApiOrgPublic
-	cfg := DefaultHTTPRequestRetryConfig()
-	cfg.Operation = "get cloud organization"
-	if err := RetryHTTPRequest(ctx, cfg, func() (*http.Response, error) {
+	if err := common.RetryRequest(ctx, "get cloud organization", func() (*http.Response, error) {
```
nachogiljaldo: nit: this is the same as https://github.com/grafana/terraform-provider-grafana/pull/2865/changes#diff-860d92b6a5e3a054d7dbe964b8097163f0694372824dab918edae7957bd90d30R53 but hey have different messages, also maybe we should have an utility to pull the org with retries to ensure we use the same retry policy across the board.

internal/resources/cloud/data_source_private_data_source_connect_networks.go
```diff
@@ -105,8 +107,12 @@ func (r *PDCNetworksDataSource) Read(ctx context.Context, req datasource.ReadReq
 	if data.RegionFilter.ValueString() != "" {
 		regions = append(regions, data.RegionFilter.ValueString())
 	} else {
-		apiResp, _, err := r.client.StackRegionsAPI.GetStackRegions(ctx).Execute()
-		if err != nil {
+		var apiResp *gcom.GetStackRegions200Response
+		if err := common.RetryRequest(ctx, "list stack regions", func() (*http.Response, error) {
```
nachogiljaldo: same here re: extracting an utility

internal/resources/cloud/data_source_private_data_source_connect_networks.go
```diff
@@ -117,8 +123,12 @@ func (r *PDCNetworksDataSource) Read(ctx context.Context, req datasource.ReadReq
 
 	data.PrivateDataSourceNetworks = []PDCNetworksDataSourcePolicyModel{}
 	for _, region := range regions {
-		apiResp, _, err := r.client.AccesspoliciesAPI.GetAccessPolicies(ctx).Region(region).Execute()
-		if err != nil {
+		var apiResp *gcom.GetAccessPolicies200Response
+		if err := common.RetryRequest(ctx, "list access policies", func() (*http.Response, error) {
```
nachogiljaldo: same for policies.

internal/resources/cloud/resource_cloud_access_policy_token_internal_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+func TestUnitDeleteCloudAccessPolicyToken_StatusCodes(t *testing.T) {
+	// Token delete was made idempotent in review: a 404 now counts as success.
+	tests := []struct {
+		name         string
+		script       []stubResponse
+		wantErr      string
+		wantAttempts int
+	}{
+		{name: "200 ok", script: codes(http.StatusOK), wantAttempts: 1},
```
nachogiljaldo, line 20: the tests are still not consistents across the board, this covers less status codes than https://github.com/grafana/terraform-provider-grafana/pull/2865/changes#diff-f370b6af2731c3788d99f0f9e2ee538fca165d31ff6311185a13eff4824ac3f1R59

internal/resources/cloud/resource_cloud_org_member.go
```diff
@@ -143,8 +147,32 @@ func (r *orgMemberResource) Create(ctx context.Context, req resource.CreateReque
...
-	if err != nil {
+
+	// Make create idempotent. grafana.com returns 409 when the user is already a member of
+	// the org, and a previous attempt may have added the member even though we never observed
+	// the response (transient 5xx / dropped connection). In both cases adopt the existing
+	// membership instead of failing; the follow-up read reconciles role and billing.
+	attempt := 0
+	cfg := common.DefaultHTTPRequestRetryConfig()
+	cfg.Operation = "create org member"
+	cfg.ErrorAnalyzer = func(httpResp *http.Response, err error) error {
```
nachogiljaldo: Maybe I'm wrong, but I don't think this is making idempotent because it does not check the role?

internal/resources/cloud/resource_cloud_stack.go
```diff
@@ -448,8 +453,10 @@ func createStack(ctx context.Context, d *schema.ResourceData, client *gcom.APICl
 		req := client.StacksAPI.UpdateStackV1(ctx, stackCreationResponse.Slug).StackUpdateRequestV1(gcom.StackUpdateRequestV1{
 			DeleteProtection: *gcom.NewNullableBool(&deleteProtection),
 		})
-		_, _, err := req.Execute()
-		if err != nil {
+		if err := common.RetryRequest(ctx, "enable stack delete protection", func() (*http.Response, error) {
+			_, httpResp, execErr := req.Execute()
```
nachogiljaldo: nit: general comment, if the req is just used inside the retry, maybe we should define it in there so it's scoped to it.

nachogiljaldo (commented):
Please, apply those comments generally, try to use an utility when possible (same retry logic for same inputs) and further alignment of test statuses.

internal/resources/cloud/resource_cloud_org_member.go
```diff
@@ -133,22 +171,57 @@ func (r *orgMemberResource) Create(ctx context.Context, req resource.CreateReque
...
+		_, httpResp, err := r.client.OrgsAPI.PostOrgMembers(ctx, org).PostOrgMembersRequest(*postReq).XRequestId(ClientRequestID()).Execute()
+		return httpResp, err
+	}); err != nil {
 		resp.Diagnostics.AddError("Unable to Create Resource", err.Error())
 		return
 	}
 
+	// If we adopted a pre-existing membership, its role/billing may not match the plan. Update
+	// it so create is genuinely idempotent instead of silently leaving a stale role in place.
+	if adopted {
```
nachogiljaldo: Actually, I was wrong, my apologies, I realized this is in theory a bad practice and resources should not be adopted. My bad, so I would just simplify and on 409 retry only if the member does not exist?

nachogiljaldo (commented):
Looks pretty good, I just noticed I had made a mistake when I suggested updating resources that's actually a bad practice, so we should not do that for the org membership, other than that remove that part, rebase and we should be good 🙌

internal/resources/cloud/resource_cloud_access_policy_token_internal_test.go
s-gohel: Addressed in 49f0af61 — the token delete matrix now mirrors the other idempotent deletes (200 / 404 / 400 / 403 / 409 terminal + 429 / 500 / 503 / 504 retried, all with attempt counts), so the status-code coverage is consistent across the suites.

internal/resources/cloud/resource_cloud_org_member.go
s-gohel: Superseded by your follow-up below (296406d5): rather than reconciling the role on adopt, create no longer adopts a pre-existing membership at all, so there is no stale role to reconcile. A genuine 409 now fails the apply.

internal/resources/cloud/resource_cloud_org_member.go
s-gohel: Done in 296406d5 — removed the adopt/reconcile logic. Create now fails on a genuine 409 (member already exists) instead of adopting it; the only 409 we retry is the spurious/racy case where the membership does not actually exist on read (added a TransientErrorAnalyzer that reads the member and retries only when it's absent). Updated the tests accordingly.

nachogiljaldo (approved):

PR #2870, opened 2026-07-09
fix(user): skip user update if only is_admin or password are being modified

volcanonoodle:
Related to https://github.com/grafana/grafana/issues/125915

This allows updating `is_admin` of an externally synced user (e.g. a user that uses SSO) without running the unnecessary `PUT /users/{user_id}` request that will always fail for externally synced users. The fix just checks if any of the `email`, `name` and `username` fields are being modified and, if not, skips the user-update request.

### DIFF
--- a/internal/resources/grafana/resource_user.go
+++ b/internal/resources/grafana/resource_user.go
@@ -164,13 +164,15 @@ func UpdateUser(ctx context.Context, d *schema.ResourceData, meta any) diag.Diag
 	if err != nil {
 		return diag.FromErr(err)
 	}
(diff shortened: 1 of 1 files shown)


volcanonoodle:
I've tested it locally by creating 2 users outside of Terraform: a normal user and a Generic OAuth external user. With the PR, the former can have all its fields modified while the latter can have its `is_admin` field modified but not `name`, `email` or `username`.

Without the PR, the externally synced user cannot have its `is_admin` field modified either.

dmihai (approved):

PR #2872, opened 2026-07-13
fix(appplatform): populate folder uid when present in imported state

khalilhaji:
This PR fixes a bug where we would ignore folder uids set on resources if the destination resource doesn't already have a folder UID defined on it. The fix is to set folder UID when it is defined in the source state.

### DIFF
--- a/internal/resources/appplatform/resource.go
+++ b/internal/resources/appplatform/resource.go
@@ -886,8 +886,8 @@ func GetModelFromMetadata(
 		return diag
 	}
 
-	if !dst.FolderUID.IsNull() && !dst.FolderUID.IsUnknown() {
-		dst.FolderUID = types.StringValue(meta.GetFolder())
+	if folder := meta.GetFolder(); folder != "" {
+		dst.FolderUID = types.StringValue(folder)
 	}
 
 	dst.UUID = types.StringValue(string(src.GetUID()))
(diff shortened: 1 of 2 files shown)


internal/resources/appplatform/resource.go
```diff
@@ -886,8 +886,8 @@ func GetModelFromMetadata(
 		return diag
 	}
 
-	if !dst.FolderUID.IsNull() && !dst.FolderUID.IsUnknown() {
-		dst.FolderUID = types.StringValue(meta.GetFolder())
+	if folder := meta.GetFolder(); folder != "" {
+		dst.FolderUID = types.StringValue(folder)
 	}
```
K-Phoen, line 893: Would it make sense to set `dst.FolderUID` to null if `meta.GetFolder() == ""`?

```go
if folder := meta.GetFolder(); folder != "" {
	dst.FolderUID = types.StringValue(folder)
} else {
	dst.FolderUID = types.StringNull()
}
```

K-Phoen (dismissed):
I just have a small question, LGTM otherwise :)

internal/resources/appplatform/resource.go
khalilhaji: yup! just made that change

K-Phoen (approved):

PR #2874, opened 2026-07-14
feat(alerting): add rulesequence resource

khalilhaji:
## Summary

Adds a new App Platform resource, grafana_apps_rules_rulesequence_v0alpha1, for Grafana Alerting rule sequences. A rule sequence groups existing alert rules and recording rules so they evaluate in order, at a shared interval, within a single folder.

Spec surface:
- trigger.interval (required) — shared evaluation interval.
- recording_rules (required, ≥1) — list of { name = <rule uid> } refs, evaluated in order.
- alerting_rules (optional) — list of { name = <rule uid> } refs.

Referenced rules and the sequence must all live in the same folder (metadata.folder_uid).

This depends on #2872 (populate folder uid when present in imported state)

## Changes

- New resource + schema/parser/saver: internal/resources/appplatform/rulesequence_resource.go
- Registered in pkg/provider/resources.go
- Example: examples/resources/grafana_apps_rules_rulesequence_v0alpha1/resource.tf
- Generated docs + replace_references.go (from make docs)
- SDK bump: github.com/grafana/grafana/apps/alerting/rules → the version that ships the RuleSequence kind (requires Grafana ≥ 13.1, which is where the rulesequences API is served; the acc test is gated CheckOSSTestsEnabled(t, ">=13.1.0")).

## Acceptance test teardown

The acc test's teardown is structured to work around a current server-side behavior, and should be simplified/removed once the API handles rule deletion more gracefully.

The problem: rule-sequence deletion is asynchronous (a finalizer keeps the object Terminating until the rules operator clears it), and admission rejects deleting an alert/recording rule while it still belongs to a live sequence. So a normal single-pass terraform destroy (which deletes the sequence, then immediately its member rules) races the finalizer and the member-rule deletes get a 403 "belongs to rule sequence".

The workaround, entirely in the test:
- An extra TestStep whose config drops only the sequence (keeping the member rules), so that step's apply deletes just the sequence.
- A Check (testAccCheckRuleSequenceDeleted) that polls the API until the sequence is fully gone (NotFound), not merely marked for deletion — so we wait out the finalizer.
- Only then does the framework's final auto-destroy tear down the member rules, by which point no live sequence owns them.

This is more elaborate than a normal acceptance test and exists purely to sidestep the async-delete/guardrail race. It should be removed (collapsed back into a plain single-pass destroy) once the rules app controller allows deleting a rule whose parent sequence is already tagged for deletion (has a deletionTimestamp).

## Testing

- Unit + build/vet green.
- Acceptance test (TestAccRuleSequence) run against a local Grafana 13.1 that serves the rulesequences API — create → update → import → ordered teardown all pass.
- Note: the default CI matrix tops out at 13.0.1 for released versions so this test will skip until a ≥13.1 image is in the matrix.


rwwiv (approved):
:shipit:

PR #2882, opened 2026-07-20
feat(oncall): expose inbound_email computed attribute on integration resource

musculman:
## Summary

- Add a read-only `inbound_email` computed attribute to `grafana_oncall_integration` so users can reference the email address assigned to `inbound_email`-type integrations (e.g. `grafana_oncall_integration.my_email.inbound_email`).
- For all other integration types the API returns empty, so the field is just an empty string — no diff noise.
- Add acceptance test `TestAccOnCallIntegration_inboundEmail` that creates an `inbound_email` integration and asserts the attribute is populated, plus a check in the existing basic test confirming it's empty for non-email types.

### Upgrade note

On the first plan/refresh after upgrading, Terraform may note `inbound_email` as "known after apply" for existing `grafana_oncall_integration` resources. This resolves cleanly on the next apply/refresh without modifying the integration.

## Test plan

- [ ] `TestAccOnCallIntegration_basic` — existing test, now also asserts `inbound_email` is empty for `grafana` type integrations
- [ ] `TestAccOnCallIntegration_inboundEmail` — new test, creates an `inbound_email` integration and verifies the attribute is non-empty
- [ ] `go build ./...` passes
- [ ] `go vet ./internal/resources/oncall/...` passes
- [ ] Docs regenerated (`make docs`) — already committed

### DIFF
--- a/docs/resources/oncall_integration.md
+++ b/docs/resources/oncall_integration.md
@@ -102,6 +102,7 @@ resource "grafana_oncall_integration" "test-acc-integration" {
 ### Read-Only
 
 - `id` (String) The ID of this resource.
+- `inbound_email` (String) The inbound email address for the integration. Only available for integration type `inbound_email`.
 - `link` (String) The link for using in an integrated tool.
 
 <a id="nestedblock--default_route"></a>
--- a/internal/resources/oncall/resource_integration.go
+++ b/internal/resources/oncall/resource_integration.go
@@ -164,6 +164,11 @@ func resourceIntegration() *common.Resource {
 				Computed:    true,
 				Description: "The link for using in an integrated tool.",
 			},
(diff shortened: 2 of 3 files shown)


musculman:
The errors on `cloudinstance` seem to be pre-existing test failures or flaky?

1. TestAccAssertsPromRules_inactive — in internal/resources/asserts (Asserts/Knowledge Graph)
2. TestAccResourceSloWithCustomUUID — in internal/resources/slo (SLO)

Neither test touches the oncall package or the changes in this PR but I would be happy to be corrected on my assumption

roberthorn (approved):

musculman:
looks like all currently opened PRs have the same failure - will try to ask around to see what is going on

PR #2886, opened 2026-07-21
feat(grafana_team): support team administrator memberships

Jguer:
## Summary

- Add an optional `admins` set to `grafana_team` for assigning team administrator permissions.
- Expose administrator memberships through the `grafana_team` data source.
- Handle membership additions, removals, promotions, and demotions without allowing the same user in both membership sets.
- Give an administrator discovered outside Terraform precedence over a legacy `members` configuration when `admins` is not configured, preserving the administrator permission while keeping the planned membership sets disjoint.
- Emit a plan-time warning when a user appears in both `members` and the discovered admin set, noting they will be demoted to a member in the next major version.
- Update the example, generated documentation, and acceptance/unit coverage.

## Testing

- `go test ./internal/resources/grafana -run TestUnit -count=1`


cinaglia:
## Local Testing

Tested the key scenarios locally using a dev build against Grafana 11.0.0 in Docker.

### Setup

```bash
# Built provider from PR branch
git fetch origin feat/team-admins
git checkout origin/feat/team-admins
go build .

# Started Grafana
docker compose up -d  # grafana/grafana:11.0.0

# Used dev_overrides to point Terraform at local build
```

### Scenario 1: Admin created outside Terraform (the warning path)

1. Applied config with 3 users as `members` (no `admins` attribute):
   ```hcl
   members = ["alice@example.com", "bob@example.com", "charlie@example.com"]
   ```

2. Promoted Alice to Team Admin via API (simulating UI):
   ```bash
   curl -X PUT -u admin:admin -d '{"permission": 4}' http://localhost:3000/api/teams/1/members/4
   ```

3. Ran `terraform plan`:
   ```
   Warning: Team members ignored because they are already administrators

   The following users were ignored because they are already team
   administrators: alice@example.com. These users will be set to members in
   the next major version of the Terraform provider.
   ```

4. **Result**: Alice stays an admin (not demoted), moved to `admins` in state.

### Scenario 2: Same user in both sets (validation error)

```hcl
members = ["alice@example.com", "bob@example.com"]
admins  = ["alice@example.com"]
```

```
Error: Invalid team memberships

user alice@example.com cannot be configured as both a team member and a
team admin
```

Correctly rejected at apply time.

### Scenario 3: Proper admin management

```hcl
members = ["bob@example.com", "charlie@example.com"]
admins  = ["alice@example.com"]
```

Clean plan, no warnings.

cinaglia (approved):
Good stuff!

PR #2889, opened 2026-07-23
feat(assistant): support slash commands for skills

svennergr:
## Summary
- add an optional `command_name` attribute to `grafana_assistant_skill`
- manage command enablement through the Assistant command API and reflect its enabled state during refresh/import
- document the attribute and cover command create, update, disable, and state mapping

## Test plan
- [x] `go test ./internal/common/assistantapi ./internal/resources/assistant`
- [x] `git diff --check`
- [ ] Assistant acceptance tests


kolesnikovae (approved):
LGTM

PR #2893, opened 2026-07-24
fix(grafana_team): demote TF-managed admins when omitted into members

Jguer:
## Summary

- When `admins` was previously Terraform-managed and is omitted while a former admin is listed under `members`, demote that user instead of preserving admin rights.
- Keep protect-admin behavior for unmanaged administrators (UI/import) via private-state `admins_configured`.
- Adds unit + acceptance coverage for the omit-admins demotion path.

## Test plan

- [x] `go test ./internal/resources/grafana/ -run 'TestUnit' -count=1`
- [x] OSS acceptance: `TestAccTeam_Members` (omit-admins demotion step + existing UI-promote protect step)
- [ ] Manual: apply with `admins = ["alice@..."]`, then omit `admins` and set `members = ["alice@..."]` → alice becomes member
- [ ] Manual: UI-promoted admin + `members` listing them without ever setting `admins` → still preserved with warning

### DIFF
--- a/docs/resources/team.md
+++ b/docs/resources/team.md
@@ -49,7 +49,7 @@ resource "grafana_team" "test-team" {
 
 ### Optional
 
-- `admins` (Set of String) A set of email addresses corresponding to users who should be given administrator membership to the team. Note: users specified here must already exist in Grafana.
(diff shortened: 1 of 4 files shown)


cinaglia (dismissed):
Tested locally and it worked beautifully.

cinaglia (approved):

mihai-turdean (approved):

PR #2899, opened 2026-07-29
feat(grafana_apps_servicemodel_component_v1alpha1): add resource for service center components

cesararmada:
Adds a new resource `grafana_apps_servicemodel_component_v1alpha1` for managing Grafana Cloud Service Center services (kind `Component` of the `servicemodel.ext.grafana.com/v1alpha1` API), built on the shared app platform resource machinery.

## What it does

- Manages services as catalog components: `title`, `description`, `type` (defaults to `service`), `identifiers` (maximum of 5, matching the API limit), `owner_ref` (defaults to a Grafana IAM team reference), `depends_on_refs`, and `links`.
- `metadata.uid` is validated at plan time against the service identifier naming rule (lowercase letters, numbers and dashes, 2 to 63 characters).
- Import canonicalizes server-empty values to null so a matching configuration plans clean after import.

## Design notes

- **Co-management boundary**: an apply writes the full managed spec, so fields written by other clients are not preserved. Services imported from Backstage via the catalog sync must be managed there, not in Terraform; the resource docs state this.
- **Deliberately unexposed fields**: `spec.identity` (server-derived from the object name), the string-form reference fields, and the Backstage metadata mirror are not exposed. Adding fields later is non-breaking if needs appear.

## Testing

- Unit tests for parser and saver round-trips (the import fidelity contract), the exact wire format, validators, and the plan modifier.
- Acceptance test (`TestAccServiceModelComponent_basic`, gated by `TF_ACC_CLOUD_INSTANCE`) covering minimal and full configurations, update, and import with verification.
- The full lifecycle (create, update, import, delete, and cross-resource references to `grafana_team`) was additionally verified manually against a Grafana Cloud stack with a locally built provider.


nokka (approved):
We had a chat about this one, looks good to me 🎉

PR #2904, opened 2026-07-31
feat(slo): add source_datasource_uid to grafana_slo query blocks

elainevuong:
SLO Plugin must be deployed before Terraform PR can land (since Cloud Acceptance Tests will fail)

---

Adds `source_datasource_uid` to the `freeform` and `ratio` query blocks on the `grafana_slo` resource, and surfaces it on the `grafana_slos` resource.

The SLO API evaluates the SLI query against `query.<type>.sourceDatasourceUid` and writes the generated recording and alerting rules to `destinationDatasource.uid`. The provider previously exposed only the destination, so SLOs that read from one datasource and write to another could not be managed in Terraform.

## Behavior
Field is **Optional**. Omitting the attribute runs the query against the destination datasource — the server's documented fallback — so existing configs are unaffected. In the HCL, `sourceDatasourceUid` cannot be an empty string - it will be rejected at plan by the `nonEmptyStringValidator`.

| HCL (.tf file) | Sent to API | Response from API  |
|---|---|---|
| omitted | field absent | `null` |
| `"some-uid"` | `"some-uid"` | `"some-uid"` |
| `""` | rejected at plan by `nonEmptyStringValidator` | — |

## Design notes
**Write never sends `""`** (`packOptionalString`). Null, unknown, and empty all serialize to an absent field, so clearing the attribute from HCL clears it on the API rather than storing a value the HCL can't express.
**`sourceDatasourceUid` value is sent even when if it is equivalent to the destination**.

## Testing
Unit (`go test ./... -run TestUnit`):
- read: `nil` → null, `""` → `""`, populated → value, both query types
- write: null / unknown / `""` → absent; populated → sent
- round-trip: an omitted attribute plans null and reads back null
- schema wiring for both nested blocks

Acceptance (`TestAccResourceSlo`):
- create with source ≠ destination, plus plan-only and import
- create with source == destination — pins that the API stores the redundant
  value rather than normalizing it away
- create with the attribute omitted
- update: value → different value
- update: value → unset, asserting the field is cleared **on the API**, not just
  absent from state

The source-datasource group was verified green (11/11 steps) against a live Grafana instance.

## Docs

Regenerated with `go generate ./...`. New example at
`examples/resources/grafana_slo/resource_source_datasource.tf`, referenced from
the resource template.

## Dependency

Bumps `grafana/slo-openapi-client/go/slo` to
`v0.0.0-20260730190738-75fe08b09d87` (grafana/slo-openapi-client#27) and removes
the temporary local `replace` directive.

Fixes https://github.com/grafana/slo/issues/4467


deresolution20:
Hi team, do we have a eta on this pr being merged?

elainevuong:
@deresolution20 - apologies - I was on PTO and just got back today. Will work on this and aiming for this to be merged and with a provider release sometime this week.

internal/resources/slo/resource_slo_model.go
```diff
@@ -169,6 +173,17 @@ func packQuery(ctx context.Context, query queryModel) (slo.SloV00Query, diag.Dia
 	}
 }
 
+// packOptionalString sends a null, unknown or empty Terraform value as an absent field.
+// SLO API stores "" verbatim rather than normalizing it, and treats it as
+// equivalent to unset when resolving the query's datasource.
+// Never sending "" keeps Terraform SLOs free of a stored value that would read back as a diff.
+func packOptionalString(v types.String) *string {
+	if v.IsNull() || v.IsUnknown() || v.ValueString() == "" {
```
ellisda: question: are we making this more complicated than we need to be converting terraform-defined `""` into a nil on the wire? Is it clearer for us to just decide at the API layer than sending `source_datasource_uid = ""` means the SLO is invalid?

maybe we're too far down this path to care. I'm just pointing out that it gets confusing when we do these things - it feels like the "GET your PUTs" incident all over again

ellisda (dismissed):
LGTM - one question about how we convert HCL `""` to nil when packing the request payload, since it seems like we could make this simpler if API didn't also convert `""` into nil for the response

internal/resources/slo/resource_slo_model.go
elainevuong: Good question.

A Terraform-defined `""` never actually reaches `packOptionalString` in the first place — `source_datasource_uid = ""` in HCL already fails at `terraform validate/plan`. See the `nonEmptyStringValidator` [here](https://github.com/grafana/terraform-provider-grafana/blob/ff37b60d5ba087334835b0e674b3ed7a5d818b39/internal/resources/slo/resource_slo.go#L83-L94). 

I've removed that for clarity.

ellisda (approved):

PR #2803, opened 2026-06-01
feat(assistant): add Terraform resources for Grafana Assistant

svennergr:
## Summary

Adds Plugin Framework resources to manage Grafana Assistant configuration through the Assistant app plugin proxy API (`/api/plugins/grafana-assistant-app/resources/api/v1/*`).

**New resources:**
- `grafana_assistant_rule`
- `grafana_assistant_skill`
- `grafana_assistant_quickstart`
- `grafana_assistant_mcp_server`

## Why

Teams want to manage Assistant rules, skills, quickstart prompts, and MCP server integrations as code alongside other Grafana resources.

## Key implementation notes

- New `internal/common/assistantapi` HTTP client (reuses provider URL/auth; sends `X-Resource-Scope` on update/delete).
- Nested blocks (not nested attributes) for protocol v5 compatibility in the muxed provider.
- `custom_headers` on MCP servers is sensitive/write-only (not returned by API on read).
- Tenant-scoped MCP servers require at least one auth header (e.g. `Authorization`).

## Testing

- `go test ./internal/common/assistantapi/...`
- `TF_ACC=1 TF_ACC_ASSISTANT=true GRAFANA_URL=http://localhost:3000 GRAFANA_AUTH=admin:admin go test ./internal/resources/assistant/... -run TestAccAssistant -v`
- Manual `terraform apply` / `destroy` against local Grafana with Assistant app installed

PR #2905, opened 2026-08-03
feat(assistant): add watcher and automation resources

wildum:
## Summary

Adds two Plugin Framework resources for Grafana Assistant, on the same Assistant app plugin proxy API (`/api/plugins/grafana-assistant-app/resources/api/v1/*`) that the existing assistant resources use:

- `grafana_assistant_watcher` — always-on agents that evaluate calibrated checks on a schedule
- `grafana_assistant_automation` — saved prompts that run manually or on a cron schedule

## Why

Requested by the field for the AppEnv demo kit ([grafana/appenv#2067](https://github.com/grafana/appenv/issues/2067)) — setting Watchers up by hand blocks rolling the demo environment out to all the places it needs to go. Follows the pattern established in #2803.

## Calibrating a watcher declaratively

Watchers normally reach a runnable state through an interactive, model-driven calibration session, which Terraform can't drive. There is a second path: creating a watcher leaves it in `draft`, and `POST /watcher-agents/{id}/queries` with `finalizeCalibration: true` marks it calibrated and moves it to `ready` with no model in the loop.

So the intended workflow is: calibrate once in the UI, then copy the resulting queries and calibration baseline into the resource to manage it as code.

Create is therefore multi-call — create → finalize calibration → optionally start. Update replaces the whole query list over `PUT` (the `POST` is append-only, so it's create-only), and `started` drives the separate start/pause endpoints. Partial failures persist whatever landed remotely so a half-created watcher stays under Terraform's management rather than being orphaned.

Automations cover the case a watcher can't: `triggerIntervalSeconds` is capped at 3 hours, so anything daily has to be an automation.

## Notes for reviewers

- **Slack workspace connection is not managed here.** It's a one-time OAuth install with no API. Only the delivery target (channel/DM) is declarative.
- **Webhook notifications are not modelled** on automations — their URL and credentials are write-only secure settings the API never returns, so they would diff forever. Slack and email are complete.
- **`thresholds.comparator` and Slack `target.type` are mandatory to the API but declared `Optional`.** Terraform enforces `Required` attributes inside a `SingleNestedBlock` even when the block is omitted, which would break every alerts-only check and every automation without notifications. The blocks are sent as configured so the API rejects an invalid combination with a clear error instead of silently dropping it.
- **Automation notification fields deliberately omit `omitempty`.** The API merges notification patches per provider — a null value deletes, an absent key is left alone — so an unconfigured provider has to serialize as an explicit null for a removed block to actually clear.
- **`trigger_interval_seconds` and query range selectors must move together.** When the interval changes, the API rewrites every range selector that matched the previous interval. Since `expr` is required and non-computed, an interval change that leaves the old selectors in place fails the apply. Documented on the attribute.

## Testing

- `go test ./internal/common/assistantapi/... ./internal/resources/assistant/...` — unit tests cover the create → finalize → start sequence, cursor pagination, and automation CRUD
- `golangci-lint run ./...` — 0 issues
- `go generate ./...` for docs; `terraform fmt` clean on examples
- Validated the generated schema through real Terraform with `dev_overrides`, including configs that omit every optional block, and confirmed the muxed v5 provider serves the nested `actions`/`slack`/`target` blocks

Acceptance tests (`TF_ACC_ASSISTANT=true`) still need a run against a Cloud stack with the Assistant app — Watchers and Automations are both Cloud-only, so they can't run locally. I plan to follow up with the `validate unpublished provider` workflow against appenv before this is released.


wildum:
Ran the full assistant acceptance suite against a local stack from `grafana-assistant-app` (`core` + `search` profiles, `GRAFANA_URL=http://localhost:3000`), matching how #2803 was verified. Watchers and Automations both work locally — the "Grafana Cloud-only" framing in the user docs is product positioning, not a dev-stack constraint, and `assistant.watcher-agents` defaults to enabled in local provisioning.

```
--- PASS: TestAccAssistantSkill_basic (13.51s)
--- PASS: TestAccAssistantMCPServer_basic (15.06s)
--- PASS: TestAccAssistantWatcher_basic (17.47s)
--- PASS: TestAccAssistantRule_basic (17.69s)
--- PASS: TestAccAssistantAutomation_basic (18.25s)
--- PASS: TestAccAssistantQuickstart_basic (18.50s)
ok      github.com/grafana/terraform-provider-grafana/v4/internal/resources/assistant
```

This confirms the central design question: supplying calibrated checks with `finalizeCalibration` really does take a watcher from `draft` to `ready` against a live API, with no interactive calibration session.

PR #2912, opened 2026-08-05
fix(dashboard): don't parse the sha256 sum as JSON for v2 dashboards

### LINKED ISSUE #2911: grafana_dashboard: store_dashboard_sha256 breaks schema-v2 (Kubernetes-style) dashboards
### Terraform Version

1.5.7

4.35.0 (also present in 4.43.0 — see Important Factoids)

### Grafana Version

13.1.1

### Affected Resource(s)

- `grafana_dashboard` (resource)

```hcl
provider "grafana" {
  store_dashboard_sha256 = true
}

resource "grafana_dashboard" "v2" {
  config_json = jsonencode({
    apiVersion = "dashboard.grafana.app/v2"
    kind       = "Dashboard"
    metadata   = { name = "repro-v2" }
    spec = {
      title        = "Repro V2"
      cursorSync   = "Off"
      preload      = false
      editable     = true
      links        = []
      tags         = []
      variables    = []
      annotations  = []
      elements     = {}
      layout       = { kind = "GridLayout", spec = { items = [] } }
      timeSettings = { timezone = "browser", from = "now-6h", to = "now", autoRefresh = "" }
    }
  })
}
```

### Expected Behavior

The dashboard is created and the resource is saved to state, with `config_json`
stored as a sha256 sum.

### Actual Behavior

The dashboard **is** created in Grafana, but the apply fails while reading it back,
so the resource never lands in state. Every subsequent apply repeats the error.

```
Error: invalid character '4' after top-level value
```

The exact character varies per dashboard — other observed forms include:

```
Error: invalid character '8' in literal false (expecting 'a')
```

Classic (v1) dashboards are unaffected, regardless of size.

### Steps to Reproduce

1. Set `store_dashboard_sha256 = true` on the provider.
2. `terraform apply` the configuration above (any schema-v2 envelope will do).
3. The apply fails; the dashboard exists in Grafana.
4. Set `store_dashboard_sha256 = false` and re-apply — the same config succeeds.

### Important Factoids

**Root cause**, in `internal/resources/grafana/resource_dashboard.go`:

```go
func normalizeDashboardBodyJSON(dashboardJSON map[string]any) (string, map[string]any, error) {
	normalizedDashJSON, err := cloneDashboardJSON(dashboardJSON)
	if err != nil {
		return "", nil, err
	}
	delete(normalizedDashJSON, "uid")

	normalizedJSON := NormalizeDashboardConfigJSON(normalizedDashJSON) // (1)
	normalizedMap, err := UnmarshalDashboardConfigJSON(normalizedJSON) // (2)
	if err != nil {
		return "", nil, err
	}
	return normalizedJSON, normalizedMap, nil
}
```

1. `NormalizeDashboardConfigJSON` returns `fmt.Sprintf("%x", configHash[:])` — a
   64-character hex string — when the package-level `StoreDashboardSHA256` is set.
2. That hash is passed straight to `json.Unmarshal`, which cannot parse it.

`normalizeDashboardBodyJSON` is only reached from
`normalizeKubernetesDashboardConfigJSONForState`, which only runs when
`isKubernetesStyleDashboard()` is true — which is why v1 dashboards never hit it.

The error text is a function of the hash's leading hex characters, which is why it
differs between dashboards and looks like several unrelated bugs:

| hash prefix | resulting error |
|---|---|
| `f8…` | `invalid character '8' in literal false (expecting 'a')` |
| `04…` | `invalid character '4' after top-level value` |
| `1a…` | `invalid character 'a' after top-level value` |
| `c0…` | `invalid character 'c' looking for beginning of value` |

`internal/resources/grafana/resource_dashboard.go` is byte-identical in 4.35.0 and
4.43.0 (same md5), so upgrading does not help.

**Suggested fix** — `normalizeDashboardBodyJSON` only needs the normalized map for
comparison, so it should not go through the sha256-aware formatter. Either split the
JSON-producing path from the hashing path (applying the sha256 substitution only
where the value is written to state), or marshal `normalizedDashJSON` directly here
instead of calling `NormalizeDashboardConfigJSON`.

### References

- Dashboard schema v2 / Kubernetes-style resource:
  https://grafana.com/docs/grafana/latest/visualizations/dashboards/build-dashboards/view-dashboard-json-model/

fgouteroux:
Fixes #2911.

`normalizeDashboardBodyJSON` builds its map return value by unmarshalling the
string returned from `NormalizeDashboardConfigJSON`. When `StoreDashboardSHA256`
is set, that string is a 64-character sha256 sum rather than JSON, so the
unmarshal fails and the error surfaces on apply:

```
Error: invalid character '4' after top-level value
```

The dashboard is created in Grafana, but the read-back fails, so the resource
never lands in state and every subsequent apply repeats the error.

This only affects Kubernetes-style (schema v2) dashboards, because
`normalizeDashboardBodyJSON` is reached solely via
`normalizeKubernetesDashboardConfigJSONForState`. Classic dashboards are
unaffected regardless of size.

The reported character varies per dashboard, since it depends on the leading hex
characters of the sum — which makes the same bug look like several unrelated
ones (`f8…` → `invalid character '8' in literal false`, `04…` → `invalid
character '4' after top-level value`, and so on).

### The change

`NormalizeDashboardConfigJSON` takes `map[string]any` by reference and deletes
`id`/`version` in place, so `normalizedDashJSON` is *already* the normalized map
once it returns. Recovering that map by re-parsing the returned string was
redundant even before the sha256 option existed — `cloneDashboardJSON` has
already round-tripped the value through JSON, so the types match.

Returning the map directly leaves the string return untouched. That string is
only used for the `localSpecJSON == remoteSpecJSON` comparison in
`normalizeKubernetesDashboardConfigJSONForState`, and hashing both sides
preserves equality, so sha256 storage keeps working as intended.

### Testing

- Added `TestNormalizeDashboardConfigJSONForStateWithSHA256`, covering both the
  matching and drifting remote-spec branches. It fails without the fix and
  passes with it.
- Existing `TestNormalizeDashboardConfigJSONForState` subtests still pass, so
  behaviour with `store_dashboard_sha256` disabled is unchanged.
- `go test ./...` passes.
- `golangci-lint run ./internal/resources/grafana/...` reports no issues.
- Verified end to end against Grafana 13.1.1 with a schema-v2 dashboard: apply
  fails on an unpatched build with `store_dashboard_sha256 = true`, succeeds on
  a patched build, and still succeeds with the option disabled.

No schema or documentation changes, so `docs/` is unaffected.

### DIFF
--- a/internal/resources/grafana/resource_dashboard.go
+++ b/internal/resources/grafana/resource_dashboard.go
@@ -392,12 +392,12 @@ func normalizeDashboardBodyJSON(dashboardJSON map[string]any) (string, map[strin
 	}
 	delete(normalizedDashJSON, "uid")
 
+	// NormalizeDashboardConfigJSON mutates the map in place, so normalizedDashJSON
+	// is already the normalized map. Recovering it by unmarshalling the returned
+	// string breaks when StoreDashboardSHA256 is set, because the return value is
+	// then a sha256 sum rather than JSON.
 	normalizedJSON := NormalizeDashboardConfigJSON(normalizedDashJSON)
-	normalizedMap, err := UnmarshalDashboardConfigJSON(normalizedJSON)
-	if err != nil {
-		return "", nil, err
-	}
-	return normalizedJSON, normalizedMap, nil
+	return normalizedJSON, normalizedDashJSON, nil
 }
 
 func cloneDashboardJSON(dashboardJSON map[string]any) (map[string]any, error) {
--- a/internal/resources/grafana/resource_dashboard_internal_test.go
+++ b/internal/resources/grafana/resource_dashboard_internal_test.go
@@ -9,6 +9,8 @@ import (
 
 	"github.com/go-openapi/runtime"
 	"github.com/go-openapi/strfmt"
+
+	"github.com/grafana/terraform-provider-grafana/v4/internal/common"
 )
 
 func TestIsKubernetesStyleDashboard(t *testing.T) {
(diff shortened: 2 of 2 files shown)


fgouteroux:
hi team, could you have a look, please ?

Issue #2226, opened 2025-06-24
Repeated diffs for grafana_folder_permission due to team_id/user_id format mismatch


YakshaB:
### Affected Resource(s)

I am seeing repeated diffs for grafana_folder_permission resources. The provider keeps toggling between numeric IDs (e.g., "7") and prefixed IDs (e.g., "1:7"), or between null/0 and the actual ID. This causes Terraform to always show changes, even when nothing has changed in Grafana or my configuration.

**Example diff:**  
```diff
- team_id = "7" -> null
+ team_id = "1:7"
```

**Expected behavior:**  
Terraform should not show a diff if the effective permission assignment has not changed. The provider should normalize IDs or ignore these differences.

**Steps to reproduce:**  
1. Create a `grafana_folder_permission` resource for a team or user.
2. Run `terraform apply`.
3. Run `terraform plan` again — a diff is shown even though nothing has changed.

**Provider version:**  
- grafana/grafana: v2.x

The below response I got from chatgpt:
This persistent diff is a known issue with the Grafana Terraform provider. It happens because:
The provider or Grafana API returns IDs with a prefix (like 1:7), while your state or config may use just 7.
The provider may also reorder or reformat the permissions block, causing Terraform to see a difference every time.
Why does this happen?
Grafana internally uses a composite ID format (orgID:teamID), but Terraform may store or expect just the numeric ID.The provider does not normalize these values, so Terraform always thinks there is a change.

every time when I do terraform apply Im getting this 

<img width="527" alt="Image" src="https://github.com/user-attachments/assets/b900aa3c-e58d-4cb5-bc79-bcb036456c4c" />

any suggestion/workaround to fix this ?

### Terraform Configuration Files

```
resource "grafana_folder_permission" "folder_permissions" {
  for_each   = { for p in var.folder_permissions : "${p.team_key}-${p.folder_key}" => p }
  folder_uid = grafana_folder.folders[each.value.folder_key].uid

  permissions {
    team_id    = grafana_team.teams[each.value.team_key].id
    role       = each.value.role
    permission = each.value.permission
  }
}

resource "grafana_folder_permission" "user_folder_permissions" {
  for_each   = { for p in var.user_folder_permissions : "${p.user_id}-${p.folder_key}" => p }
  folder_uid = grafana_folder.folders[each.value.folder_key].uid

  permissions {
    user_id    = each.value.user_id
    role       = each.value.role
    permission = each.value.permission
  }
}
```

tfvars
[...]

PR #2929, opened 2026-08-10
fix(folder_permission): stop wiping unchanged permissions on Grafana Cloud (#2928)

### LINKED ISSUE #2928: grafana_folder_permission silently wipes existing permissions on Grafana Cloud due to server-side full-replace semantics change
### Terraform Version

opentofu  v1.12.3

4.39.0

### Grafana Version

grafana cloud

### Affected Resource(s)

Other (please describe in the issue), grafana_folder_permission (resource)

Minimal reproducer — two team_ids on one folder, one is added later:

```
resource "grafana_folder" "team_parent" {
  title = "team-parent"
  uid   = "team-parent"
}
resource "grafana_folder_permission" "team_parent" {
  folder_uid = grafana_folder.team_parent.uid
  # first apply — both permissions
  permissions {
    role       = "Viewer"
    permission = "View"
  }
  permissions {
    role       = "Editor"
    permission = "Edit"
  }
}
```

Then edit the second block to change ONLY the Viewer entry (e.g. drop
one team member). The provider computes a diff and posts only the
changed Viewer item. Editor is silently wiped.

# Expected

The provider posts a partial diff and only the specified permissions change. Unrelated Editor/team/user grants are preserved.
This is the historical behavior of POST /api/access-control/folders/{uid} — items in the payload were treated as an incremental set of commands to apply.

ACTUAL BEHAVIOR
============================================================
On Grafana Cloud stacks with the new App Platform ResourcePermission backend, the same endpoint now interprets the posted items as the complete permission set. Anything not in the payload is deleted.
Result:
1. Apply posts a partial diff (only the changed items).
2. Server treats the partial payload as full-replace and wipes every other permission on the folder.
3. Provider reads back a permission set that does not match what it computed and returns:
4. Every subsequent plan shows the wiped grants as additions. Re-running apply re-triggers the wipe on any partial diff. There is no clean path to convergence via the provider.

Real HCL

```hcl
locals {
  teams_files = fileset("${path.module}/../../.cortex/teams", "*.yaml")
  team_names  = [for team in local.teams_files : trimsuffix(team, ".yaml")]

  parent_folder_uid = "team"

  team_okta_groups = {
    for team in local.team_names : team => [
      for g in try(
        yamldecode(file("${path.module}/../../.cortex/teams/${team}.yaml")).info["x-cortex-team"].groups,
        []
      ) : g.name if g.provider == "OKTA"
    ]
  }

  teams_with_okta = {
    for team, groups in local.team_okta_groups : team => groups
    if length(groups) > 0
  }
}

resource "grafana_team" "team" {
  for_each = toset(local.team_names)
  name     = yamldecode(file("${path.module}/../../.cortex/teams/${each.key}.yaml")).info.title
}

resource "grafana_folder" "teams_parent" {
  uid                          = local.parent_folder_uid
  title                        = "Team"
  prevent_destroy_if_not_empty = true
}

resource "grafana_folder_permission" "teams_parent" {
  folder_uid = grafana_folder.teams_parent.uid

  permissions {
    role       = "Viewer"
    permission = "View"
  }
  permissions {
    role       = "Editor"
    permission = "Edit"
  }
}

resource "grafana_folder" "team" {
  for_each                     = toset(local.team_names)
  uid                          = each.key
  title                        = each.key
  parent_folder_uid            = grafana_folder.teams_parent.uid
  prevent_destroy_if_not_empty = true
}

resource "grafana_team_external_group" "team" {
  for_each = local.teams_with_okta
  team_id  = grafana_team.team[each.key].id
  groups   = each.value
}

resource "grafana_folder_permission" "team" {
  for_each   = toset(local.team_names)
  folder_uid = grafana_folder.team[each.key].uid

  depends_on = [grafana_folder_permission.teams_parent]

  permissions {
    team_id    = grafana_team.team[each.key].id
    permission = "Admin"
  }
}
```

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Steps to Reproduce

_No response_

### Important Factoids

_No response_

### References

_No response_

danstadler-pdx:
## What this fixes

Fixes #2928.

`grafana_folder_permission` (and the three sibling bulk-permission
resources — `grafana_dashboard_permission`,
`grafana_datasource_permission`, `grafana_service_account_permission`)
silently wipe unchanged permissions on any partial update when running
against a Grafana Cloud stack whose folder/dashboard/datasource/SA
permissions are handled by the App Platform ResourcePermission backend.

Symptom the user sees:
```
```
…followed by discovering that every managed permission on the resource
is gone, not just the one they removed from config.

## Root cause

`setResourcePermissions` in `common_resource_permission.go` computes a
diff between current server state and desired state, then posts a mix
of add and `Permission: ""` delete commands to
`POST /api/access-control/{resourceType}/{uid}`.

On the App Platform ResourcePermission backend, that endpoint no longer
has incremental "apply these commands" semantics. Empirically it
behaves as follows on that backend (I verified each case against a live
Grafana Cloud stack):

| Payload | Server behavior |
|---|---|
| Partial upsert list | Items in payload are upserted. Items NOT in payload are left untouched. |
| Payload containing any `Permission:""` delete-command | **Entire managed permission set on the resource is wiped**, regardless of what else is in the payload. |
| Full desired-set upsert | Upserts everything in the payload; leaves items not in payload alone. |
| Per-target `SetResourcePermissionsFor{User,Team,BuiltInRole}` with `Permission:""` | Deletes that specific entry only. Same semantics as legacy. |

The old diff-based implementation always emitted `Permission:""`
delete-commands as part of the payload whenever a user removed an item
from config, which under the new semantics is the wipe trigger.

## The fix

Rework `setResourcePermissions` to a two-phase approach:

1. **Bulk POST** the full desired permission set. This upserts
   everything the user declared.
2. **GET** the current state; for any managed entry not in the desired
   set, issue a **per-target DELETE** via
   `SetResourcePermissionsFor{User,Team,BuiltInRole}` with
   `Permission: ""`. Those per-target endpoints have reliable
   single-item semantics on both backends.

I also updated `readBulkPermissions` to filter out the acting
identity's auto-attached Admin grant — the server auto-grants Admin to
whichever user/service-account creates a resource, and that grant is
server-managed rather than user-declared. Without the filter, the
first apply on any fresh resource returned "Provider produced
inconsistent result after apply" because the readback contained one
more entry than the plan declared. The old diff-based code was
accidentally masking this by explicitly deleting the SA-Admin as part
of every apply (via the same wipe-triggering `Permission:""` command
that was breaking everything else).

## Blast radius

Four resources share the `setResourcePermissions` helper. All four
are fixed by this one change:

- `grafana_folder_permission`
- `grafana_dashboard_permission`
- `grafana_datasource_permission`
- `grafana_service_account_permission`

## Behavior on legacy backends

Unchanged. The resource docs already state that these resources manage
the *entire set* of permissions:

> "Permissions that aren't specified when applying this resource will
> be removed."

The old diff-based code was a premature optimization; the new
implementation converges the code behavior with the documented
contract, on both backends.

## Test evidence

I built the patched provider locally and ran a Terraform-based
reproducer against a Grafana Cloud stack with the App Platform
ResourcePermission backend active. The reproducer:

1. Creates a folder + 3 permissions {Viewer=View, Editor=Edit, Team=Admin}
2. Changes config to remove the Editor entry only
3. Reads live permissions via legacy `/api/folders/{uid}/permissions`

**Before this PR (provider v4.44.0):**
```
[baseline]         PASS  — 3 permissions created OK
[partial-update]   TF apply ERRORED — 'inconsistent result after apply'
[verify]           FAIL  — permissions WIPED (0 remain)
```

**With this PR applied:**
```
[baseline]         PASS  — 3 permissions created OK
[partial-update]   TF apply succeeded (no error)
[verify]           PASS  — live permissions match expected 2-permission set
```

I also confirmed a plan-apply-plan cycle is idempotent (no drift,
no re-flapping).

I'd be happy to contribute the reproducer as an acceptance test
under `internal/resources/grafana/` if the maintainers want it — happy
to submit as a follow-up PR to keep this change minimal.

## Not addressed in this PR

- **Not touching `_item` resources.** `grafana_folder_permission_item`
  and siblings already use the per-target endpoints and don't hit this
  bug.
- **Not fixing #2226.** That is a separate `team_id` / `user_id`
  normalization bug that causes cosmetic plan-flapping but is
  non-destructive. Different fix, different scope.

## Alternatives considered

I explored several approaches before landing on the two-phase design:

- **Option A: post desired set only** — simple, but doesn't remove
  drift or user-initiated deletions on the new backend (leaves stale
  items behind, breaks the "manages entire set" contract).
- **Option B: post desired set + include acting SA's own Admin** —
  simpler than this PR, but takes away the user's ability to declare
  the SA's permission explicitly (even to deny it).
- **This PR (Option C): two-phase upsert + per-target delete for drift**
  — a bit more code, but honors the documented contract and works on
  both backends without side effects.
- **Server-side fix** — the App Platform ResourcePermission backend
  could be changed so `Permission:""` in a bulk payload behaves as a
  single-item delete instead of a whole-set wipe. That would allow
  reverting to the old provider approach entirely. Out of scope for
  this PR, but worth considering for platform consistency.

Happy to iterate on any of the above if the maintainers prefer a
different approach.

## For the maintainer

Three optional follow-ups I'm happy to do based on your preference:

1. **Regression test.** Add an acceptance test covering the partial-update
   scenario. Would need CI matrix to include a Grafana Cloud stack with
   the App Platform ResourcePermission backend to catch regressions of
   this class going forward.
2. **Additional scenarios in the reproducer** (out-of-band drift
   correction, N-team partial update, import + partial update) — happy
   to convert any of these into acceptance tests too.
3. **Documentation.** Nothing user-facing changed, but the new two-phase
   flow could be worth a short design note in `contribute/` for future
   contributors.

### DIFF
--- a/internal/resources/grafana/common_resource_permission.go
+++ b/internal/resources/grafana/common_resource_permission.go
@@ -305,9 +305,25 @@ func (r *resourcePermissionBulkBase) readBulkPermissions(client *client.GrafanaH
 		return nil, diag.Diagnostics{diag.NewErrorDiagnostic("Failed to read permissions", err.Error())}
 	}
 
+	// Filter out the acting identity's auto-attached Admin grant that the
+	// server adds when this identity creates a resource. On Grafana Cloud,
+	// whichever user or service account runs `terraform apply` is
+	// auto-granted Admin on any folder/dashboard/etc it creates. That entry
+	// is server-managed, not user-declared, so if we don't filter it out
+	// the readback contains one more entry than the plan declared,
+	// producing "Provider produced inconsistent result after apply".
+	actingUserID := getActingUserID(client)
+
 	var items []bulkPermissionItemModel
 	for _, perm := range resp.Payload {
 		if !perm.IsManaged || perm.IsInherited {
+			continue
+		}
+		// Skip the acting identity's auto-Admin grant. Users who genuinely
+		// want to manage this SA/user via TF can use the _item resource
+		// (grafana_folder_permission_item et al.), which does not go
+		// through this bulk readback path.
+		if actingUserID > 0 && perm.UserID == actingUserID {
 			continue
 		}
 		item := bulkPermissionItemModel{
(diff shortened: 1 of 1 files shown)


danstadler-pdx:
This PR is tied to this issue: https://github.com/grafana/terraform-provider-grafana/issues/2928

mihai-turdean (changes requested):
Blocking on the acting-identity filtering.

Both new filters identify a supposedly server-owned grant using only `UserID == actingUserID`. That also matches an explicitly declared permission for the caller. `readBulkPermissions` will therefore hide desired state, while the delete phase makes that identity exempt from reconciliation. In particular, Terraform can no longer reliably downgrade or remove an acting user/service account permission, potentially preserving elevated access contrary to configuration.

There is no ownership signal here that distinguishes an auto-attached Admin grant from a user-managed entry; identity alone is insufficient (and the filter is not even restricted to `Permission == "Admin"`). Please remove this filtering or base it on authoritative server metadata. Fixing the legacy-to-K8s redirect to preserve incremental command semantics is the safer primary fix; it avoids needing this provider-side exception.

PR #2945, opened 2026-08-20
fix(k6): send k6 publisher token via /initialized

Lantero:
Make terraform provider call the plugin's `/initialized` endpoint, using the service account token it already holds, so the plugin itself can propagate the underlying token it holds.

A warning message is displayed to the user if this operation is not successful, prompting an app navigation. This is mostly error handling, my e2e tests proved the happy path works.


internal/resources/cloud/resource_k6_installation_publisher_token.go
```diff
@@ -0,0 +1,115 @@
...
+		"stack_id":  d.Get("stack_id"),
+		"confirmed": stored != nil,
+		"elapsed":   time.Since(started).Round(time.Millisecond).String(),
+	})
+	return nil
+}
+
+func forwardPublisherToken(ctx context.Context, client *http.Client, stackURL, saToken, userAgent string) (*bool, error) {
+	var result struct {
+		PublisherTokenPresent *bool `json:"publisher_token_present"`
```
Lantero: This is a new field coming from https://github.com/grafana/k6-cloud-general-api/pull/2809. We can merge independently, the code is not strict about this being present.

internal/resources/cloud/resource_k6_installation_publisher_token.go
```diff
@@ -0,0 +1,162 @@
...
+
+type stackInitialization struct {
+	Initialized           *bool `json:"initialized"`
+	PublisherTokenPresent bool  `json:"publisher_token_present"`
+	FoldersInitialized    bool  `json:"folders_initialized"`
+	GrafanaRBACEnabled    bool  `json:"grafana_rbac_enabled"`
+}
+
+func (s stackInitialization) pending() []string {
+	var pending []string
```
d14c: nit: would be nice if the var name does not match the func name (pendingCfgs, missing, checks..)

d14c (approved):
LGTM 🚀

PR #2947, opened 2026-08-20
fix(data_source): warn when assumeRoleArn is used instead of sigV4AssumeRoleArn for core Prometheus

### LINKED ISSUE #2927: [Bug]grafana_data_source: assumeRoleArn is silently ignored on core prometheus data sources
### Terraform Version

_No response_

4.25.0 and 4.41.0

### Grafana Version

Grafana Cloud

### Affected Resource(s)

grafana_data_source (resource)

```hcl
resource "grafana_data_source" "amp" {
  type = "prometheus"
  name = "amp"
  url  = "https://aps-workspaces.${var.region}.amazonaws.com/workspaces/${var.workspace_id}"

  json_data_encoded = jsonencode({
    httpMethod    = "POST"
    sigV4Auth     = true
    sigV4AuthType = "default"
    sigV4Region   = var.region
    assumeRoleArn = var.assume_role_arn # ignored
  })
}
```

### Expected Behavior

Either the role gets assumed, or the apply warns that `assumeRoleArn` is not read by the core `prometheus` data source.

### Actual Behavior

Apply succeeds with no warning. The data source switches to assume-role auth with an empty ARN and every query fails. Migrating off SigV4 keys this way takes the data source down with no signal at plan or apply time.

### Steps to Reproduce

1. Create a `type = "prometheus"` data source using SigV4 access keys.
2. Set `assumeRoleArn` in `json_data_encoded` and drop `sigV4AccessKey` / `sigV4SecretKey`.
3. `terraform apply` succeeds.
4. Query the data source. Authentication fails.

### Important Factoids

Two Prometheus data sources read different keys for the same field. Core `prometheus` reads `sigV4AssumeRoleArn`. The `grafana-amazonprometheus-datasource` plugin reads `assumeRoleArn`.

Since `json_data_encoded` is an untyped passthrough, the wrong key gets stored and never read. Grafana builds `SigV4Config` with an empty `AssumeRoleARN`, `grafana-aws-sdk` skips the assume-role step), and credentials fall back to the base provider chain, which has no access to the account.

The UI hides the difference. `SIGV4ConnectionConfig.tsx` names the form field `assumeRoleArn` but persists `sigV4AssumeRoleArn`. The plugin docs document `assumeRoleArn` correctly for their own type, which is where people copy it from.

Suggested fix: converging the key names is not an option, since `sigV4AssumeRoleArn` is load-bearing for existing data sources. A warning in `checkDeprecatedPrometheusAuth` would catch it at plan time, and that function is already scoped to `type == "prometheus"`: warn when `assumeRoleArn` is set and `sigV4AssumeRoleArn` is not.

A core-Prometheus SigV4 example in `examples/resources/grafana_data_source/resource.tf` would help too. The only assume-role example there is CloudWatch with `grafana_assume_role`.

angelinaaziz:
Fixes #2927

## Problem

When configuring a core `prometheus` datasource with `assumeRoleArn` in `json_data_encoded`, Grafana silently ignores the key. The correct key for core Prometheus is `sigV4AssumeRoleArn` — `assumeRoleArn` is the key used by the `grafana-amazonprometheus-datasource` plugin.

This is easy to get wrong because:
- The `grafana-amazonprometheus-datasource` plugin docs and the CloudWatch datasource both use `assumeRoleArn`
- Grafana's own UI form field is named `assumeRoleArn` but internally persists `sigV4AssumeRoleArn`

The result: IAM role assumption fails silently — Grafana builds `SigV4Config` with an empty `AssumeRoleARN`, falls back to the base credential chain, and every query returns an auth error with no hint about the misconfigured key.

## Solution

Extend `checkDeprecatedPrometheusAuth` (which already handles `sigV4Auth` and `azureCredentials` misconfigurations for core Prometheus) to emit a `diag.Warning` when `assumeRoleArn` is set on a `type = "prometheus"` datasource without `sigV4AssumeRoleArn`.

The warning fires only when `sigV4AssumeRoleArn` is absent — if both keys are present, `sigV4AssumeRoleArn` is used correctly and no warning is needed.

## Changes

- `resource_data_source.go` — new check in `checkDeprecatedPrometheusAuth`
- `resource_data_source_deprecated_auth_test.go` — two new unit tests
- `examples/resources/grafana_data_source/resource.tf` — adds an AMP example showing the correct `sigV4AssumeRoleArn` key
- Fixes a typo in the existing Azure credentials warning (`"lease install"` → `"Please install"`)


angelinaaziz:
@svennergr this PR fixes the silent misconfiguration reported in #2927 — would love a review when you get a chance!

angelinaaziz:
Hey @rjcrutch, could you sign the CLA?  Also any idea of when/if this fix can go in?

rjcrutch:
> Hey @rjcrutch, could you sign the CLA? Also any idea of when/if this fix can go in?

Thanks for the callout. It looks like my CLA has now propagated and is showing as signed. I’ll follow up with the Data Source Plugins team internally to have this reviewed and approved this week as we return from our offsite. I’ll let you know once I have an update on the next steps.

angelinaaziz:
> > Hey @rjcrutch, could you sign the CLA? Also any idea of when/if this fix can go in?
> 
> Thanks for the callout. It looks like my CLA has now propagated and is showing as signed. I’ll follow up with the Data Source Plugins team internally to have this reviewed and approved this week as we return from our offsite. I’ll let you know once I have an update on the next steps.

Thank you 🙏

internal/resources/grafana/resource_data_source.go
```diff
@@ -493,9 +493,20 @@ func checkDeprecatedPrometheusAuth(d *schema.ResourceData) diag.Diagnostics {
 		diags = append(diags, diag.Diagnostic{
 			Severity: diag.Warning,
 			Summary:  "Deprecated authentication method",
-			Detail:   "Azure authentication is deprecated for the core Prometheus data source. lease install Amazon Managed Service for Prometheus found here: https://grafana.com/grafana/plugins/grafana-azureprometheus-datasource/ and then change the type of your data source to 'grafana-azureprometheus-datasource'.",
+			Detail:   "Azure authentication is deprecated for the core Prometheus data source. Please install Azure Managed Prometheus found here: https://grafana.com/grafana/plugins/grafana-azureprometheus-datasource/ and then change the type of your data source to 'grafana-azureprometheus-datasource'.",
```
bossinc:
```suggestion
			Detail:   "Azure authentication is deprecated for the core Prometheus data source. Please install Azure Managed Service for Prometheus found here: https://grafana.com/grafana/plugins/grafana-azureprometheus-datasource/ and then change the type of your data source to 'grafana-azureprometheus-datasource'.",
```

Thank you for finding this typo!

internal/resources/grafana/resource_data_source.go
```diff
@@ -493,9 +493,20 @@ func checkDeprecatedPrometheusAuth(d *schema.ResourceData) diag.Diagnostics {
...
 		})
 	}
 
+	if assumeRoleArn, ok := jsonData["assumeRoleArn"].(string); ok && assumeRoleArn != "" {
+		sigV4Arn, hasSigV4Arn := jsonData["sigV4AssumeRoleArn"].(string)
+		if !hasSigV4Arn || sigV4Arn == "" {
+			diags = append(diags, diag.Diagnostic{
+				Severity: diag.Warning,
+				Summary:  "Incorrect key for IAM role assumption",
+				Detail:   "The 'assumeRoleArn' key is not supported for the core Prometheus data source and will be silently ignored. Use 'sigV4AssumeRoleArn' instead. The 'assumeRoleArn' key is used by the Amazon Managed Service for Prometheus plugin - if you are using that plugin, change the data source type to 'grafana-amazonprometheus-datasource'.",
```
bossinc: `sigV4AssumeRoleArn` and `assumeRoleArn` are no longer supported in the normal prom data source. We should direct the user to use the AWS prom data source if we find these settings, similar to when we find `sigV4Auth`.

For more context, when Grafana reads AWS settings for prom data source from terraform, it will convert the data source to AWS prom.

bossinc (changes requested):
Thank you for picking this issue up! It just needs a small tweak and we should be good :)

internal/resources/grafana/resource_data_source.go
angelinaaziz: updated accordingly 😄

angelinaaziz:
@bossinc made the tweaks a s suggested :D

bossinc (approved):

bossinc:
@angelinaaziz :wave: everything looks good but there are two commits that do not have verified signatures. Please 
This repository requires all commits to be signed. See [GitHub docs on commit signature verification](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification).
```
remote: Review all repository rules at https://github.com/grafana/terraform-provider-grafana/rules?ref=refs%2Fheads%2Ffix%2Fwarn-assume-role-arn-core-prometheus
remote:
remote: - Commits must have verified signatures.
remote:   Found 2 violations:
remote:
remote:   23ffb3f0461d6eed521538f9060f66d30e0980f5
remote:   fd6f9c6deab35f3933fa46c2129597cad6dac4d1
remote:
To github.com:grafana/terraform-provider-grafana.git```

PR #2957, opened 2026-09-07
feat(appplatform): add per-resource org_id override for multi-org management

MissingRoberto:
Slack thread: https://raintank-corp.slack.com/archives/C04RVCAG9B5/p1787923950659599

## Summary

App Platform resources (e.g. `grafana_apps_provisioning_repository_v0alpha1` / GitSync) currently derive their target Kubernetes namespace **solely from the provider-level `org_id`/`stack_id`**, resolved once at `Configure` time and bound to a single namespaced client (`resource.go` `namespaceForClient` → `NewNamespaced`). As a result, a single provider configuration can only manage resources in **one** Grafana organization.

This is a real pain point for self-hosted Enterprise customers who manage many orgs from one Terraform config (created via a `for_each` over `grafana_organization`). Today the only workaround is one **aliased provider block per org** — which can't be generated dynamically, since Terraform evaluates `provider` blocks statically. The classic SDKv2 resources already solve this with a per-resource `org_id`; App Platform resources never got the equivalent.

This PR adds an optional **`metadata.org_id`** override to the App Platform resources, mirroring that SDKv2 convention.

## What changed

- **Per-operation client resolution.** `Resource[T,L]` now keeps a namespace-agnostic typed client plus the provider-default namespaced client, and each CRUD op resolves the namespaced client via `clientForModel` based on `metadata.org_id`. Omitted/null → provider default (unchanged behavior).
- **Schema.** New `metadata.org_id` (`Optional`, `RequiresReplace`). Because it lives on the generic `Resource[T,L]`, it applies to **all App Platform typed resources** (repository, connection, dashboards, servicemodel, secret keeper/securevalue, rules, …).
- **Custom resources too.** The one non-generic resource, `secret_keeper_activation` (its own client + 6-attribute metadata), gets the same per-operation namespace resolution wired into its custom CRUD, so it is multi-org as well.
- **Self-hosted only.** A non-zero `org_id` is **rejected** (attribute-level diagnostic) when the provider targets a Grafana Cloud stack (`stack_id`), rather than silently ignored. Cross-stack management from one config isn't possible anyway (separate endpoints/tokens).
- **Import.** Import IDs accept `"<orgID>:<uid>"` (in addition to `"<uid>"`) to import a resource from a specific org; the org is reflected back into state.
- **Metadata models.** The framework requires struct/object attribute sets to match **exactly**, so `org_id` was threaded consistently through both the generic `ResourceMetadataModel`/`metadataAttrTypes` and the secret `secretMetadataModel`/`secretMetadataBlock`.
- **Docs** regenerated (`go generate ./...`). **Unit tests** added for `splitImportID`, `orgIDFromMetadata`, and the Cloud-rejection path.

## Usage

```hcl
resource "grafana_organization" "orgs" {
  for_each = toset(["team_a", "team_b"])
  name     = each.key
}

# One provider config, many orgs — no aliases
resource "grafana_apps_provisioning_repository_v0alpha1" "gitsync" {
  for_each = grafana_organization.orgs

  metadata {
    uid    = "gitsync-${each.key}"
    org_id = each.value.org_id
  }
  spec {
    title = each.key
    type  = "github"
    github { url = "https://github.com/acme/${each.key}" branch = "main" }
    sync   { enabled = true target = "folder" }
  }
}
```

## Notes / open questions

- Requires **basic auth** for org switching (API keys are already org-scoped; the provider rejects `org_id > 1` with an API key).
- `provider_schema.json` is regenerated by CI automation (needs the `terraform` CLI, unavailable in my environment) — not included here.
- Draft for design review: does exposing `org_id` on **every** App Platform resource (vs. repository-only) match the intended direction? `org_id` was chosen over a raw `namespace` string for consistency with the rest of the provider and direct `grafana_organization.org_id` wiring.

## Test plan

- [x] `go build ./...`
- [x] `go vet` + `gofmt` clean
- [x] `go test ./internal/resources/appplatform/...`
- [x] `go generate ./...` (docs)
- [ ] Acceptance tests against a multi-org Enterprise instance (needs reviewer env)

### DIFF
--- a/internal/common/client.go
+++ b/internal/common/client.go
@@ -40,6 +40,10 @@ type Client struct {
 	GrafanaAppPlatformAPIClientID string
 	GrafanaOrgID                  int64
 	GrafanaStackID                int64
+	// GrafanaAppPlatformUsesAPIKey reports whether the App Platform client authenticates with
+	// an API key / service account token (as opposed to basic auth). API keys are already
+	// org-scoped, so per-resource org_id overrides are unsupported in that mode.
+	GrafanaAppPlatformUsesAPIKey bool
 
 	GrafanaCloudAPI            *gcom.APIClient
 	SMAPI                      *SMAPI.Client
--- a/docs/resources/apps_alertenrichment_alertenrichment_v1beta1.md
+++ b/docs/resources/apps_alertenrichment_alertenrichment_v1beta1.md
@@ -192,6 +192,7 @@ Required:
 Optional:
 
 - `folder_uid` (String) The UID of the folder to save the resource in. For example, it's supported for dashboards and folders. To know if it's supported for the specific resource you're using check the documentation.
+- `org_id` (Number) The Grafana organization ID this resource belongs to, for self-hosted OSS or Enterprise Grafana. When set, it overrides the provider's `org_id` for this resource only, so a single provider configuration can manage App Platform resources across multiple organizations. Not supported on Grafana Cloud (configure a stack with `stack_id` instead). Changing this value forces the resource to be recreated in the new organization.
 
 Read-Only:
 
--- a/docs/resources/apps_dashboard_dashboard_v1beta1.md
+++ b/docs/resources/apps_dashboard_dashboard_v1beta1.md
@@ -57,6 +57,7 @@ Required:
 Optional:
 
 - `folder_uid` (String) The UID of the folder to save the resource in. For example, it's supported for dashboards and folders. To know if it's supported for the specific resource you're using check the documentation.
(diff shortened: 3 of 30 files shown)


MissingRoberto:
@codex review

MissingRoberto:
@codex review

internal/common/client.go
```diff
@@ -40,6 +40,11 @@ type Client struct {
 	GrafanaAppPlatformAPIClientID string
 	GrafanaOrgID                  int64
 	GrafanaStackID                int64
+	// GrafanaAppPlatformBasicAuth reports whether the App Platform client authenticates with
+	// basic auth (user:password). Only basic auth can switch organizations; API keys are
+	// already org-scoped and anonymous auth cannot switch orgs, so per-resource org_id
+	// overrides are only supported in this mode.
+	GrafanaAppPlatformBasicAuth bool
```
MissingRoberto: Not sure about this change.

PR #2969, opened 2026-09-16
fix(grafana_synthetic_monitoring_check_alerts): treat a missing check as already deleted

fabioaraujopt:
## Problem

`grafana_synthetic_monitoring_check_alerts` has no existence independent of its check, but neither its Read nor its Delete tolerates the check having already been removed.

Delete is implemented as an update to an empty alert list. Against a check that has already been deleted, the SM API returns:

```
check alerts update request: status="404 Not Found", msg="invalid updated check", err="check not found"
```

and the resource returns that unchanged as a hard error. Because the parent is never coming back, every retry hits the same 404 and the resource can never be deleted.

This surfaces most sharply through [crossplane-provider-grafana](https://github.com/grafana/crossplane-provider-grafana), which wraps this resource with upjet. Deleting a `Check` and its `CheckAlerts` managed resources together is a race that the `Check` usually wins, after which the `CheckAlerts` object is left with a finalizer that never releases and reports the error above indefinitely. The only recovery is to remove the finalizer by hand.

Read has the same gap: a check deleted out of band fails the refresh with the raw 404 instead of being reported as missing and recreated, which is also what breaks the Crossplane Observe step.

## Root cause

The package already has the convention for this. `resource_check.go` and `resource_probe.go` both detect a missing resource by matching `"404 Not Found"` in the error text and return `common.WarnMissing`. `resource_check_alerts.go` never adopted it.

## Change

- **Delete** returns success when the check is already gone. There is nothing left to remove, and the alerts were deleted along with their check.
- **Read** returns `common.WarnMissing("check alerts", d)` when the check is gone, matching how the check and probe resources behave.
- Adds `isCheckNotFound`, which matches on both the HTTP status text (as the sibling resources do) and the API's `"check not found"` message, since the client surfaces both in the error string.

No schema change, so no generated docs to update.

## Testing

- `gofmt` and `go vet` clean on the package.
- New unit test `TestIsCheckNotFound` covers the exact error string observed from the API, bare `404 Not Found`, bare `check not found`, and — deliberately — asserts that a `400 "invalid check probes"`, a `500`, and a `401` are **not** treated as "not found", so genuine failures still surface.
- `go test ./internal/resources/syntheticmonitoring/ -run TestIsCheckNotFound` passes.

I have not run the acceptance suite, which needs a live SM tenant. The delete-after-parent-removal path would be worth exercising there: create a check with alerts, delete the check directly via the API, then `terraform destroy` — before this change that fails with the 404 above; after it, it succeeds.

I also have not run `golangci-lint` locally (only the Docker wrapper is available to me); the change follows the existing pattern in the package, and I'm happy to fix anything CI flags.


ka3de (commented):
This LGTM. You will have to rebase on top of `main`.
I'll delegate the final review to @VikaCep as she's more familiar with the TF provider.

internal/resources/syntheticmonitoring/resource_check_alerts.go
```diff
@@ -177,12 +184,29 @@ func resourceCheckAlertDelete(ctx context.Context, d *schema.ResourceData, c *sm
...
 	return nil
 }
 
+// isCheckNotFound reports whether the SM API rejected a check-alerts operation
+// because the parent check no longer exists. The client returns the HTTP status
+// and the API message in the error text; this matches the same way the check and
+// probe resources in this package detect a missing resource.
+func isCheckNotFound(err error) bool {
+	msg := err.Error()
+	return strings.Contains(msg, "404 Not Found") || strings.Contains(msg, "check not found")
```
digitalcrab, line 207: any error containing `check not found` is classified as missing, regardless of HTTP status
The SM client returns a typed `*smapi.HTTPError` with a Code field.

VikaCep (commented):
The lifecycle change looks good to me. I agree with Maksim: please use the typed HTTP error’s status code and update the tests, including non-404 responses containing “check not found”. Please also cover the Read/Delete callbacks and rebase as requested.

